PR #22, opened 2017-08-16
implement h2::server::Stream::send_reset(Reason) and Body::is_empty()

olix0r:


src/proto/streams/state.rs
```diff
@@ -243,6 +243,19 @@ impl State {
         }
     }
 
+    /// Indicates that the local side will not send more data to the local.
+    pub fn send_reset(&mut self, reason: Reason) -> Result<(), ConnectionError> {
+        match self.inner {
+            Idle => Err(ProtocolError.into()),
+            Closed(..) => Ok(()),
+            _ => {
+                trace!("send_reset: HalfClosedRemote => Closed");
```
olix0r: copy pasta

src/proto/streams/state.rs
```diff
@@ -243,6 +243,19 @@ impl State {
         }
     }
 
+    /// Indicates that the local side will not send more data to the local.
+    pub fn send_reset(&mut self, reason: Reason) -> Result<(), ConnectionError> {
+        match self.inner {
+            Idle => Err(ProtocolError.into()),
```
olix0r: probably InvalidId?

src/proto/streams/streams.rs
```diff
@@ -352,6 +362,15 @@ impl<B> StreamRef<B>
         })
     }
 
+    pub fn is_recv_empty(&self) -> bool {
+        let mut me = self.inner.lock().unwrap();
+        let me = &mut *me;
+
+        let stream = me.store.resolve(self.key);
+
+        stream.state.is_recv_closed() && me.actions.recv.is_empty()
```
carllerche: Is this correct? It seems to be checking that there are no buffered frames on the entire connection vs. stream?

olix0r: good catch

src/proto/streams/buffer.rs
```diff
@@ -36,6 +36,14 @@ impl<B> Buffer<B> {
             slab: Slab::new(),
         }
     }
+
+    pub fn len(&self) -> usize {
```
carllerche: Looks like these fns aren't used anymore?

src/proto/streams/recv.rs
```diff
@@ -74,6 +74,14 @@ impl<B> Recv<B> where B: Buf {
         }
     }
 
+    pub fn is_empty(&self) -> bool {
```
carllerche: These aren't used anymore either?

src/proto/streams/send.rs
```diff
@@ -107,6 +107,17 @@ impl<B> Send<B> where B: Buf {
         stream.state.send_close()
     }
 
+    pub fn send_reset(&mut self, reason: Reason,
+                      stream: &mut store::Ptr<B>)
+        -> Result<(), ConnectionError>
+    {
+        stream.state.send_reset(reason)?;
```
carllerche: It looks like the reset will still be sent if the stream state is already closed (potentially reset already). Is this expected?

src/proto/streams/streams.rs
```diff
@@ -352,6 +362,15 @@ impl<B> StreamRef<B>
         })
     }
 
+    pub fn is_recv_empty(&self) -> bool {
+        let mut me = self.inner.lock().unwrap();
+        let me = &mut *me;
+
+        let stream = me.store.resolve(self.key);
+
+        stream.state.is_recv_closed() && stream.pending_recv.is_empty()
```
carllerche: `stream.pending_recv` could only contain a trailers frame, and `is_recv_empty()` would return false, yet there are no more data frames?

olix0r: is it worth just tracking a bool for this?

olix0r: Actually, i think this is okay if it's only meant to determine whether eos may be set...

src/proto/streams/send.rs
```diff
@@ -107,6 +107,20 @@ impl<B> Send<B> where B: Buf {
         stream.state.send_close()
     }
 
+    pub fn send_reset(&mut self, reason: Reason,
+                      stream: &mut store::Ptr<B>)
+        -> Result<(), ConnectionError>
+    {
+        if !stream.state.is_closed() {
```
olix0r: but it should perhaps be an error to send a reset on a closed stream...

src/lib.rs
```diff
@@ -60,6 +60,13 @@ pub struct Chunk<B: IntoBuf> {
 
 // ===== impl Body =====
 
+impl<B: IntoBuf> Body<B> {
+    pub fn is_empty(&self) -> bool {
+        // If the recv side is closed and the receive queue is empty, the body is empty.
+        self.inner.is_recv_eos()
```
carllerche: As per the comment thread below, this will return `false` if the body is consumed and the next pending frame is a trailer.

olix0r: i think that's correct fsvo is_empty. I intend to expose "is there a body at all" and not "is there remaining body to be consumed".  i should probably rename this

carllerche:
Merging. #yolo

PR #24, opened 2017-08-17
Switch Akamai example from rust-openssl to Rustls.

briansmith:
With this change, h2 can build and run without any manual configuration
steps for -msvc targets. Previously manual installation of OpenSSL
libraries was required.


briansmith:
Note: This PR contains two commits. The first commit is the commit from PR #23.

PR #25, opened 2017-08-18
New send flow control

carllerche:
**This is a work in progress**


src/proto/streams/flow_control.rs
```diff
@@ -1,121 +1,86 @@
...
-        if self.next_window_update == 0 {
-            None
-        } else {
-            Some(self.next_window_update)
-        }
+    /// Decrements the window reflecting data has actually been sent. The caller
+    /// must ensure that the window has capacity.
+    pub fn send_data(&mut self, sz: WindowSize) {
+        assert!(sz <= self.window_size as WindowSize);
+        self.window_size -= sz as i32;
```
olix0r: as noted in chat, we'll also need to account for what happens when the initial window size is decreased.

src/proto/streams/prioritize.rs
```diff
@@ -42,83 +39,174 @@ impl<B> Prioritize<B>
...
+    /// Request capacity to send data
+    pub fn reserve_capacity(&mut self, mut capacity: WindowSize, stream: &mut store::Ptr<B>)
+        -> Result<(), ConnectionError>
+    {
+        if capacity == stream.requested_send_capacity {
+            // Nothing to do
+            return Ok(());
+        } else if capacity < stream.requested_send_capacity {
+            // TODO: release capacity
+            unimplemented!();
```
olix0r: ... ;)

carllerche:
Ok, I restructured the send flow control such that data writes are always accepted. This will make it easier to accept SETTINGS frames that decrease the window.

PR #28, opened 2017-08-23
remove possibly sensitive data from default Debug output

seanmonstar:
I noticed logs included bytes from data and headers frames. It's possible for these things to include secrets or sensitive data, so it should probably not be too easy to accidentally get that data stored in a logfile somewhere.

### DIFF
--- a/src/frame/headers.rs
+++ b/src/frame/headers.rs
@@ -17,7 +17,6 @@ use std::io::Cursor;
 /// Header frame
 ///
 /// This could be either a request or a response.
-#[derive(Debug)]
 pub struct Headers {
     /// The ID of the stream with which this frame is associated.
     stream_id: StreamId,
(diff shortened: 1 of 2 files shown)


src/frame/headers.rs
```diff
@@ -343,6 +342,17 @@ impl<T> From<Headers> for Frame<T> {
     }
 }
 
+impl fmt::Debug for Headers {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_struct("Headers")
+            .field("stream_id", &self.stream_id)
+            .field("stream_dep", &self.stream_dep)
+            .field("flags", &self.flags)
+            // `fields` and `pseudo` purposefully not included
```
carllerche: Shouldn't this be handled by the sensitive flag on `HeaderValue`?

seanmonstar: Depends.

Would a path or query ever contain a secret? Probably. So Psuedos should be excluded.
Would a header ever be sent that isn't marked sensitive that could include a secret? Like `Authorization`, `Proxy-Authorization`, or any custom header?

carllerche: Do you think this is an issue w/ trace level debugging? Removing headers will make debugging noticeably harder.

seanmonstar: I lean towards yes, personally. But I'm not an immovable object on this, either.

For debugging that wants to see the actual bytes, an IO layer that wraps and logs bytes read and written would allow seeing everything, wouldn't it? At least then, it's explicit that you're dumping all the bytes, instead of how easy it is for someone to just enable trace logging entirely for their application.

carllerche: I think it's probably fine to not dump headers at the server log level... dunno if @olix0r has any thoughts on the matter.

carllerche:
This change was partially pulled in by an earlier commit. I applied the rest of this change here: 0b1fbc4b39bc2aa35cac7e142c025f055afdf5be

Thanks!

PR #31, opened 2017-08-24
Do not reuse next ptr for multiple linked lists

carllerche:
Because, you might think that each linked list has exclusive access to
the next pointer, but then there is an edge case that proves otherwise.
Also, debugging this kind of thing is annoying.

### DIFF
--- a/src/proto/streams/stream.rs
+++ b/src/proto/streams/stream.rs
@@ -83,8 +83,20 @@ pub(super) struct NextSendCapacity;
 pub(super) struct NextWindowUpdate;
 
 impl<B> Stream<B> {
-    pub fn new(id: StreamId) -> Stream<B>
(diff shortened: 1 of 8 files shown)


src/proto/streams/stream.rs
```diff
@@ -8,20 +8,14 @@ pub(super) struct Stream<B> {
...
-    /// Once the stream is accepted, this is used for the linked list of streams
-    /// waiting to flush prioritized frames to the socket.
-    pub next: Option<store::Key>,
-
-    /// Set to true when the stream is queued
-    pub is_queued: bool,
-
     // ===== Fields related to sending =====
 
+    /// Next node in the accept linked list
```
olix0r: this comment is incorrect

PR #41, opened 2017-08-28
Update TLS-related dependencies.

briansmith:
Use a temporary private fork of tokio-rustls that uses Rustls 0.12
until tokio-rustls 0.4 is released.

This upgrades, among other things, *ring* to 0.12, which will ensure
that it still builds in the Rust 1.20 release coming this week even if
backward-compatibility-breaking changes to rustc aren't fixed before
the release.


carllerche:
Thanks. I rebased against master which should pull in the CI fix.

PR #42, opened 2017-08-30
Unbox client handshake future

hawkw:
I've rewritten the client handshake to store the inner future with a named type, rather than as a `Box<Future>`. This also means that the `'static` lifetime bounds on the type parameters `T` and `B` can be removed.

I've ran the tests against this branch and everything passes. Since no new functionality was added, I haven't added any additional tests.

Potential drawbacks of this change are hat it does make some code in `client.rs` a bit more complex, and that the type of the inner future in `Handshake` may need to be updated if certain aspects of the implementation of `Client::handshake2` are changed. However, there isn't currently any other way to remove the `Box` in this situation.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -58,38 +58,31 @@ impl<T, B> Client<T, B>
 
         debug!("binding client connection");
 
-        // this is a function so that we can pass a function pointer (rather
-        // than a closure) to `and_then` in order to refer to the concrete
-        // type of the unboxed handshake future.
-        // see https://tokio.rs/docs/going-deeper-futures/returning/#named-types
-        fn bind<I, U>((io, _): (I, &'static [u8]))
-                     -> Result<Client<I, U>, ConnectionError>
-            where I: AsyncRead + AsyncWrite,
-                  U:  IntoBuf
-        {
-            debug!("client connection bound");
+        let bind: fn((T, &'static [u8]))
+                    -> Result<Client<T, B>, ConnectionError> =
+            |(io, _)| {
+                debug!("client connection bound");
 
-            let mut framed_write = proto::framed_write(io);
-            let settings = frame::Settings::default();
+                let mut framed_write = proto::framed_write(io);
+                let settings = frame::Settings::default();
 
-            // Send initial settings frame
-            match framed_write.start_send(settings.into()) {
-                Ok(AsyncSink::Ready) => {
-                    let connection = proto::from_framed_write(framed_write);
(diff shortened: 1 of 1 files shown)


hawkw:
N.B. that I'd also like to unbox the future in `server::Handshake` as well. The way the `Server::handshake2()` is currently implemented (with a `move` closure that captures the value of `framed_write`) complicates this somewhat.

src/client.rs
```diff
@@ -52,26 +58,40 @@ impl<T, B> Client<T, B>
...
-
-                // Send initial settings frame
-                match framed_write.start_send(settings.into()) {
-                    Ok(AsyncSink::Ready) => {
-                        let connection = proto::from_framed_write(framed_write);
-                        Ok(Client { connection })
-                    }
-                    Ok(_) => unreachable!(),
-                    Err(e) => Err(ConnectionError::from(e)),
+        // this is a function so that we can pass a function pointer (rather
```
seanmonstar: FWIW, [as of Rust 1.19](https://blog.rust-lang.org/2017/07/20/Rust-1.19.html), a closure that doesn't capture over its environment can be cast to a function pointer.

hawkw: Ooh, fancy – I must have missed that change.

src/client.rs
```diff
@@ -5,20 +5,27 @@ use proto::{self, Connection, WindowSize};
...
+            MapErr<WriteAll<T, &'static [u8]>, fn(IoError) -> ConnectionError>,
+            Result<Client<T, B>, ConnectionError>,
+            fn((T, &'static [u8])) -> Result<Client<T, B>, ConnectionError>
+        >
+
 }
 
 /// Marker type indicating a client peer
-pub struct Client<T, B: IntoBuf> {
+pub struct Client<T: AsyncRead + AsyncWrite, B: IntoBuf> {
```
carllerche: Adding this bound is fine w/ me. I'm just curious if this became required due to the change or it was added for some other reason.

carllerche:
I'll merge this after #35 lands. It will be easier to fix any merge conflicts here than the other way around...

src/client.rs
hawkw: If I recall correctly, this bound was necessary for this change in 658c80b6a9535b4a5645b4c29f5275ca516cec28. I'm not sure if it's still necessary as of a67fd958fdc1977c1d4f8438dd2e898cf78680c5?

PR #46, opened 2017-08-31
Improved testing infrastructure

carllerche:
This used to be a testing overhaul PR, but just reorganizing the code + error types turned out to be a large amount of work.

This also introduces a h2-codec crate that exports some private h2 types. This crate will not be published to crates.io (for the immediate future at least). However, it can be used by depending on it from git and is useful for tests.


carllerche:
I'm going to cap this PR as is.

Generally, this does a bunch of refactoring, mostly around error types, but it also paves the way to allow `Codec` to be used standalone.

* `Codec` (and `FramedRead` / `FramedWrite`) is broken out into a `codec` module.
* An h2-codec crate is created that re-exports the `frame` and `codec` modules.
* New error types are introduced in the internals:
  * `RecvError` represents errors caused by trying to receive a frame.
  * `SendError` represents errors caused by trying to send a frame.
  * `UserError` is an enum of potential errors caused by invalid usage by the user of the lib.
  * `ProtoError` is either a `Reason` or an `io::Error`. However it doesn't specify connection or stream level.
  * `h2::Error` is an opaque error type and is the only error type exposed by the public API (used to be `ConnectionError`.

There are misc code changes to enable this as well.

PR #52, opened 2017-09-04
Unbox server handshake future

hawkw:
Server-side version of #42. I've rewritten `server::Handshake` to store the inner future with a named type, rather than as a `Box<Future>`. In addition to removing a `Box`, this also means that the `'static` lifetime bounds on the type parameters `T` and `B` can be removed.

The type of the server handshake future is somewhat more complex than the client-side handshake future. Note also that I've had to re-export `proto::streams::Prioritized` as `pub(crate)` from `proto`, as it appears in the type of the handshake future.

I've ran the tests against this branch and everything passes. Since no new functionality was added, I haven't added any additional tests.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -479,50 +479,38 @@ impl<T, B: IntoBuf> Future for Handshake<T, B>
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         use server::Handshaking::*;
         let result;
+        macro_rules! poll {
+            (|$option:ident, $codec:ident| $body:block) => {{
+                let mut state = $option
+                    .take()
+                    .expect("Handshaking::poll: state option already taken!");
+                match state.poll() {
+                    Ok(Async::Ready($codec)) => { $body },
+                    Ok(Async::NotReady) => {
+                        result = Ok(Async::NotReady);
+                        Handshaking::from(state)
+                    }
+                    Err(e) => {
+                        result = Err(e);
+                        Done
+                    }
+                }
+            }}
+        }
         self.state = match self.state {
-            Flushing(ref mut flush_option) => {
-                let mut flushing = flush_option
-                    .take()
-                    .expect("flush_option already taken!");
-                match flushing.poll() {
-                    Ok(Async::Ready(codec)) => {
-                        let reading = ReadPreface::new(codec);
-                        result = Ok(Async::NotReady);
-                        Handshaking::from(reading)
-                    },
-                    Ok(Async::NotReady) => {
-                        result = Ok(Async::NotReady);
-                        Handshaking::from(flushing)
-                    }
-                    Err(e) => {
-                        result = Err(e);
-                        Done
-                    }
-                }
-            },
-            ReadingPreface(ref mut read_option) => {
-                let mut reading = read_option
-                    .take()
-                    .expect("read_option already taken!");
-                match reading.poll() {
-                    Ok(Async::Ready(codec)) => {
-                        let connection =
-                            Connection::new(codec, &self.settings, 2.into());
-                        result = Ok(Async::Ready(Server { connection }));
-                        Done
-
-                    },
-                    Ok(Async::NotReady) => {
-                        result = Ok(Async::NotReady);
-                        Handshaking::from(reading)
-                    }
-                    Err(e) => {
-                        result = Err(e);
-                        Done
-                    }
-                }
-            },
-
+            Flushing(ref mut flush_option) =>
+                poll!(|flush_option, codec| {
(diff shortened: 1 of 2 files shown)


src/server.rs
```diff
@@ -375,7 +392,7 @@ impl<T, B: IntoBuf> Future for Handshake<T, B> {
 }
 
 impl<T, B> fmt::Debug for Handshake<T, B>
-    where T: fmt::Debug,
+    where T: AsyncRead + AsyncWrite + fmt::Debug,
```
carllerche: I think this bound could be removed as well.

hawkw: The bound is required by `Handshake<T: AsyncRead + AsyncWrite, B: IntoBuf>` now, since the `AsyncRead + AsyncWrite` bound is required by the implementation of `Future` for `ReadPreface` (which is then required by the `ReadPreface` type parameter on `future::AndThen` in `Handshake`.

hawkw: Double-checked this to make sure, and yeah, the compiler will not permit the bound to be removed:
<details>
<summary>
cargo build output
</summary>

```rust
   Compiling http v0.1.0 (https://github.com/carllerche/http?rev=2dd15d9#2dd15d9d)
   Compiling h2 v0.1.0 (file:///Users/eliza/Code/h2)
error[E0277]: the trait bound `T: tokio_io::AsyncRead` is not satisfied
   --> src/server.rs:394:12
    |
394 | impl<T, B> fmt::Debug for Handshake<T, B>
    |            ^^^^^^^^^^ the trait `tokio_io::AsyncRead` is not implemented for `T`
    |
    = help: consider adding a `where T: tokio_io::AsyncRead` bound
    = note: required by `server::Handshake`

error[E0277]: the trait bound `T: tokio_io::AsyncWrite` is not satisfied
   --> src/server.rs:394:12
    |
394 | impl<T, B> fmt::Debug for Handshake<T, B>
    |            ^^^^^^^^^^ the trait `tokio_io::AsyncWrite` is not implemented for `T`
    |
    = help: consider adding a `where T: tokio_io::AsyncWrite` bound
    = note: required by `server::Handshake`

error[E0277]: the trait bound `T: tokio_io::AsyncRead` is not satisfied
   --> src/server.rs:398:5
    |
398 | /     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
399 | |         write!(fmt, "server::Handshake")
400 | |     }
    | |_____^ the trait `tokio_io::AsyncRead` is not implemented for `T`
    |
    = help: consider adding a `where T: tokio_io::AsyncRead` bound
    = note: required by `server::Handshake`

error[E0277]: the trait bound `T: tokio_io::AsyncWrite` is not satisfied
   --> src/server.rs:398:5
    |
398 | /     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
399 | |         write!(fmt, "server::Handshake")
400 | |     }
    | |_____^ the trait `tokio_io::AsyncWrite` is not implemented for `T`
    |
    = help: consider adding a `where T: tokio_io::AsyncWrite` bound
    = note: required by `server::Handshake`

error: aborting due to 4 previous errors

```
</details>

carllerche: Ok... I think it would be worth trying to figure out if there is a route by which the bound could be avoided... I don't know off the top of my head how.

hawkw: I'm pretty sure if we want to have the unboxed future in the `Handshake` struct, there's no way to get rid of the `AsyncRead + AsyncWrite` bound on the struct (without removing it from `impl Future for ReadPreface`, which seems unlikely).

seanmonstar: FWIW, the `client::Handshake` already has these bounds...

carllerche: Ok, I thought it didn't? Maybe it was added more recently.

Anyway, I'm fine w/ this. It would be good to merge. We can audit bounds when we do a pre 0.1 API audit pass.

hawkw:
@carllerche hmm, looks like this no longer works with the `master` version of `Server::handshake2`, since the `settings` argument now has to be passed to `Connection::new`, and the closure now captures scope (so we can't treat it as a fn pointer any longer).

seanmonstar:
It could be written without using the combinators, it would just require `Handshake` to implement the state transitions itself.

hawkw:
@seanmonstar yeah, and I'm happy to look into that; may just take a little bit longer.

carllerche:
Is the plan to pursue or abandon (for now) this change?

hawkw:
@carllerche I'm planning to continue working on this; there are just a couple other tasks up-stack from it at the moment.

PR #53, opened 2017-09-06
Remove AsyncRead + AsyncWrite constraint from h2::Client

olix0r:
This type constraint shouldn't be necessary, and is cumbersome.


carllerche:
Yeah, can merge the fixes.

PR #57, opened 2017-09-06
Make Codec max frame length configurable

hawkw:
There was a TODO comment in `Codec` where the maximum frame length was set saying "make this configurable and allow it to be changed during runtime."

I've added methods `get_max_frame_length` and `set_max_frame_length` on `codec::FramedRead` that wrap the underlying methods of the same name on `tokio_io::length_delimited::FramedRead`, and the same methods to `Codec` wrapping the methods on `codec::FramedRead`.

Added a unit test for this, based on the `update_max_frame_len_at_rest()` test in `tokio_io`. Would like to add a test for updating the max frame length when frames are in flight as well, but it'll take a little more work to mock and given that we're just farming this out to `tokio_io`, it'll _probably_ be fine, right?

### DIFF
--- a/src/codec/mod.rs
+++ b/src/codec/mod.rs
@@ -30,11 +30,11 @@ impl<T, B> Codec<T, B>
     /// Returns a new `Codec` with the default max frame size
     #[inline]
     pub fn new(io: T) -> Self {
-        Self::with_max_frame_length(io, frame::DEFAULT_MAX_FRAME_SIZE as usize)
+        Self::with_max_frame_size(io, frame::DEFAULT_MAX_FRAME_SIZE as usize)
     }
 
-    /// Returns a new `Codec` with the given maximum frame length
-    pub fn with_max_frame_length(io: T, max_frame_length: usize) -> Self {
+    /// Returns a new `Codec` with the given maximum frame size
+    pub fn with_max_frame_size(io: T, max_frame_size: usize) -> Self {
         // Wrap with writer
         let framed_write = FramedWrite::new(io);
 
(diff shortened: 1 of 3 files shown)


src/codec/mod.rs
```diff
@@ -37,15 +44,35 @@ impl<T, B> Codec<T, B>
...
     }
+
+    /// Updates the max frame setting.
+    ///
+    /// The change takes effect the next time a frame is decoded. In other
+    /// words, if a frame is currently in process of being decoded with a frame
+    /// size greater than `val` but less than the max frame length in effect
+    /// before calling this function, then the frame will be allowed.
+    #[inline]
+    pub fn set_max_frame_length(&mut self, val: usize) {
```
carllerche: Could this fn be moved to the impl block below (without the trait bounds) and be named `set_max_recv_frame_size` (to match the pattern for sending)?

I used _size as a suffix to match the terminology in the HTTP/2.0 spec.

src/codec/mod.rs
```diff
@@ -37,15 +44,35 @@ impl<T, B> Codec<T, B>
...
+        self.inner.set_max_frame_length(val)
+
+    }
+
+    /// Returns the current max frame setting
+    ///
+    /// This is the largest size this codec will accept from the wire. Larger
+    /// frames will be rejected.
+    #[inline]
+    pub fn max_frame_length(&self) -> usize {
```
carllerche: Could this fn be moved to the impl block below (without the trait bounds) and be named `max_recv_frame_size` (to match the pattern for sending)?

I used _size as a suffix to match the terminology in the HTTP/2.0 spec.

src/codec/mod.rs
hawkw: Ah, sure, this naming scheme was based on `FramedRead`  I'll change that.

PR #58, opened 2017-09-06
Add codecov.io coverage reporting

hawkw:
I've added an `after_success` step to `.travis.yml` to report code coverage to codecov.io.

Since I'm not a repo owner, I think somebody else will have to actually configure the repository on codecov.


carllerche:
Is there a link for more info? What env var is the upload token supposed to be set to?

carllerche:
I guess we will find out if it works...

PR #61, opened 2017-09-07
workaround for travis-cargo not passing the right flags to kcov

hawkw:
@carllerche, let me know if this works right?

I also added slack webhooks to `codecov.yml`.


hawkw:
Okay, looks like a7402b3 did not fix it, I'm not really sure what's going on here. Might be an issue on Codecov's end?

carllerche:
Superseded by ##62

PR #59, opened 2017-09-06
Fix codecov

carllerche:

PR #62, opened 2017-09-07
Split Travis build into stages

hawkw:
This PR rewrites our Travis config to use Travis' [Build Stages](https://docs.travis-ci.com/user/build-stages/) feature. It also includes my PR with fixes to codecov.io coverage uploading (#61). I've rewritten the step for publishing RustDoc to GitHub Pages to use a Travis Deploy stage, removing the dependency on `travis-cargo`.

Closes #59 
Closes #61


hawkw:
Okay, so I have code coverage working finally using `tarpaulin`, but it looks like running the hpack tests with coverage enabled takes so long that Travis kills the build. I suspect this would have occurred with `kcov` as well, if it didn't segfault. Will try and see if I can configure `tarpaulin` to skip those tests...

hawkw:
see xd009642/tarpaulin#32

hawkw:
thanks to xd00942/tarpaulin@8e3fe3a8c907260185adf43a09965314760206b9, coverage may actually work now (fingers crossed!)

carllerche:
Looks like stuff is happening... What are the next steps?

hawkw:
Okay @carllerche, I think this will be ready to merge after [build #320](https://travis-ci.org/carllerche/h2/builds/275111815) is done.

I've disabled the codecov.io bot comment (since it's configured in YAML), but I haven't disabled the coveralls bot, since that has to be done in the repo's settings on Coveralls (which I don't have access to).

carllerche:
Looks good. We probably should stick w/ one option to limit noise. Could you remove coveralls?

.ci/coverage.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# this is in a separate script from `.travis.yml` so we can set this option.
+shopt -s extglob
```
olix0r: always `set -eu` unless it's really clear that this isn't necessary.

.ci/coverage.sh
```diff
@@ -0,0 +1,23 @@
...
+
+wget https://github.com/SimonKagstrom/kcov/archive/master.tar.gz &&
+tar xzf master.tar.gz &&
+cd kcov-master &&
+mkdir build &&
+cd build &&
+cmake .. &&
+make &&
+make install DESTDIR=../tmp &&
+cd ../..
```
olix0r: if you use `set -e`, don't need these long `&&` chains

.ci/coverage.sh
```diff
@@ -0,0 +1,23 @@
...
+
+if [ $TRAVIS_RUST_VERSION = nightly ]; then
+    ./kcov-master/tmp/usr/local/bin/kcov --exclude-pattern=/.cargo --coveralls-id=$TRAVIS_JOB_ID --verify  target/kcov target/debug/*-*[!.]? &&
```
olix0r: be careful to quote "$TRAVIS_JOB_ID" so it can't be used to inject nefarious things ;)

(PSA `shellcheck` is a great tool that will complain about this sort of thing

.ci/coverage.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# this is in a separate script from `.travis.yml` so we can set this option.
+shopt -s extglob
+
+wget https://github.com/SimonKagstrom/kcov/archive/master.tar.gz &&
+tar xzf master.tar.gz &&
```
olix0r: just so i have context...

will this all run on every commit?  can we cache the install directory via travis so that we only need to build if the cache is gone?  If so, this should probably be guarded by ` if [ -x tmp/kvoc ] ... ` or something?

.ci/coverage.sh
hawkw: this script is actually no longer used – I should rm it

.ci/coverage.sh
hawkw: can I `brew install shellcheck`?

hawkw:
@carllerche / @olix0r, this should be ready to merge now?

carllerche:
Yes! Thanks for putting in the work and getting this done.

I'll merge and any further tweaks can happen separately.

carllerche:
I merged in master, I guess it has to go through CI again!

hawkw:
It's really strange to me that 1c2b030 broke this but 3abc489 built fine?

PR #64, opened 2017-09-07
add ergonomics to testing with frames

seanmonstar:
- Add `support::frames` builders, to ease creating frames to compare with or send on the mocked IO.
- Add extension methods for futures that return a `Handle`, such as `send_frame` and `recv_frame`.

Ends up looking something like this:

```rust
let mock = srv.assert_client_handshake().unwrap()
    .recv_settings()
    .recv_frame(
        frames::headers(1)
            .request("GET", "https://http2.akamai.com/")
            .eos()
    )
    .send_frame(
        frames::headers(1)
            .response(200)
    )
    // send 2 data frames
    .send_frame(frames::data(1, &payload[..]))
    .send_frame(frames::data(1, &payload[..]).eos())
    // gotta end the connection
    .map(drop);
```

### DIFF
--- a/tests/flow_control.rs
+++ b/tests/flow_control.rs
@@ -57,84 +57,39 @@ fn release_capacity_sends_window_update() {
 
     let (io, srv) = mock::new();
 
-    fn make_request() -> Request<()> {
-        Request::builder()
-            .method(Method::GET)
-            .uri("https://http2.akamai.com/")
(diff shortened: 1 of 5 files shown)


carllerche:
Tests look much nicer.

It's a bit late for me to digest all the type related implementation detail, but merge it if you want to. We can always iterate on master.

PR #67, opened 2017-09-08
increment stream window capacity when sending window update

olix0r:
Fixes https://github.com/carllerche/h2/issues/65

### DIFF
--- a/tests/flow_control.rs
+++ b/tests/flow_control.rs
@@ -54,7 +54,7 @@ fn send_data_without_requesting_capacity() {
 fn release_capacity_sends_window_update() {
     let _ = ::env_logger::init();
 
-    let payload = vec![0u8; 65_535];
+    let payload = vec![0u8; 16384];
 
     let (io, srv) = mock::new();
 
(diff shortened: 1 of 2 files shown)


carllerche:
Looks good, it would be worth adding a test to avoid regressions :)

tests/flow_control.rs
```diff
@@ -69,19 +69,17 @@ fn release_capacity_sends_window_update() {
...
-        .send_frame(frames::data(1, &payload[0..16_384]))
-        .send_frame(frames::data(1, &payload[16_384..16_384 * 2]))
-        .send_frame(frames::data(1, &payload[16_384 * 2..16_384 * 3]))
+        .send_frame(frames::data(1, &payload[..]))
+        .send_frame(frames::data(1, &payload[..]))
+        .send_frame(frames::data(1, &payload[..]))
+        .send_frame(frames::data(1, &payload[1..]))
         .recv_frame(
-            frames::window_update(0, 32_768)
+            frames::window_update(0, 65_535)
```
seanmonstar: This test was checking that window updates are sent if the difference is over the ratio. The modifications to this test mean it's no longer testing that specifically, right?

This is testing that after the full window is used, sending an update of the initial window size correctly allows more data?

PR #68, opened 2017-09-08
check for StreamId overflow

### LINKED ISSUE #48: handle stream id overflow
[RFC7540§5.1.1 Stream Identifiers](http://httpwg.org/specs/rfc7540.html#StreamIdentifiers) says:

> Stream identifiers cannot be reused. Long-lived connections can result in an endpoint exhausting the available range of stream identifiers. A client that is unable to establish a new stream identifier can establish a new connection for new streams. A server that is unable to establish a new stream identifier can send a GOAWAY frame so that the client is forced to open a new connection for new streams.

On the client side, we should return some sort of refusal (either through a `poll_ready()` with a dedicated failure type or with an error type that returns the `Request<T>` to the application).

On the server side, we'll need to send a GOAWAY with a last stream id of the maximum stream id.

seanmonstar:
Closes #48 

No tests... With #64, it'd be pretty easy to add a test to stream_states when a peer tries to open a new stream once the id space is at max. Adding a test that a `UserError` is returned will require adding some way to configure the next stream ID manually, or we'd have a test case submitting some 2 million requests...

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -15,8 +15,8 @@ use std::marker::PhantomData;
 
 /// In progress H2 connection binding
 pub struct Handshake<T: AsyncRead + AsyncWrite, B: IntoBuf = Bytes> {
+    builder: Builder,
     inner: MapErr<WriteAll<T, &'static [u8]>, fn(io::Error) -> ::Error>,
-    settings: Settings,
     _marker: PhantomData<B>,
 }
 
@@ -36,9 +36,10 @@ pub struct Body<B: IntoBuf> {
 }
 
 /// Build a Client.
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 pub struct Builder {
     settings: Settings,
+    stream_id: StreamId,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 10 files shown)


carllerche:
A test AND a benchmark...

carllerche:
This should ideally get merged in. I think it's been pending on a test.

Maybe we could add a configuration setting to the builders to specify the starting `StreamId` and flag that with `unstable`.

seanmonstar:
I added an unstable client builder method to change the starting stream ID, and a test that when making requests and the send stream id overflows, the user gets a user error back.

carllerche:
Given that there is a real case for configuring the initial stream ID (HTTP 1.1 upgrade), I think it would be reasonable to have the config option be generally available (not unstable).

seanmonstar:
> Given that there is a real case for configuring the initial stream ID (HTTP 1.1 upgrade)

Yes, but after thinking about it some more, I'm not sure this is the correct API to expose it to a user. For a client, they would want to do `Client::upgrade(io)` or similar, which means that stream ID of 1 is local-half-closed, and a response should be coming on stream 1. A server likewise would probably want to do something like `Server::upgrade(io)`.

The spec does say that an HTTP/1 upgrade uses stream ID 1, not any other, so it could be enforced in h2. (However, internally we do need to be able to say that the next available client stream ID is 3, so this still has use besides this 1 test.)

src/client.rs
```diff
@@ -182,6 +183,14 @@ impl Builder {
         self
     }
 
+    /// Set the first stream ID to something other than 1.
+    #[cfg(feature = "unstable")]
+    pub fn starting_stream_id(&mut self, stream_id: u32) -> &mut Self {
```
carllerche: A pretty dumb bikeshed... but maybe `initial` instead of `starting` since that's the lingo everything else uses.

PR #71, opened 2017-09-08
Live the rustfmt lifestyle

olix0r:
This change applies `rustfmt` on the entire repo, and updates CI to check formatting after running tests.  We use an explicit, committed rustfmt.toml to enforce a consistent style.  The only modification from the default style, at this point, is to set `closure_block_indent_threshold = -1` (per prior discussion with @carllerche).

I am happy to rebase ongoing work onto this as it lands, but I'd like feedback on if there are any settings that absolutely drive you crazy to the point where it will be hard for you to work in the repo.  Modulo that, I'd like to pick a configuration and roll with it.


examples/client.rs
```diff
@@ -49,52 +49,51 @@ impl Future for Process {
...
+    let tcp = TcpStream::connect(&"127.0.0.1:5928".parse().unwrap(), &handle);
 
     let tcp = tcp.then(|res| {
         let tcp = io_dump::Dump::to_stdout(res.unwrap());
         Client::handshake(tcp)
-    })
-    .then(|res| {
-        let mut client = res.unwrap();
+    }).then(|res| {
+            let mut client = res.unwrap();
```
seanmonstar: 8 space indent here?

src/client.rs
```diff
@@ -261,14 +256,16 @@ impl proto::Peer for Peer {
...
         use http::request::Parts;
 
-        let (Parts { method, uri, headers, .. }, _) = request.into_parts();
+        let (Parts {
+                 method,
+                 uri,
+                 headers,
+                 ..
+             },
+             _) = request.into_parts();
```
seanmonstar: This looks pretty weird...

src/codec/error.rs
```diff
@@ -6,10 +6,7 @@ use std::{error, fmt, io};
 #[derive(Debug)]
 pub enum RecvError {
     Connection(Reason),
-    Stream {
-        id: StreamId,
-        reason: Reason,
-    },
+    Stream { id: StreamId, reason: Reason },
```
seanmonstar: In cases like this, having the fields on separate lines means better diffs/blames...

src/frame/mod.rs
```diff
@@ -52,13 +52,8 @@ pub use self::stream_id::StreamId;
 pub use self::window_update::WindowUpdate;
 
 // Re-export some constants
-pub use self::settings::{
-    DEFAULT_SETTINGS_HEADER_TABLE_SIZE,
-    DEFAULT_INITIAL_WINDOW_SIZE,
-    DEFAULT_MAX_FRAME_SIZE,
-    MAX_INITIAL_WINDOW_SIZE,
-    MAX_MAX_FRAME_SIZE,
-};
```
seanmonstar: I know its not that common, but I personally prefer the imports on separate lines like this. The same line way gets weird and uglier as you start needing new lines eventually, but then with odd alignment indenting... And diffs.

src/frame/reason.rs
```diff
@@ -22,8 +22,12 @@ pub enum Reason {
 
 // ===== impl Reason =====
 
+const CONNECT_ERROR_MSG: &'static str = {
+    "connection established in response to a CONNECT request was reset or abnormally closed"
+};
```
seanmonstar: A block containing just the static string? Seems odd...

src/hpack/decoder.rs
```diff
@@ -474,8 +476,9 @@ impl Table {
         debug_assert!(size <= self.max_size);
 
         while self.size + size > self.max_size {
-            let last = self.entries.pop_back()
-                .expect("size of table != 0, but no headers left!");
+            let last = self.entries.pop_back().expect(
```
seanmonstar: Hm? Is this because the line is over the line length?

olix0r: yep

src/codec/error.rs
hawkw: if we really cared about having better diffs/blames, we'd use comma-first style... 😜

/me is forcibly removed from the repo contributors

src/hpack/huffman/mod.rs
```diff
@@ -185,7 +180,8 @@ mod test {
             "cookie",
             "B=76j09a189a6h4&b=3&s=0b",
             "TE",
-            LOREM_IPSUM,
+            "Lorem ipsum dolor sit amet, consectetur adipiscing elit. \
+             Morbi non bibendum libero. Etiam ultrices lorem ut.",
```
hawkw: 🙌🙌🙌

src/codec/error.rs
seanmonstar: I feel comma-first became a thing because of languages that don't allow trailing commas. Rust doesn't suffer from this fault. 😁

src/frame/mod.rs
olix0r: as far as i can tell rustfmt doesn't offer a way to configure this at the moment. is this a blocker?

src/frame/reason.rs
olix0r: yep, fixed

src/codec/error.rs
olix0r: this is [controlled by a length](https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md#struct_variant_width).  Is there consensus about this _always_ being split?

src/hpack/decoder.rs
olix0r: fixed this by changing `fn_call_style`

olix0r:
```diff --git a/.rustfmt.toml b/.rustfmt.toml
index e99c7d5..d7eb57d 100644
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -31,15 +31,15 @@ where_pred_indent = "Visual"
 generics_indent = "Block"
 struct_lit_style = "Block"
 struct_lit_multiline_style = "PreferSingle"
-fn_call_style = "Block"
+fn_call_style = "Visual"
 report_todo = "Never"
 report_fixme = "Never"
 chain_indent = "Block"
 chain_one_line_max = 60
 chain_split_single_child = false
-reorder_imports = false
-reorder_imports_in_group = false
-reorder_imported_names = false
+reorder_imports = true
+reorder_imports_in_group = true
+reorder_imported_names = true
 single_line_if_else_max_width = 50
 format_strings = false
 force_format_strings = false
@@ -62,7 +62,7 @@ spaces_around_ranges = false
 spaces_within_angle_brackets = false
 spaces_within_square_brackets = false
 spaces_within_parens = false
-use_try_shorthand = false
-write_mode = "Replace"
+use_try_shorthand = true
+write_mode = "Overwrite"
 condense_wildcard_suffixes = false
 combine_control_expr = true```

File: src/frame/mod.rs (hunk 4 above)
Comment (olix0r) [reply src/frame/mod.rs]: oh, nope. found a setting

**Comment by olix0r** [comment]
I think I've addressed _most_ of the actionable notes.  We should figure out if we want to set `struct_variant_width`.

Also one wrinkle: in order to configure import_layouts and import_indent (per @seanmonstar's request), we must rely on rustfmt-nightly (which requires a nightly toolchain).  This may be a minor pain in devel, but CI checks it, at least...

File: src/codec/error.rs (hunk 3 above)
Comment (hawkw) [reply src/codec/error.rs]: @seanmonstar trailing commas are ugly

Comment (hawkw) [reply src/codec/error.rs]: (comma first style is somehow not ugly because I'm Smart and Always Right)

File: examples/client.rs (hunk 1 above)
Comment (carllerche) [reply examples/client.rs]: It lines it up w/ the pipe, which is really weird. I don't like it :(

File: src/codec/error.rs (hunk 3 above)
Comment (carllerche) [reply src/codec/error.rs]: @olix0r In type definitions for structs, I like always breaking it up across lines.

**Comment by carllerche** [comment]
@olix0r unfortunately, nightly is the most supported version.

File: src/codec/error.rs (hunk 3 above)
Comment (olix0r) [reply src/codec/error.rs]: changing

```diff
-struct_lit_multiline_style = "PreferSingle"
+struct_lit_multiline_style = "ForceMulti"
```
pick your poison:

```diff
diff --git a/src/client.rs b/src/client.rs
index ed7b343..7bfab0e 100644
--- a/src/client.rs
+++ b/src/client.rs
@@ -79,7 +79,9 @@ where
             match codec.start_send(settings.into()) {
                 Ok(AsyncSink::Ready) => {
                     let connection = Connection::new(codec);
-                    Ok(Client { connection })
+                    Ok(Client {
+                           connection,
+                       })
                 }
                 Ok(_) => unreachable!(),
                 Err(e) => Err(::Error::from(e)),
@@ -91,7 +93,9 @@ where
             .map_err(::Error::from as fn(IoError) -> ::Error)
             .and_then(bind);
 
-        Handshake { inner: handshake }
+        Handshake {
+            inner: handshake,
+        }
     }
 
     /// Returns `Ready` when the connection can initialize a new HTTP 2.0
@@ -109,7 +113,11 @@ where
         self.connection
             .send_request(request, end_of_stream)
             .map_err(Into::into)
-            .map(|stream| Stream { inner: stream })
+            .map(|stream| {
+                Stream {
+                    inner: stream,
+                }
+            })
     }
 }
 
diff --git a/src/codec/error.rs b/src/codec/error.rs
index 1713e65..33c753f 100644
--- a/src/codec/error.rs
+++ b/src/codec/error.rs
@@ -53,7 +53,9 @@ impl error::Error for RecvError {
 
         match *self {
             Connection(ref reason) => reason.description(),
-            Stream { ref reason, .. } => reason.description(),
+            Stream {
+                ref reason, ..
+            } => reason.description(),
             Io(ref e) => e.description(),
         }
     }
diff --git a/src/codec/mod.rs b/src/codec/mod.rs
index 8d81150..01db79b 100644
--- a/src/codec/mod.rs
+++ b/src/codec/mod.rs
@@ -50,7 +50,9 @@ where
 
         let inner = FramedRead::new(delimited);
 
-        Codec { inner }
+        Codec {
+            inner,
+        }
     }
 }
 ```

examples/client.rs
steveklabnik: yeah this seems like a bug

PR #73, opened 2017-09-08
Ref count stream state and release when final.

carllerche:
**Work in progress**

Try to actually free memory...

### DIFF
--- a/src/proto/streams/store.rs
+++ b/src/proto/streams/store.rs
@@ -2,8 +2,9 @@ use super::*;
 
 use slab;
 
+use ordermap::{self, OrderMap};
+
 use std::ops;
-use std::collections::{HashMap, hash_map};
 use std::marker::PhantomData;
 
 /// Storage for streams
@@ -12,7 +13,7 @@ pub(super) struct Store<B, P>
     where P: Peer,
 {
     slab: slab::Slab<Stream<B, P>>,
-    ids: HashMap<StreamId, usize>,
+    ids: OrderMap<StreamId, usize>,
 }
 
 /// "Pointer" to an entry in the store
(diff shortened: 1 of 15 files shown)


src/proto/streams/stream.rs
```diff
@@ -1,5 +1,18 @@
...
+///
+/// # Reference counting
+///
+/// There can be a number of outstanding handles to a single Stream. These are
+/// tracked using reference counting. The `ref_count` field represents the
+/// number of outstanding userspace handles that can reach this stream.
+///
+/// It's important to note that when the stream is placed in an internal queue
+/// (such as an accept queue), this is **not** tracked by a reference count.
+/// Thus, `ref_count` can be zero and the stream still has to be kept around.
```
hawkw: ty for rustdoc :)

carllerche:
Another option besides ref counting would be for `StreamRef` to store both the key (index in the slab) and the stream identifier. Since the stream identifier is monotonically assigned, the `StreamRef` deref fn would lookup the stream state from the slab then ensure that the IDs match.

The downside is that stream reset on drop is harder, but that might be possible to implement statically (by managing type transitions).

hawkw:
@carllerche I, for one, am extremely in favor of representing this w/ type transitions.

src/proto/streams/store.rs
```diff
@@ -170,6 +200,19 @@ impl<B, P> ops::IndexMut<Key> for Store<B, P>
     }
 }
 
+#[cfg(feature = "unstable")]
+impl<B, P> Store<B, P>
```
olix0r: This appears not to work for me. When compiling a project that depends on this:

```
error[E0599]: no method named `num_active_streams` found for type `proto::streams::store::Store<B, P>` in the current scope
   --> /home/ver/.cargo/git/checkouts/h2-5627e19d2c95dbee/4b4aeef/src/proto/streams/streams.rs:440:18
    |
440 |         me.store.num_active_streams()
    |                  ^^^^^^^^^^^^^^^^^^

error[E0599]: no method named `num_wired_streams` found for type `proto::streams::store::Store<B, P>` in the current scope
   --> /home/ver/.cargo/git/checkouts/h2-5627e19d2c95dbee/4b4aeef/src/proto/streams/streams.rs:445:18
    |
445 |         me.store.num_wired_streams()
    |                  ^^^^^^^^^^^^^^^^^
```

src/proto/streams/streams.rs
```diff
@@ -384,6 +430,22 @@ impl<B> Streams<B, client::Peer>
     }
 }
 
+// #[cfg(feature = "unstable")]
```
olix0r: this needs to be uncommented i think

Issue #75, opened 2017-09-09
API polish


carllerche:
* `Client::request` should be renamed `Client::send_request`
* `send_trailers` and `send_reset` should take `self`
* `Client` is missing `send_reset`
* Dropping the read body should reset the stream.

hawkw:
@carllerche w.r.t the "`send_trailers` and `send_reset` should take `self`" point, is this the `send_{trailers, reset}` method on `StreamRef`, `Streams`, `proto::Send`, `server::Stream`, or all of the above?

seanmonstar:
A possiblity is for the `Stream` to implement `Sink`, and so all actions of sending new frames are down on the `Stream`, instead of the `Client`. That'd mean you don't need to remember the stream ID, and wouldn't need to keep both types in scope, as all sending could be done on the `Stream`. And then `shutdown` could send a reset.

carllerche:
So, I think that we determined that `send_trailers` and `send_reset` cannot take `self` because you can send those frames and then use `self` to receive the response.

carllerche:
I'm also going to break this issue up into individual issues for better tracking.

carllerche:
Ok, I've split this issue out to individual items. I'm going to close this issue now.

PR #76, opened 2017-09-10
Factor out reset on RecvError::Stream

hawkw:
This is a pretty minor change, but I was working on #75 & noticed the TODO comment. I extracted the code for resetting a stream on `RecvError::Stream` & moved it to a function on `Actions`.

Notes:
+ I briefly considered making the function on `store::Ptr`(so it could be invoked as `stream.reset_on_recv_stream_err(actions, result)`, which seems somewhat more semantically correct, but putting the function on `Ptr` didn't seem to be as good a separation of concerns.
+ Method name feels too long, but I couldn't come up with a shorter one that was descriptive of what the method does...

### DIFF
--- a/src/proto/streams/streams.rs
+++ b/src/proto/streams/streams.rs
@@ -114,15 +114,7 @@ impl<B, P> Streams<B, P>
                 actions.recv.recv_trailers(frame, stream)
             };
 
-            // TODO: extract this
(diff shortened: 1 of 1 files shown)


hawkw:
@carllerche, merge conflicts fixed in 9c5edb4

PR #78, opened 2017-09-11
Fix send flow control bug

carllerche:
Pending test.


carllerche:
@olix0r I removed the comment & created #84 to track adding docs. I'm trying to get this PR in before the big #71 :)

Issue #40, opened 2017-08-27
Allow application to configure initial window size, etc


olix0r:
gRPC clients, for instance, set an initial window of ~1MB. We should have a way to configure window sizes on clients and servers (i.e. esp prior to handshake).

seanmonstar:
Both the Client and the Server now have builders that can currently configure:

- `initial_window_size`
- `max_frame_size`

seanmonstar:
A client can now disable push promises.

I think that perhaps we can close this, and any individual settings required can be separate issues, since they aren't trivial things to add.

Issue #32, opened 2017-08-24
Account for invalid frames against connection flow control window


carllerche:
> A receiver that receives a flow-controlled frame MUST always account for its contribution against the connection flow-control window, unless the receiver treats this as a connection error (Section 5.4.1). This is necessary even if the frame is in error. The sender counts the frame toward the flow-control window, but if the receiver does not, the flow-control window at the sender and receiver can become different.

If a RST_STREAM frame is sent, any data frames received on this stream must be accounted for against the connection level flow control window.

carllerche:
The code should be audited for any error that results in an stream level reset.

PR #79, opened 2017-09-11
Intial Client Configuration

seanmonstar:
Only started with a single setting, `initial_window_size`, instead of offer a bunch that aren't actually hooked up. As we want more (and we do), we will have to be sure that the setting actually does what it says.

While I was working on this, and coming up with a test case to make sure it worked, I ended up fixing some other things too... Oh well, they're broken into logical commits.

cc #40 (server config still missing)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -48,7 +48,6 @@ impl<T> Client<T, Bytes>
 where
     T: AsyncRead + AsyncWrite,
 {
-
     /// Bind an H2 client connection.
     ///
     /// Returns a future which resolves to the connection value once the H2
@@ -69,8 +68,9 @@ impl Client<(), Bytes> {
 }
 
 impl<T, B> Client<T, B>
-where T: AsyncRead + AsyncWrite,
-      B: IntoBuf
+where
+    T: AsyncRead + AsyncWrite,
+    B: IntoBuf,
 {
     fn handshake2(io: T, settings: Settings) -> Handshake<T, B> {
         use tokio_io::io;
@@ -78,8 +78,7 @@ where T: AsyncRead + AsyncWrite,
         debug!("binding client connection");
 
         let msg: &'static [u8] = b"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n";
-        let handshake = io::write_all(io, msg)
-            .map_err(::Error::from as _);
+        let handshake = io::write_all(io, msg).map_err(::Error::from as _);
 
         Handshake {
             inner: handshake,
@@ -179,8 +178,9 @@ impl Builder {
     /// It's important to note that this does not **flush** the outbound
     /// settings to the wire.
     pub fn handshake<T, B>(&self, io: T) -> Handshake<T, B>
-        where T: AsyncRead + AsyncWrite,
-              B: IntoBuf
+    where
+        T: AsyncRead + AsyncWrite,
+        B: IntoBuf,
     {
         Client::handshake2(io, self.settings.clone())
     }
@@ -204,8 +204,9 @@ where
         let mut codec = Codec::new(io);
 
         // Send initial settings frame
-        codec.buffer(self.settings.clone().into())
-           .expect("invalid SETTINGS frame");
+        codec
+            .buffer(self.settings.clone().into())
+            .expect("invalid SETTINGS frame");
 
         let connection = Connection::new(codec, &self.settings);
         Ok(Async::Ready(Client {
--- a/src/client.rs
+++ b/src/client.rs
@@ -4,21 +4,20 @@ use frame::Reason::*;
 use proto::{self, Connection, WindowSize};
 
 use bytes::{Bytes, IntoBuf};
-use futures::{AndThen, Async, AsyncSink, Future, MapErr, Poll, Sink};
+use futures::{Async, Future, MapErr, Poll};
 use http::{HeaderMap, Request, Response};
 use tokio_io::{AsyncRead, AsyncWrite};
 use tokio_io::io::WriteAll;
 
 use std::fmt;
-use std::io::Error as IoError;
+use std::io;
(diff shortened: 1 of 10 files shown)


carllerche:
It looks like this will also fix #79

src/client.rs
```diff
@@ -38,14 +35,20 @@ pub struct Body<B: IntoBuf> {
     inner: proto::StreamRef<B::Buf, Peer>,
 }
 
+/// Configure a Client from the defaults.
+#[derive(Debug, Default)]
+pub struct Config {
```
carllerche: I would consider renaming this to `Builder` as this looks more like a builder pattern.

src/client.rs
```diff
@@ -60,41 +63,22 @@ impl<T, B> Client<T, B>
     ///
     /// It's important to note that this does not **flush** the outbound
     /// settings to the wire.
-    pub fn handshake2(io: T) -> Handshake<T, B> {
+    pub fn handshake2(io: T, config: Config) -> Handshake<T, B> {
```
carllerche: I think I would either call it `Config` and pass it as an arg here (and remove `handshake2` on `Config`) **or** call it `Builder` and remove `pub` from this fn.

I think that I prefer the builder approach personally.

src/client.rs
```diff
@@ -173,7 +157,25 @@ where T: AsyncRead + AsyncWrite {
...
+        debug!("client connection bound");
+
+        // Create the codec
+        let mut codec = Codec::new(io);
+
+        // Create the initial SETTINGS frame
+        let settings = self.config.settings.clone();
+
+        // Send initial settings frame
+        match codec.start_send(settings.clone().into()) {
```
carllerche: Since you are here, could you switch this to use [`Codec::buffer`](https://github.com/carllerche/h2/blob/9448a194081e943e5768bc874d39aa7fe842196a/src/codec/mod.rs#L129) instead.

src/client.rs
```diff
@@ -274,6 +276,31 @@ impl<B: IntoBuf> ::futures::Stream for Body<B> {
...
+// ===== impl Config =====
+
+impl Config {
+    /// Set the initial window size of the remote peer.
+    pub fn initial_window_size(&mut self, size: u32) -> &mut Self {
+        self.settings.set_initial_window_size(Some(size));
+        self
+    }
+
+    pub fn handshake<T>(&self, io: T) -> Handshake<T, Bytes>
```
carllerche: I wonder if this fn should be removed and `handshake2` renamed to `handshake`. Then `Client::handshake` could be the "simple" option.

src/proto/connection.rs
```diff
@@ -56,13 +56,14 @@ impl<T, P, B> Connection<T, P, B>
           P: Peer,
           B: IntoBuf,
 {
-    pub fn new(codec: Codec<T, Prioritized<B::Buf>>) -> Connection<T, P, B> {
+    pub fn new(codec: Codec<T, Prioritized<B::Buf>>, settings: frame::Settings) -> Connection<T, P, B> {
```
carllerche: This could probably be passed in by ref and it would avoid the clone / move from above.

src/proto/streams/recv.rs
```diff
@@ -262,10 +264,21 @@ impl<B, P> Recv<B, P>
...
         // on the stream.
-        if self.flow.window_size() < sz || stream.recv_flow.window_size() < sz {
+        if self.flow.window_size() < sz {
             return Err(RecvError::Connection(FlowControlError));
         }
 
+        // The spec gives us the option of returning either a connection error
+        // or stream error when a stream overflows its window (but not the
+        // connection). So, we opt to send a stream error.
+        if stream.recv_flow.window_size() < sz {
```
carllerche: If you use a stream level error here, then `sz` has to be applied against the connection level flow control.

> A receiver that receives a flow-controlled frame MUST always account for its contribution against the connection flow-control window, unless the receiver treats this as a connection error (Section 5.4.1). This is necessary even if the frame is in error. The sender counts the frame toward the flow-control window, but if the receiver does not, the flow-control window at the sender and receiver can become different.

This relates to #32 and generally speaking there probably are existing bugs related to this in the repo.

src/proto/streams/recv.rs
```diff
@@ -562,17 +567,27 @@ impl<B, P> Recv<B, P>
...
+    fn schedule_recv<T>(&mut self, stream: &mut Stream<B, P>)
+        -> Poll<Option<T>, proto::Error>
+    {
+        match stream.state.ensure_recv_open() {
+            Ok(()) => {
+                // Request to get notified once more frames arrive
+                stream.recv_task = Some(task::current());
+                Ok(Async::NotReady)
+            },
+            Err(proto::Error::Proto(Reason::NoError)) => {
```
carllerche: Could you explain this logic? I'm not sure I follow when `ensure_recv_open` would return `NoError` and why this is mapped as an end of stream.

seanmonstar: This would return `NoError` when either half closed, or closed with no error. This is treating the `NoError` variant similar to how IO treats the `WouldBlock` variant: for the purposes of this function, it's not an error, but means the stream is closed.

Alternatively, I can change `ensure_recv_open` to return a `Poll<(), Error>`, and convert `NoError` into an `Ok(NotReady)` in that function, and just pass it along here.

seanmonstar:
Comments addressed, replied about `ensure_recv_open`, rebased with rustfmt changes.

carllerche:
Ack, sorry... looks like I merged your rustfmt PR first :'(

seanmonstar:
Updated with new formatting again.

seanmonstar:
Adjusted `ensure_recv_open` to not return `NoError` when half closed or organically closed, instead returning `Ok(false)`.

PR #85, opened 2017-09-13
rustfmt: add trailing commas in match arms, set fn call to block style

seanmonstar:
I didn't notice these in the big rustfmt PR. I think both of these settings are nicer:

- `fn_call_style = "Block"` means much less indenting. Especially nice when the function name is already long, and then the argument names are also long, and it's super cramped at the right of the window.
- `match_block_trailing_comma = true` means we just always have trailing commas, and don't have to think about if we need them because there's `{}` or not.


src/frame/reset.rs
```diff
@@ -32,15 +32,17 @@ impl Reset {
...
 
     pub fn encode<B: BufMut>(&self, dst: &mut B) {
-        trace!("encoding RESET; id={:?} code={}",
-               self.stream_id,
-               self.error_code);
+        trace!(
+            "encoding RESET; id={:?} code={}",
+            self.stream_id,
+            self.error_code
+        );
```
olix0r: fwiw, i changed fn_call_style so that cases like this wouldn't change.

carllerche:
Personally I don't like match_block_trailing_comma (more chars), but I also don't care much

PR #86, opened 2017-09-13
Notify connection on connection window expansion

carllerche:
When capacity is released back to the connection and a connection level
window update needs to be sent out, the connection task needs to be
notified in order for the send to actually happen.


olix0r:
:+1: lgtm.  but, yeah, better with tests

PR #88, opened 2017-09-13
Configure max frame size

seanmonstar:
- Adds `max_frame_size` to client and server builders
- Pushes max_frame_size into Codec
- Detects when the Codec triggers an error from a frame too big
- Sends a GOAWAY when FRAME_SIZE_ERROR is encountered reading a frame

Depends on #87

### DIFF
--- a/src/codec/framed_read.rs
+++ b/src/codec/framed_read.rs
@@ -1,6 +1,6 @@
 use codec::RecvError;
-use frame::{self, Frame, Kind};
-use frame::DEFAULT_SETTINGS_HEADER_TABLE_SIZE;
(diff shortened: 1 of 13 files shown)


src/codec/framed_read.rs
```diff
@@ -258,3 +263,15 @@ where
         }
     }
 }
+
+fn map_err(err: io::Error) -> RecvError {
+    use std::error::Error;
+
+    if let io::ErrorKind::InvalidData = err.kind() {
+        // woah, brittle...
```
carllerche: Could you add a comment referencing the issue in tokio-io to fix this?

PR #91, opened 2017-09-14
add Client config to disable server push

seanmonstar:
- Adds `Client::builder().enable_push(false)` to disable push
- Client sends a GO_AWAY if receiving a push when it's disabled

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -176,6 +176,12 @@ impl Builder {
         self
     }
 
+    /// Enable or disable the server to send push promises.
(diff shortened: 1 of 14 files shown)


carllerche:
Should PUSH_PROMISE be enabled or disabled by default?

seanmonstar:
The spec claims the enable push setting is enabled by default.

seanmonstar:
However, if a user doesn't plan to accept push promises, by not polling for them, then perhaps it's better for them to be disabled, so a server knows... But then, if a user did want them, they might assume it is enabled by default, because the spec says it is...

PR #96, opened 2017-09-18
Add `fixtures`, `examples`, and `util` to Codecov ignore list

hawkw:
We probably don't care about covering code in these directories, so exclude them from test coverage.


carllerche:
Could you add `src/hpack/test`?

PR #107, opened 2017-09-20
split Client into Connection and Client

seanmonstar:
The Connection type is a `Future` that drives all of the IO of the
client connection.

The Client type is separate, and is used to send requests into the
connection.

An unresolved question is what order of the tuple is better:

- `let (client, conn) = await!(handshake)?;`
- `let (conn, client) = await!(handshake)?;`

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,7 +1,7 @@
 use codec::{Codec, RecvError};
 use frame::{Headers, Pseudo, Settings, StreamId};
 use frame::Reason::*;
-use proto::{self, Connection, WindowSize};
+use proto::{self, WindowSize};
 
 use bytes::{Bytes, IntoBuf};
 use futures::{Async, Future, MapErr, Poll};
@@ -21,8 +21,13 @@ pub struct Handshake<T: AsyncRead + AsyncWrite, B: IntoBuf = Bytes> {
 }
 
 /// Marker type indicating a client peer
-pub struct Client<T, B: IntoBuf> {
-    connection: Connection<T, Peer, B>,
+pub struct Client<B: IntoBuf> {
+    inner: proto::Streams<B::Buf, Peer>,
+    pending: Option<proto::StreamKey>,
+}
+
+pub struct Connection<T, B: IntoBuf> {
+    inner: proto::Connection<T, Peer, B>,
 }
 
 #[derive(Debug)]
@@ -45,10 +50,9 @@ pub struct Builder {
 #[derive(Debug)]
 pub(crate) struct Peer;
 
-impl<T> Client<T, Bytes>
-where
-    T: AsyncRead + AsyncWrite,
-{
+// ===== impl Client =====
+
+impl Client<Bytes> {
     /// Bind an H2 client connection.
     ///
     /// Returns a future which resolves to the connection value once the H2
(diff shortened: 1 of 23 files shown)


src/client.rs
```diff
@@ -216,13 +206,40 @@ impl Default for Builder {
...
+}
+
 // ===== impl Handshake =====
 
 impl<T, B: IntoBuf> Future for Handshake<T, B>
 where
     T: AsyncRead + AsyncWrite,
 {
-    type Item = Client<T, B>;
+    type Item = (Connection<T, B>, Client<B>);
```
carllerche: So, I don't really care about the order between `(Connection, Client)`, but I would like to try to be consistent.

I've been using `(Self, Other)` when adding a fn in a `impl Self`. 

That said, it seems more idiomatic in Rust to have fns that return tuples to be free fns (like `mpsc`). Though, I felt that was strange when the module was the same name as one of the types being returned...

seanmonstar:
> I believe that, as is, the PR would break Client::poll_ready behavior. Each Client handle is going to need its own Task to get notified when readiness transitions -> ready.

I haven't yet traced the `send_request` and `poll_ready` methods to the end, but I can see how this could be true. As a starting point, I added a test that send the `Client` to another thread, while waiting on the `Connection`. It seems to just work, but maybe that's a consequence of using the `mock` IO?

Also, re-ordered to `let (client, conn) = ...`.

src/proto/streams/streams.rs
```diff
@@ -442,6 +483,10 @@ where
         let me = &mut *me;
 
         let stream = me.store.resolve(self.key);
+        if stream.is_pending_open {
+            //XXX: no?
```
seanmonstar: This differs from the behavior when the user has an opened `Stream`, but tries to send more data than there is capacity in the window. We could stick with this (with a different error variant), or allow buffering data even on a pending stream, to be more consistent for the user.

src/proto/streams/streams.rs
```diff
@@ -625,7 +680,15 @@ where
     P: Peer,
 {
     fn drop(&mut self) {
-        let mut me = self.inner.lock().unwrap();
+        let mut me = match self.inner.lock() {
+            Ok(inner) => inner,
+            Err(_) => if ::std::thread::panicking() {
+                trace!("StreamRef::drop; mutex poisoned");
```
seanmonstar: I noticed quite a few times when hitting assertions that this drop function was causing aborts. This code could likely stay, because aborting is 💩 .

tests/support/mod.rs
```diff
@@ -1,5 +1,11 @@
 //! Utilities to support tests.
 
+#[cfg(not(feature = "unstable"))]
+compile_error!(
```
seanmonstar: If you forget to enable the `unstable` feature when running the tests, the error message is quite bizarre (about duplicate functions on `frames::Mock` and so forth). We can make it super clear with this.

src/client.rs
```diff
@@ -91,7 +100,7 @@ where
     /// Returns `Ready` when the connection can initialize a new HTTP 2.0
     /// stream.
     pub fn poll_ready(&mut self) -> Poll<(), ::Error> {
-        self.connection.poll_send_request_ready()
+        self.inner.poll_pending_open(self.pending.as_ref())
```
seanmonstar: We may want to add a `fn ready(self) -> impl Future` to ease dealing with this.

I also think it may be nice to add a similar future to `Stream` for when capacity is available (maybe combines with when the pending stream is opened, and so a user could send data, depends on dicussion lower in this PR).

src/proto/streams/prioritize.rs
```diff
@@ -476,122 +484,142 @@ where
...
+                while let Some(mut stream) = self.pending_open.pop(store) {
+                    trace!("pop_frame; pending_open {:?}", stream.id);
+                    if let Some(frame) = stream.open_headers.take() {
+                        let end_stream = frame.is_end_stream();
+                        let mut task = stream.open_task.take();
+                        // Update the state
+                        if let Err(_e) = stream.state.send_open(end_stream) {
+                            // other peer may have closed this stream between us pending
+                            // and finally opening
+                            error!("pending open errored");
```
seanmonstar: This may not be great. I originally had `send_open(..).expect("this shouldn't happen")`...

seanmonstar:
Updated with in-person comments.

- Opening a pending stream is almost just like opening a new stream. If there is no room for another concurrent stream, it is added to a `pending_open` queue.
- Moved popping of the queue out of `pop_frame`, and into the loop of `poll_complete`.

carllerche:
Oh rustfmt

PR #108, opened 2017-09-20
Rewrite flow control tests to use newer mock API

hawkw:
I've rewritten the tests `flow_control::stream_close_by_data_frame_releases_capacity()` and `flow_control::stream_close_by_trailers_frame_releases_capacity()` to use the new mock API. This will make modifying these tests a bit easier in order to expect the correct behavior in my [reset on drop branch](https://github.com/carllerche/h2/commits/eliza/reset-on-drop).

### DIFF
--- a/tests/flow_control.rs
+++ b/tests/flow_control.rs
@@ -400,11 +400,11 @@ fn stream_close_by_data_frame_releases_capacity() {
 #[test]
 fn stream_close_by_trailers_frame_releases_capacity() {
     let _ = ::env_logger::init();
-    let (m, mock) = mock::new();
+    let (io, srv) = mock::new();
 
     let window_size = frame::DEFAULT_INITIAL_WINDOW_SIZE as usize;
 
-    let h2 = Client::handshake(m).unwrap().and_then(|mut h2| {
+    let h2 = Client::handshake(io).unwrap().and_then(|mut h2| {
         let request = Request::builder()
             .method(Method::POST)
             .uri("https://http2.akamai.com/")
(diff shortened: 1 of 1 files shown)


tests/flow_control.rs
```diff
@@ -380,55 +380,31 @@ fn stream_close_by_data_frame_releases_capacity() {
...
-            assert_eq!(data.stream_id(), 3);
-            assert_eq!(data.payload(), "hello");
-            assert!(data.is_end_stream());
-
-            Ok(())
-        });
-
-    let _ = h2.join(mock).wait().unwrap();
+    let srv = srv.assert_client_handshake().unwrap()
+        .ignore_settings()
```
seanmonstar: Prefer `.recv_settings()` unless it can't work.

tests/flow_control.rs
```diff
@@ -380,55 +380,31 @@ fn stream_close_by_data_frame_releases_capacity() {
...
+        .recv_frame(
+            frames::headers(1).request("POST", "https://http2.akamai.com/")
+        )
+        .send_frame(frames::headers(1).response(200))
```
seanmonstar: The code being replaced didn't send frames back, is this difference important?

tests/flow_control.rs
```diff
@@ -380,55 +380,31 @@ fn stream_close_by_data_frame_releases_capacity() {
...
+        .recv_frame(
+            frames::headers(3).request("POST", "https://http2.akamai.com/")
+        )
+        .send_frame(frames::headers(3).response(200))
+        .recv_frame(frames::data(1, &b""[..]).eos())
+        .recv_frame(frames::data(3, &b"hello"[..]).eos())
+        .close()
+        ;
```
seanmonstar: rustfmt won't like this :D

tests/flow_control.rs
```diff
@@ -475,44 +451,25 @@ fn stream_close_by_trailers_frame_releases_capacity() {
...
+        .send_frame(frames::headers(1).response(200))
+        .recv_frame(
+            frames::headers(3)
+                .request("POST", "https://http2.akamai.com/")
+        )
+        .send_frame(frames::headers(3).response(200))
+        .recv_frame(frames::headers(1).eos())
+        .recv(frames::data(3, b"hello"[..]).eos())
+        .close()
+        ;
```
seanmonstar: Same comments as test above.

tests/flow_control.rs
hawkw: It's not currently, but it will be – in order to make these tests work with the reset-on-drop PR, the client will need to actually read frames back (to prevent the streams from being dropped).

PR #109, opened 2017-09-22
StreamRef sends RST_STREAM on drop

### LINKED ISSUE #100: Dropping a `Stream` / `Body` should reset it
When all handles related to a stream are dropped, interest has been canceled. If the stream has not yet been closed, this should result in a `RST_STREAM` being sent out.

Originally from #75
Relates to #95

hawkw:
This PR modifies the `Drop` implementation for `StreamRef` to reset the underlying stream if it is the last reference to that stream. Since both `Stream` and `Body` are internally just a `StreamRef`, this means they will both reset the stream on drop; thus, this closes #100.

The assertion that the store no longer contains the dropped stream ID at the end of the `Drop` method  had to be removed, as the stream has to be reset from inside of a `transition` block (which now manages releasing that ID for us), and the `transition` closure moves the value of `stream`, making the assertion no longer possible.

Modifications to some of the tests in `flow_control.rs` were also necessary, in order to prevent `StreamRef`s from being dropped too early.

### DIFF
--- a/src/proto/streams/send.rs
+++ b/src/proto/streams/send.rs
@@ -134,6 +134,9 @@ where
         // Transition the state
         stream.state.set_reset(reason);
 
+        // TODO: this could be a call to `recv_err`, but that will always
+        //       clear the send queue. could we pass whether or not to clear
+        //       the send queue to that method?
         if clear_queue {
             // Clear all pending outbound frames
             self.prioritize.clear_queue(stream);
@@ -149,9 +152,7 @@ where
         trace!("send_reset -- queueing; frame={:?}", frame);
         self.prioritize.queue_frame(frame.into(), stream, task);
 
-        // Re-assign all capacity to the connection
-        self.prioritize
-            .assign_connection_capacity(available, stream);
+
     }
 
     pub fn send_data(
@@ -258,6 +259,19 @@ where
         }
 
         Ok(())
+    }
+
+    pub fn recv_err(&mut self, stream: &mut store::Ptr<B, P>) {
+        // Clear all pending outbound frames
+        self.prioritize.clear_queue(stream);
+
+        // Reclaim all capacity assigned to the stream and re-assign it to the
+        // connection
+        let available = stream.send_flow.available();
+        stream.send_flow.claim_capacity(available);
+        // Re-assign all capacity to the connection
+        self.prioritize
+            .assign_connection_capacity(available, stream);
     }
 
     pub fn apply_remote_settings(
--- a/src/proto/streams/send.rs
+++ b/src/proto/streams/send.rs
@@ -98,22 +98,34 @@ where
         let is_closed = stream.state.is_closed();
         let is_empty = stream.pending_send.is_empty();
         trace!(
-            "send_reset; id={:?}; is_reset={:?}; is_closed={:?}; \
-             pending_send.is_empty={:?}\
+            "send_reset(..., stream={:?}, ..., clear_queue={:?}); \
+             is_reset={:?}; is_closed={:?}; pending_send.is_empty={:?}; \
+             state={:?} \
             ",
             stream.id,
+            clear_queue,
             is_reset,
             is_closed,
             is_empty,
+            stream.state
         );
         if stream.state.is_reset() {
             // Don't double reset
+            trace!(
+                " -> not sending RST_STREAM ({:?} is already reset)",
+                stream.id
+            );
             return;
         }
 
         // If closed AND the send queue is flushed, then the stream cannot be
-        // reset either
-        if is_closed && is_empty {
+        // reset explicitly, either. Implicit resets can still be queued.
+        if is_closed && (is_empty || !clear_queue) {
+            trace!(
+                " -> not sending explicit RST_STREAM ({:?} was closed \
+                     and send queue was flushed)",
+                stream.id
+            );
             return;
         }
 
--- a/src/proto/streams/send.rs
+++ b/src/proto/streams/send.rs
@@ -26,7 +26,6 @@ where
 
 impl<B, P> Send<B, P>
 where
-    B: Buf,
     P: Peer,
 {
     /// Create a new `Send`
@@ -157,7 +156,10 @@ where
         frame: frame::Data<B>,
         stream: &mut store::Ptr<B, P>,
         task: &mut Option<Task>,
-    ) -> Result<(), UserError> {
+    ) -> Result<(), UserError>
+    where
+        B: Buf,
+    {
         self.prioritize.send_data(frame, stream, task)
     }
 
@@ -191,6 +193,7 @@ where
     ) -> Poll<(), io::Error>
     where
         T: AsyncWrite,
+        B: Buf,
     {
         self.prioritize.poll_complete(store, counts, dst)
     }
--- a/src/proto/streams/send.rs
+++ b/src/proto/streams/send.rs
@@ -81,11 +81,18 @@ where
         Ok(())
     }
 
+    /// Send an RST_STREAM frame
+    ///
+    /// # Arguments
+    /// + `clear_queue`: if true, all pending outbound frames will be cleared,
+    ///    if false, the RST_STREAM frame will be appended to the end of the
+    ///    send queue.
     pub fn send_reset(
         &mut self,
         reason: Reason,
         stream: &mut store::Ptr<B, P>,
         task: &mut Option<Task>,
+        clear_queue: bool
     ) {
         let is_reset = stream.state.is_reset();
         let is_closed = stream.state.is_closed();
@@ -114,7 +121,9 @@ where
         stream.state.set_reset(reason);
 
         // Clear all pending outbound frames
-        self.prioritize.clear_queue(stream);
+        if clear_queue {
+            self.prioritize.clear_queue(stream);
+        }
 
         // Reclaim all capacity assigned to the stream and re-assign it to the
         // connection
@@ -225,7 +234,7 @@ where
     ) -> Result<(), Reason> {
         if let Err(e) = self.prioritize.recv_stream_window_update(sz, stream) {
             debug!("recv_stream_window_update !!; err={:?}", e);
-            self.send_reset(FlowControlError.into(), stream, task);
+            self.send_reset(FlowControlError.into(), stream, task, true);
 
             return Err(e);
         }
--- a/src/proto/streams/send.rs
+++ b/src/proto/streams/send.rs
@@ -87,6 +87,18 @@ where
         stream: &mut store::Ptr<B, P>,
         task: &mut Option<Task>,
     ) {
+        let is_reset = stream.state.is_reset();
+        let is_closed = stream.state.is_closed();
+        let is_empty = stream.pending_send.is_empty();
+        trace!(
+            "send_reset; id={:?}; is_reset={:?}; is_closed={:?}; \
+             pending_send.is_empty={:?}\
+            ",
+            stream.id,
+            is_reset,
+            is_closed,
+            is_empty,
+        );
         if stream.state.is_reset() {
             // Don't double reset
             return;
@@ -94,7 +106,7 @@ where
 
         // If closed AND the send queue is flushed, then the stream cannot be
         // reset either
-        if stream.state.is_closed() && stream.pending_send.is_empty() {
+        if is_closed && is_empty {
             return;
         }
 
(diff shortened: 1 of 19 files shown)


src/proto/streams/streams.rs
```diff
@@ -622,25 +624,32 @@ where
...
-                return;
+        let actions = &mut me.actions;
+        // the reset must be sent inside a `transition` block.
+        // `transition_after` will release the stream if it is
+        // released.
+        me.counts.transition(stream, |_, stream|
+            // if this is the last reference to the stream,
+            // reset the stream.
+            if stream.ref_count == 0 {
+                actions.send.send_reset(
```
seanmonstar: I haven't looked in all the right places yet, but a reset shouldn't be sent if stream is already closed (because of EOS flags on frames sent and received). Does `send_reset` already take that into consideration? Or does it need to be done here?

tests/flow_control.rs
```diff
@@ -530,7 +536,16 @@ fn recv_window_update_on_stream_closed_by_data_frame() {
...
 
             // Wait for the connection to close
-            h2.unwrap()
+            h2.map(|h2| {
+                // keep `stream` from being dropped in order to prevent
+                // it from sending an RST_STREAM frame.
+                std::mem::forget(stream);
+                // i know this is kind of evil, but it's necessary to
+                // ensure that the stream is closed by the EOS frame,
+                // and not by the RST_STREAM.
```
seanmonstar: Would be better to just wait on receiving the frame back from the server instead.

src/proto/streams/streams.rs
carllerche: I believe so: https://github.com/carllerche/h2/blob/master/src/proto/streams/send.rs#L92-L101

tests/flow_control.rs
hawkw: I tried `h2.drive(stream)` but that doesn't seem to be triggered by the window update & never terminates. Am I missing something?

hawkw:
Also, it appears that the `beta` build of this PR passed, but the `nightly` and `stable` builds have both failed with an ICE. Very strange...

tests/flow_control.rs
seanmonstar: Hm, can we receive a response from the server then?

hawkw: @seanmonstar sorry, how would you suggest I do that, exactly? Nothing I've tried here has worked so far...

seanmonstar: Can the mock `srv` just send a headers frame after the window update?

hawkw: Oh, sure! I wasn't sure if it was okay for me to modify the mock server's behavior in these tests...this could've been resolved a long time ago if I'd just asked about that, sorry.

hawkw: UPDATE: that doesn't seem to work either; sending the EOS frame closes the mock, so it panics with
```

Comment (hawkw) [reply tests/flow_control.rs]: Can't make it work in the prioritization test either, it seems to wait forever?

Comment (seanmonstar) [reply tests/flow_control.rs]: Hm, OK, that's fine. I didn't love the `forget`, but getting around it seems to be annoying, and we don't yet have a way to wait on the `stream` to receive window updates, so that can't be used.

**Comment by hawkw** [comment]
Okay, so after rebasing, it looks like this no longer ICEs, but it looks like the CI build is now failing because these changes [don't pass h2spec](https://travis-ci.org/carllerche/h2/jobs/280151587#L1410-L1618).

@carllerche, I'm looping you back in – do you think this change is worth continuing with?

**Comment by carllerche** [comment]
Looks good to me. I would assume the failed h2spec is most likely due to the example server not being correct after this change.

**Comment by hawkw** [comment]
Okay, the good news is: as of 8411b74fcb713368419a744f2b3558ff79bcb4a1, this PR passes `h2spec` again; the bad news is that the current solution is basically a CPU leak. Since the stream-resetting logic had to be moved to a `store.for_each` in `Streams::poll_complete`, long lived connections will get slower and slower as all closed streams have to be tested and skipped...

**Comment by seanmonstar** [comment]
Ideally, `counts.transition_after` would be able to detect that a stream has no refs and should possibly reset, but there isn't a good way to access `actions` from there.

This suggestion feels hacky, but it would at least prevent the big loop on every poll: counts could possibly grow a `streams_that_could_reset`, and then `poll_complete` can look at it and in most cases, not need to loop all streams...

File: src/proto/streams/send.rs [hunk 3]
```diff
@@ -81,28 +81,61 @@ where
         Ok(())
     }
 
+    /// Send an RST_STREAM frame
+    ///
+    /// # Arguments
+    /// + `clear_queue`: if true, all pending outbound frames will be cleared,
+    ///    if false, the RST_STREAM frame will be appended to the end of the
+    ///    send queue.
```
seanmonstar: I think this all works, but if @carllerche wants to glance over, that wouldn't hurt.

tests/flow_control.rs
```diff
@@ -345,49 +345,55 @@ fn stream_close_by_data_frame_releases_capacity() {
         // Send request
         let mut s1 = client.send_request(request, false).unwrap();
 
-        // This effectively reserves the entire connection window
-        s1.reserve_capacity(window_size);
+            // This effectively reserves the entire connection window
+            s1.reserve_capacity(window_size);
```
seanmonstar: 🤔

src/proto/streams/streams.rs
```diff
@@ -687,18 +709,29 @@ where
...
-                return;
-            }
-
-            stream.remove()
-        };
+        let mut stream = me.store.resolve(self.key);
+        // decrement the stream's ref count by 1.
+        stream.ref_dec();
 
-        debug_assert!(!me.store.contains_id(&id));
```
seanmonstar: Why remove this assertion?

src/proto/streams/streams.rs
```diff
@@ -19,9 +19,9 @@ where
 }
 
 /// Reference to the stream state
-#[derive(Debug)]
 pub(crate) struct StreamRef<B, P>
 where
+    B: Buf,
```
seanmonstar: Just curious, why is `B: Buf` needed?

tests/flow_control.rs
hawkw: Oh, looks like this happened while rebasing, my bad - I'll fix that.

src/proto/streams/streams.rs
hawkw: i believe this was originally removed due to ownership reasons - it may no longer be necessary to remove, I can double check.

src/proto/streams/streams.rs
hawkw: @seanmonstar the `Drop` impl requires `B: Buf` because `Send<B, P>::send_reset` is bounded on `B: Buf`. 

```
error[E0599]: no method named `send_reset` found for type `proto::streams::send::Send<B, P>` in the current scope
   --> src/proto/streams/streams.rs:725:30
    |
725 |                 actions.send.send_reset(
    |                              ^^^^^^^^^^
    |
    = note: the method `send_reset` exists but the following trait bounds were not satisfied:
            `B : bytes::Buf`
```
if the `Drop` impl is bounded on `B: Buf`, then the struct definition must also be bounded.

src/proto/streams/streams.rs
hawkw: @seanmonstar the assertion also doesn't make sense here any longer, because whether or not the stream needs to be removed from the store is now handled by `transition`.

src/proto/streams/streams.rs
seanmonstar: Aww, shucks. Very well.

seanmonstar: I think, for extra credit, someone could go in and make the impl on `Send` only need `B: Buf` for `send_data`, and have all the other methods be in an impl without that bounds, but not a big deal.

hawkw: @seanmonstar started looking into that but it looks like there are a few other things bounded on `Buf` so it may not be quite that simple - at the very least, probably another PR.

hawkw: @seanmonstar WDYT of 5079b35fa9634bd879cf2875a8475656b41a08b3?

seanmonstar: Neat!

src/proto/streams/send.rs
```diff
@@ -81,28 +80,61 @@ where
...
+                stream.id
+            );
             return;
         }
 
         // If closed AND the send queue is flushed, then the stream cannot be
-        // reset either
-        if stream.state.is_closed() && stream.pending_send.is_empty() {
+        // reset explicitly, either. Implicit resets can still be queued.
+        if is_closed && (is_empty || !clear_queue) {
```
carllerche: It took me a bit to figure out this logic. Would it be simpler to move the `clear_queue` bit before doing this check? Then you can keep the original check (is_closed && is_empty).

src/proto/streams/streams.rs
```diff
@@ -687,18 +721,29 @@ where
...
+        let mut stream = me.store.resolve(self.key);
+        // decrement the stream's ref count by 1.
+        stream.ref_dec();
 
-        debug_assert!(!me.store.contains_id(&id));
+        let actions = &mut me.actions;
+        // the reset must be sent inside a `transition` block.
+        // `transition_after` will release the stream if it is
+        // released.
+        let recv_closed = stream.state.is_recv_closed();
```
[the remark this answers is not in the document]
seanmonstar: Oh hey, I missed this, I thought it was if the recv half is **NOT** closed...

tests/flow_control.rs
```diff
@@ -527,13 +535,23 @@ fn recv_window_update_on_stream_closed_by_data_frame() {
...
+                // and not by the RST_STREAM.
+                h2
+            }).unwrap()
         });
 
     let srv = srv.assert_client_handshake()
         .unwrap()
         .recv_settings()
-        .recv_frame(frames::headers(1).request("POST", "https://http2.akamai.com/"))
+        .recv_frame(
```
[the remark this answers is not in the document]
hawkw: Yeah, I ran rustfmt on the whole project before committing, I can revert that

src/proto/streams/send.rs
hawkw: @seanmonstar suggested this conditional, but I agree it's not super clear. I can rewrite this.

src/proto/streams/streams.rs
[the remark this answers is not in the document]
hawkw: Fixed in 49627649ef6b2cea50bc81f228a21741a463636d

src/proto/streams/send.rs
carllerche: Not critical either way. I think the logic is correct.

src/proto/streams/streams.rs
```diff
@@ -732,7 +732,7 @@ where
         let recv_closed = stream.state.is_recv_closed();
         me.counts.transition(stream, |_, stream|
             // if this is the last reference to the stream, reset the stream.
-            if stream.ref_count == 0 && recv_closed {
+            if stream.ref_count == 0 && recv_closed == false {
```
carllerche: I think a `!recv_closed` would be more idiomatic.

src/proto/streams/streams.rs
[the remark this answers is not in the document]
carllerche: So, `!recv_closed` makes some more sense, but I"m not sure I follow why we are waiting for the remote to close. If all handles are dropped, then we can't do anything w/ frames received from the remote. This is why we want to send a `RST_STREAM`, so that the remote can stop doing work?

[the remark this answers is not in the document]
seanmonstar: Yea, the point is to reset the stream if the user is no longer going to do anything with it. But, the the remote is already closed, there is no need to reset. At that point, it's just an unnecessary frame, that hopefully the remote will discard correctly, but it seems trivial for us to know right here whether the frame is needed...

src/proto/streams/streams.rs
hawkw: Hmm, I generally prefer `== false` in complex conditionals – I feel like it's easy to miss a single exclamation point but much harder to miss the equality check. But I'm happy to change it.

src/proto/streams/streams.rs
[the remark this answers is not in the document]
carllerche: Ok, so what happens when the sender is still sending the body, but never sends an EOS and all the handles drop? At that point, a RST_STREAM should be sent out due to the sender aborting... I'm not sure where that happens.

PR #111, opened 2017-09-24
Update so dependents can compile

olix0r:
Projects cannot currently depend on the h2 repo due the following compile errors:

```
error: method is never used: `new`
   --> /Users/ver/b/rs/h2/src/frame/headers.rs:265:5
    |
265 | /     pub fn new(
266 | |         stream_id: StreamId,
267 | |         promised_id: StreamId,
268 | |         pseudo: Pseudo,
...   |
279 | |         }
280 | |     }
    | |_____^
    |
note: lint level defined here
   --> /Users/ver/b/rs/h2/src/lib.rs:1:9
    |
1   | #![deny(warnings, missing_debug_implementations)]
    |         ^^^^^^^^
    = note: #[deny(dead_code)] implied by #[deny(warnings)]

error: method is never used: `into_parts`
   --> /Users/ver/b/rs/h2/src/frame/headers.rs:339:5
    |
339 | /     pub fn into_parts(self) -> (Pseudo, HeaderMap) {
340 | |         (self.header_block.pseudo, self.header_block.fields)
341 | |     }
    | |_____^

error: method is never used: `fields`
   --> /Users/ver/b/rs/h2/src/frame/headers.rs:343:5
    |
343 | /     pub fn fields(&self) -> &HeaderMap {
344 | |         &self.header_block.fields
345 | |     }
    | |_____^

error: method is never used: `into_fields`
   --> /Users/ver/b/rs/h2/src/frame/headers.rs:347:5
    |
347 | /     pub fn into_fields(self) -> HeaderMap {
348 | |         self.header_block.fields
349 | |     }
    | |_____^

error: method is never used: `set_max_concurrent_streams`
  --> /Users/ver/b/rs/h2/src/frame/settings.rs:77:5
   |
77 | /     pub fn set_max_concurrent_streams(&mut self, max: Option<u32>) {
78 | |         self.max_concurrent_streams = max;
79 | |     }
   | |_____^

error: aborting due to 5 previous errors

To learn more, run the command again with --verbose.
```

This change adds two `allow(dead_code)` annotations to prevent these compile errors.

Furthermore, `cargo fmt` has been run.

### DIFF
--- a/src/frame/headers.rs
+++ b/src/frame/headers.rs
@@ -261,7 +261,7 @@ impl fmt::Debug for Headers {
 
 // ===== impl PushPromise =====
 
-#[allow(dead_code)]
+#[cfg(feature = "unstable")]
 impl PushPromise {
     pub fn new(
         stream_id: StreamId,
--- a/src/frame/headers.rs
+++ b/src/frame/headers.rs
@@ -261,6 +261,7 @@ impl fmt::Debug for Headers {
 
 // ===== impl PushPromise =====
 
+#[allow(dead_code)]
 impl PushPromise {
     pub fn new(
         stream_id: StreamId,
--- a/src/frame/headers.rs
+++ b/src/frame/headers.rs
@@ -6,7 +6,7 @@ use http::{uri, HeaderMap, Method, StatusCode, Uri};
 use http::header::{self, HeaderName, HeaderValue};
 
 use byteorder::{BigEndian, ByteOrder};
-use bytes::{BufMut, Bytes, BytesMut};
+use bytes::{Bytes, BytesMut};
 use string::String;
 
 use std::fmt;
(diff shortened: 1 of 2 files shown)


carllerche:
It looks like this is also running a `cargo fmt`. Same request as in #112 could we make a standalone format PR and then rebase these PRs on that?

src/frame/headers.rs
```diff
@@ -261,6 +261,7 @@ impl fmt::Debug for Headers {
 
 // ===== impl PushPromise =====
 
+#[allow(dead_code)]
```
carllerche: This should actually be flagged w/ unstable I think. I'll update the PR.

seanmonstar:
Any errors from lints in dependencies should never fail compilation, cargo caps lints to warnings. If this isn't the case, that sounds like a bug in cargo or something.

carllerche:
I updated the PR, so I probably shouldn't approve it?

carllerche:
@seanmonstar I'm guessing it is failing because it is a git dep.

carllerche:
Ugh, I think this isn't quite right.

carllerche:
Ok, I think this should fix it.

PR #117, opened 2017-09-25
Only run tests once in CI

olix0r:
We run codecov after successful test runs. This ~doubles the length of
CI jobs, sicne the codecov run reproduces nearly all tests.

To avoid this, we just run the codecov tests as part of our normal test
script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,12 +21,15 @@ install:
   - bash <(curl https://raw.githubusercontent.com/xd009642/tarpaulin/master/travis-install.sh)
 
 script:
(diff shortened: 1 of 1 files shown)


hawkw:
@olix0r after this lands, I can see about passing the flag to tarpaulin to skip counting the number of times a line was covered, which should make coverage much faster (possibly even fast enough to run hpack under coverage?).

carllerche:
Looks like the answer is no?

hawkw:
@carllerche it looks like the failure is unrelated to this branch?

PR #118, opened 2017-09-25
Make coverage faster

hawkw:
This PR changes the coverage step on CI to pass the `--skip clean` flag (which would prevent `tarpaulin` from recompiling) and the `--no-count` flag, which disables counting repeated hits of the same line under coverage. This should make the coverage step much faster.

Depending on the performance improvement from this PR, we might be able to start running hpack under coverage again.


hawkw:
Okay, so this actually ran _with_ hpack running under coverage, but it looks like it [managed it in 12 minutes or less](https://travis-ci.org/carllerche/h2/builds/279622156) – I suspect if we were skipping hpack properly, this would be _much_ faster.

hawkw:
Okay, I've rebased this against the current `master` & it appears to work correctly now - the PR is marked as failing because it results in a very slight decrease in coverage (probably we should increase the threshold so that -0.07% drops in coverage don't fail a PR?).

hawkw:
It looks like this shaves about 100 seconds off the coverage step vs master. I wonder what it would look like if we were running hpack under coverage?

hawkw:
So, as of ba52239, it looks like we can now run the HPACK tests under coverage in _roughly_ the same amount of time it takes to run tests with HPACK excluded from coverage on `master`?

PR #123, opened 2017-09-26
Make CI build flatter (and therefore faster)

hawkw:
Removed build stages from travis configuration. `rustfmt` and doc publishing run are run on the nightly build while `h2spec` runs on the stable build (as suggested by @carllerche). This also adds the changes to make coverage faster added in #118.

Closes #118.


hawkw:
This shaves off about 5 minutes from the total runtime, so it's better but not a massive improvement.

carllerche:
Seems good to me... but test coverage was reduced by 0.03%!!!! 🔨

PR #124, opened 2017-09-27
Set codecov threshold to 5%

hawkw:
This PR sets the threshold by which code coverage can drop before Codedov marks the commit as failing or notifies Slack to 5%, as requested by @seanmonstar.


carllerche:
Looks good to me! I think it needs to be rebased though.

Issue #83, opened 2017-09-12
Current handling of GOAWAY is pretty broken


carllerche:
Right now, `GOAWAY` is interpreted as a signal to reset all streams. This, however, is not correct behavior.

seanmonstar:
Handling GOAWAY is a complicated mess. A GOAWAY frame does **not** necessarily mean the connection is terminating (now or even that soon). It contains a reason, and a last stream ID. The connection should continue to try to process frames for any stream below the listed stream ID, and not make any new streams.

It seems a graceful shutdown involves one side sending a GOAWAY, both sides finishing up the existing streams until EOS, and then the other side sends GOAWAY, and the connection terminates.

- When **receiving** a GOAWAY:
  - Close the ability to create new streams
  - Drop any streams higher than the reported last stream ID. (There is no need to send a RST_STREAM, it's considered dead.)
  - Keep processing all streams lower than or equal to the last stream ID.
  - Once all streams are dropped or EOS, send GOAWAY and close.
  - `Connection::poll()` should then resolve:
    - If there was `NO_ERROR`, then `Ok(Async::Ready(()))`
    - Otherwise, `Err(goaway_frame.reason())`.
    - If there is an IO or protocol error *after* receiving a GOAWAY, what should be reported to the user?
      - It's possible to send multiple GOAWAY frames as well, in case graceful shutdown encountered an error.
- When we encounter a protocol error, and need to **send** a GOAWAY:
  - We need to determine if that error is recoverable.
    - A frame parse error, or a `FRAME_SIZE_ERROR`, or a `COMPRESSION_ERROR`, etc may mean that there is no way to continue using the connection at all. The best we can do is send a GOAWAY and then close the connection, not continuing to process frames.
    - Other errors might be recoverable, such as receiving an invalid frame type on a stream not accepting those frames. The spec says to send a GOAWAY, but we can still try to finish up any existing valid streams.
      - In this case, we can act just like when we receive a GOAWAY.

carllerche:
When *receiving* a GOAWAY.

I would add that all streams higher than the reported stream ID should behave the same *as if* they received a RST_STREAM. Specifically, the GOAWAY reason is bubbled up to the stream handles.

> If there is an IO or protocol error after receiving a GOAWAY, what should be reported to the user? 

I'd say the same thing that happens if there wasn't a GOAWAY frame before hand. (probably the same thing that happens now?)

When *sending*

[...]

PR #133, opened 2017-10-04
when receiving a GOAWAY, allow earlier streams to still process

seanmonstar:
This is a first step to better GOAWAY support (cc #83). This patch will close all streams greater than the last stream ID in the GOAWAY frame, but allow streams less than to still finish.

### DIFF
--- a/src/frame/go_away.rs
+++ b/src/frame/go_away.rs
@@ -16,7 +16,6 @@ impl GoAway {
         }
     }
 
-    #[cfg(feature = "unstable")]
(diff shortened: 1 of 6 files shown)


seanmonstar:
Looks like this broke h2spec, expecting the connection to have closed, so I'll need to add the second step into this PR as well: detecting when the streams are finished and return the goaway + close.

carllerche:
LGTM so far (minus h2spec failure)

seanmonstar:
Updated to now look for when all active streams have finished, and responds with a new GOAWAY and shuts down.

PR #134, opened 2017-10-05
remove unused pieces from PingPong

seanmonstar:
Adds some extra tests as well, to be sure.

We probably want to expose an API for a user to be able to send pings, to allow them to record latency and otherwise know that the remote is still there, but until we do, these fields are just bloat.


carllerche:
This will have to be handled better one day, but I can create individual issues to track.

PR #135, opened 2017-10-06
Remove most instances of unimplemented!

seanmonstar:
Trying to increase robustness and reduce panics, I've swept through the `unimplemented!`s, and converted most of them into something more appropriate. The ones I left required more thought, and are sufficiently rare that I didn't want to block these.

### DIFF
--- a/src/proto/streams/recv.rs
+++ b/src/proto/streams/recv.rs
@@ -263,9 +263,9 @@ where
     ) -> Result<(), RecvError> {
         let sz = frame.payload().len();
 
-        if sz > MAX_WINDOW_SIZE as usize {
-            unimplemented!();
-        }
+        // This should have been enforced at the codec::FramedRead layer, so
+        // this is just a sanity check.
+        assert!(sz <= MAX_WINDOW_SIZE as usize);
 
         let sz = sz as WindowSize;
 
@@ -611,7 +611,7 @@ where
                 // TODO: This is a user error. `poll_trailers` was called before
                 // the entire set of data frames have been consumed. What should
                 // we do?
-                unimplemented!();
+                panic!("poll_trailers called before data has been consumed");
             },
             None => self.schedule_recv(stream),
         }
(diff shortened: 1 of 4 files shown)


src/proto/streams/recv.rs
```diff
@@ -610,7 +611,7 @@ where
                 // TODO: This is a user error. `poll_trailers` was called before
                 // the entire set of data frames have been consumed. What should
                 // we do?
-                unimplemented!();
+                panic!("poll_trailers called before data has been consumed");
```
carllerche: Is a panic correct here? I'm a bit worried that it will be a common user error. Also, maybe we can actually support polling trailers before polling all body data?

Anyway, I would recommend leaving this as `unimplemented` for now and tracking this case individually w/ an issue.

seanmonstar: Yea, I wondered a little about this. Things we could do:

- Return a `UserError`. You done goofed. Downside is that returning an `Err` usually means the future is dead, and can't be used anymore. Would that mean we just have to trash the state of this stream altogether? Unhappy unfun times.
- We *could* recover from this, by pushing the trailers back to the front of the queue, and return `Async::NotReady`. This might have subtly wrong cases, though. Returning `NotReady` means we should have a way to notify the task when it *is* ready... and when is that? When the user has later polled off all the remaining data frames? If they've done that, we still won't have a `Task` to notify...

carllerche: Well... I wonder if you can just `pop_back` instead of `pop_front`. (maybe w/ a peek first). Since the trailer will be at the end of the buffer and it is a double ended queue... then one could actually support this behavior.

carllerche: Either way, I don't think fixing this properly should block the commit. Just track it and merge this.

seanmonstar: I opened #139 to track this.

PR #136, opened 2017-10-06
reset streams when receiving malformed pseudo headers

### LINKED ISSUE #120: panic on invalid :authority
When the value of `:authority` is invalid, we get the following panic:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidUriBytes(InvalidUri(InvalidUriChar))', src/libcore/result.rs:906:4

   9: core::result::unwrap_failed
  10: <h2::server::Peer as h2::proto::peer::Peer>::convert_poll_message
  11: <h2::proto::connection::Connection<T, P, B>>::poll2
  12: <h2::server::Server<T, B> as futures::stream::Stream>::poll
```

This should probably be a User error and not a panic.

seanmonstar:
Closes #120 

(also includes some rustfmt changes I've forgotten in the past few PRs...)

### DIFF
--- a/src/frame/headers.rs
+++ b/src/frame/headers.rs
@@ -211,6 +211,11 @@ impl Headers {
         (self.header_block.pseudo, self.header_block.fields)
     }
 
+    #[cfg(feature = "unstable")]
+    pub fn pseudo_mut(&mut self) -> &mut Pseudo {
+        &mut self.header_block.pseudo
+    }
+
     pub fn fields(&self) -> &HeaderMap {
         &self.header_block.fields
     }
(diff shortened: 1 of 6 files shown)


carllerche:
Is it possible to omit the rustfmt changes? It makes it really hard to review the change as the formatting adds a lot of noise.

seanmonstar:
Yea, I knew I should have put them in a separate commit. I forgot to make sure I'd made a commit before running `cargo fmt`, and then it was too late. I'll see if I can separate them, but while doing so, I can point at the relevant pieces.

- The removals of `unwrap` in `src/server.rs`
- The test added in `tests/server.rs`

seanmonstar:
Removed unrelated formatting.

carllerche:
Thanks. I think we just have to give up on rustfmt until it becomes stable.

PR #144, opened 2017-10-08
Fail CI builds if tests could not be compiled (workaround for #143)

### LINKED ISSUE #143: Errors compiling tests don't break the build on CI
`tarpaulin` apparently exits with 0 if there are compiler errors in `cargo test`. Since we are only running tests on CI using tarpaulin, and not compiling them first, this means that if the crate compiles correctly, but the tests don't compile, the build will appear to have succeeded when, in fact, tests were never even run.

See [build 605](https://travis-ci.org/carllerche/h2/jobs/285119212#L531-L551) for an example of this problem.

hawkw:
Tarpaulin – the coverage tool we're using on CI – exits with 0 if a compile error occurred when compiling tests. Since we are currently only building and running tests with Tarpaulin, this means that builds where the tests couldn't be compiled successfully are erroneously being marked as "successful" – see #143 for details.

Closes #143.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,11 @@ script:
   # Check with unstable flag
   - cargo check --features unstable
 
-  # Run tests, uploading results to codecov..
+  # Build the test executables in a separate step, since Tarpaulin
+  # will not fail on test compile errors.
+  - cargo build --tests --features unstable
+
+  # Run tests, uploading results to codecov.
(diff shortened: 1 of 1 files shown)


hawkw:
Note that merging this will break the build, since merging 1c179f7bf2b5722320e8df01b0c0eb37db3d8d07 _should_ have broken the build but didn't.

PR #145, opened 2017-10-09
Use remote window size for send flow control

olix0r:
The initial window size of both the send and recv flow controllers are
configured from the same setting. The send flow controller should be
configured from the remote's initial window size.

I'm unsure of how to test this.

### DIFF
--- a/src/proto/streams/prioritize.rs
+++ b/src/proto/streams/prioritize.rs
@@ -51,11 +51,11 @@ where
     pub fn new(config: &Config) -> Prioritize<B, P> {
         let mut flow = FlowControl::new();
 
-        flow.inc_window(config.local_init_window_sz)
+        flow.inc_window(config.remote_init_window_sz)
             .ok()
             .expect("invalid initial window size");
 
-        flow.assign_capacity(config.local_init_window_sz);
+        flow.assign_capacity(config.remote_init_window_sz);
 
         trace!("Prioritize::new; flow={:?}", flow);
 


olix0r:
This isn't a complete fix -- the connection window is not actually set to 654321...
```
1507559240.159027 TRACE h2::proto::streams::recv recv_data; size=9; connection=65535; stream=654321
```

I see this comment in recv.rs, which I don't quite understand:
```

        // connections always have the default window size, regardless of
        // settings
        flow.inc_window(DEFAULT_INITIAL_WINDOW_SIZE)
            .expect("invalid initial remote window size");
        flow.assign_capacity(DEFAULT_INITIAL_WINDOW_SIZE);
```

I suspect that this means that the receive window setting should only be applied once the local peer's setting has been acknowledged by the remote?

seanmonstar:
What you're seeing is according to the spec:

> A SETTINGS frame cannot alter the connection flow-control window.

The connection window can only be changed with WINDOW_UPDATE frames.

Cargo.toml
```diff
@@ -55,3 +55,6 @@ io-dump = { git = "https://github.com/carllerche/io-dump" }
 rustls = "0.11"
 tokio-rustls = { git = "https://github.com/briansmith/tokio-rustls", tag = "b/p1" }
 webpki-roots = "0.13"
+
+[replace]
+"http:0.1.0" = { git = "https://github.com/hyperium/http" }
```
seanmonstar: This shouldn't be needed any more, we just published 0.1.1 that includes the changes the tests were relying upon.

hawkw: 🙌

seanmonstar:
> ./ci/h2spec.sh exited with 0.
> Done. Your build exited with 1.

 (╯°□°）╯︵ ┻━┻

carllerche:
Clearly we should be new typing receive vs. send flow values differently...

seanmonstar:
> Clearly we should be new typing receive vs. send flow values differently...

Not sure if joking, but I kinda think it's a good idea. `WindowSize<Local>` and `WindowSize<Remote>` (not exposed externally, just to save ourselves).

carllerche:
It's Ok, nobody ever knows when I'm joking.

PR #147, opened 2017-10-09
fix ICE when compiling prioritization tests

seanmonstar:


carllerche:
Ok, I'm going to force merge this... because ICE.

PR #149, opened 2017-10-09
add set_target_window_size methods to Server and Client

### LINKED ISSUE #101: expose APIs to set target window size
There is no API to change the target window size to be different from SETTINGS_INITIAL_WINDOW_SIZE.  An endpoint may wish to this on either a connection or individual stream.

Clients, servers, and their streams should probably have `window_target(&self) -> u32` and `set_window_target(&mut self, sz: u32)` functions.

seanmonstar:
We already track how much of the window is "reserved" by body streams, with connection flows having the default initial size of 65,535. As streams reserve some and release some, the connection window gets reset back up to that number (because we always subtract and then add the same amount to it).

This PR adds a new method to allow stating that the target window size is no longer the default, but something else. If the difference from the default is large enough, a WINDOW_UPDATE is sent out (because of our queuing of updates). Besides that, as streams use the connection window, and then release it back, we'll continue to send updates to the remote to get the window back to the target size (again, based on queuing window updates till a threshold is reached).

Closes #101

### DIFF
--- a/src/proto/streams/flow_control.rs
+++ b/src/proto/streams/flow_control.rs
@@ -1,6 +1,8 @@
 use frame::Reason;
 use proto::{WindowSize, MAX_WINDOW_SIZE};
 
+use std::fmt;
+
 // We don't want to send WINDOW_UPDATE frames for tiny changes, but instead
 // aggregate them when the changes are significant. Many implementations do
 // this by keeping a "ratio" of the update version the allowed window size.
@@ -25,32 +27,41 @@ fn sanity_unclaimed_ratio() {
 
 #[derive(Copy, Clone, Debug)]
 pub struct FlowControl {
-    /// Window size as indicated by the peer. This can go negative.
-    window_size: i32,
+    /// Window the peer knows about.
+    ///
+    /// This can go negative if a SETTINGS_INITIAL_WINDOW_SIZE is received.
+    ///
+    /// For example, say the peer sends a request and uses 32kb of the window.
+    /// We send a SETTINGS_INITIAL_WINDOW_SIZE of 16kb. The peer has to adjust
+    /// its understanding of the capacity of the window, and that would be:
+    ///
+    /// ```notrust
+    /// default (64kb) - used (32kb) - settings_diff (64kb - 16kb): -16kb
+    /// ```
+    window_size: Window,
 
-    /// The amount of the window that is currently available to consume.
-    available: WindowSize,
+    /// Window that we know about.
+    ///
+    /// This can go negative if a user declares a smaller target window than
+    /// the peer knows about.
+    available: Window,
 }
 
 impl FlowControl {
     pub fn new() -> FlowControl {
         FlowControl {
-            window_size: 0,
-            available: 0,
+            window_size: Window(0),
+            available: Window(0),
         }
     }
 
     /// Returns the window size as known by the peer
     pub fn window_size(&self) -> WindowSize {
-        if self.window_size < 0 {
-            0
-        } else {
-            self.window_size as WindowSize
-        }
+        self.window_size.as_size()
     }
 
     /// Returns the window size available to the consumer
-    pub fn available(&self) -> WindowSize {
+    pub fn available(&self) -> Window {
         self.available
     }
 
@@ -60,11 +71,10 @@ impl FlowControl {
             return false;
         }
 
-        self.window_size as WindowSize > self.available
+        self.window_size > self.available
     }
 
     pub fn claim_capacity(&mut self, capacity: WindowSize) {
-        assert!(self.available >= capacity);
         self.available -= capacity;
     }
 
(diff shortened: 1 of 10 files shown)


src/proto/streams/recv.rs
```diff
@@ -246,6 +250,19 @@ where
         Ok(())
     }
 
+    pub fn set_target_connection_window(&mut self, target: WindowSize, task: &mut Option<Task>) {
+        let current = self.flow.available() + self.in_flight_data;
+        if target > current {
```
seanmonstar: Now that I file this, it occurred to me that this won't really allow turning the target window down, only up. I'll go fix that.

src/proto/streams/recv.rs
```diff
@@ -20,6 +20,8 @@ where
     /// Connection level flow control governing received data
     flow: FlowControl,
 
+    in_flight_data: WindowSize,
```
carllerche: Would you mind leaving a comment describing what this field is used for?

src/proto/streams/recv.rs
```diff
@@ -246,6 +250,19 @@ where
         Ok(())
     }
 
+    pub fn set_target_connection_window(&mut self, target: WindowSize, task: &mut Option<Task>) {
+        let current = self.flow.available() + self.in_flight_data;
+        if target > current {
+            self.flow.assign_capacity(target - current);
+
+            if self.flow.unclaimed_capacity().is_some() {
+                if let Some(task) = task.take() {
```
carllerche: For my benefit (as I am losing context of the code already), could we add a comment here indicating that `unclaimed_capacity` being `Some` means that we have passed the window threshold to send a `WINDOW_UPDATE` (we don't always want to send a `WINDOW_UPDATE` for small deltas). And the `task` here is the task that manages the HTTP/2.0 connection, which is responsible for sending out the window updates.

seanmonstar:
This grew a bit...

- Added `proto::streams::flow_control::Window` as a newtype around `i32`.
  
  This tries to centralize integer casts to a single location. The type represents the ability for a window capacity go negative, while keeping `WindowSize` a `u32` as the type that is sent in WINDOW_UPDATE frames.
- Adds ability for setting the target connection window as smaller than the current target. That might result in `available` going negative, if there is already reserved capacity in streams, and so `unclaimed_capacity()` won't return `Some` until the streams have released enough capacity to reach the new smaller target.
- Adds some tests increasing and decreasing target connection size.
- Adds comments to save Future Us.

src/proto/streams/flow_control.rs
```diff
@@ -25,32 +27,40 @@ fn sanity_unclaimed_ratio() {
...
         FlowControl {
-            window_size: 0,
-            available: 0,
+            window_size: Window(0),
+            available: Window(0),
         }
     }
 
     /// Returns the window size as known by the peer
     pub fn window_size(&self) -> WindowSize {
```
carllerche: Is there a reason that this fn doesn't return `Window` now instead of `WindowSize`?

src/proto/streams/flow_control.rs
```diff
@@ -80,14 +89,14 @@ impl FlowControl {
...
-        let available = self.available as i32;
+        let available = self.available;
 
         if self.window_size >= available {
             return None;
         }
 
-        let unclaimed = available - self.window_size;
-        let threshold = self.window_size / UNCLAIMED_DENOMINATOR * UNCLAIMED_NUMERATOR;
+        let unclaimed = available.0 - self.window_size.0;
```
carllerche: If we are adding new types, maybe this could be a fn on the new type. Just an idea, I don't know if it makes sense, it also isn't important so feel free to disregard.

src/proto/streams/flow_control.rs
```diff
@@ -133,8 +142,14 @@ impl FlowControl {
...
         );
         // This should not be able to overflow `window_size` from the bottom.
-        self.window_size -= sz as i32;
-        self.available = self.available.saturating_sub(sz);
+        self.window_size -= sz;
+        self.available -= sz;
+        // This should only be called on Send flow controllers. In this case,
+        // available capacity must have been reserved, and can be freed. We
+        // don't need to go negative here.
+        if self.available.0 < 0 {
```
carllerche: I'm not sure that I follow this scenario & comment. If this is only called on the `Send` flow controller (aka, *we* are sending out `DATA` frames to a remote) then how does this happen? It should be impossible to go below zero available? (I'm probably forgetting the exact scenario).

src/proto/streams/flow_control.rs
seanmonstar: Because I wasn't sure of all the places `flow.window_size()` is used, and if it should return a possibly negative number, so I left that part alone...

src/proto/streams/flow_control.rs
seanmonstar: I was confused by this too, but the tests found it. The flow controls on the send side of a stream don't gain available capacity until it's been reserved specifically. In fact, this kind of confuses me. It seems that when a send side reserves capacity, the number gets added to `available`, almost as if this means "bytes we have reserved locally but could give back to the connection".

The situation that broke without this condition is reserving 11 bytes, so then `available == 11`, and then getting a SETTINGS frame reducing initial window size to 0, so it was called as `dec_window(65_535)`, resulting in a negative "available".

seanmonstar: To try to make this less confusing, I've completely removed touching the `available` capacity from `dec_window`. It is now two separate actions. In `Send`, after decreasing the window, the code then takes away `cmp::min(dec, available)` capacity as well.

PR #150, opened 2017-10-10
Add methods to `{client, server}::Builder` to set max concurrent streams

### LINKED ISSUE #106: Support configuration of max_concurrent_streams
`server::Builder` should support `set_max_concurrent_streams(&mut self, u32)` so that a server can constrain the number of streams per client.

Similarly, `client::Builder` should have this, though I _think_ a client can only configure the number of concurrent server-initiated streams (i.e. Push Promises).  While we're here, `client::Builder` should probalby also have a `set_push_promise_enabled(&mut self, bool)`

hawkw:
This PR adds `max_concurrent_streams()` methods to the client and server `Builder`s to set the `max_concurrent_streams` setting. I've added unit tests to ensure the correct SETTINGS frame is sent.

Closes #106

### DIFF
--- a/tests/server.rs
+++ b/tests/server.rs
@@ -23,6 +23,48 @@ fn read_preface_in_multiple_frames() {
 }
 
 #[test]
+fn server_builder_set_max_concurrent_streams() {
+    let _ = ::env_logger::init();
+    let (io, client) = mock::new();
+
+    let mut settings = frame::Settings::default();
+    settings.set_max_concurrent_streams(Some(1));
+
+    let client = client
+        .assert_server_handshake()
+        .unwrap()
+        .recv_custom_settings(settings)
+        .send_frame(
+            frames::headers(1)
+                .request("GET", "https://example.com/")
+                .eos(),
+        )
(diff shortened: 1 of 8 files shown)


tests/server.rs
```diff
@@ -23,6 +23,48 @@ fn read_preface_in_multiple_frames() {
...
+    let client = client
+        .assert_server_handshake()
+        .unwrap()
+        .recv_custom_settings(settings)
+        .send_frame(
+            frames::headers(1)
+                .request("GET", "https://example.com/")
+                .eos(),
+        )
+        .recv_frame(frames::headers(1).response(200).eos())
```
seanmonstar: To test that the setting was applied internally, this could also test that we send back a RST_STREAM of STREAM_REFUSED when the mock exceeds our advertised limit.

To so:

- Send the first request with a data frame also, but not immediately.
- Send a second request.
- Except to receive a `frames::reset(3).refused()`.
- Finish up the first request.
- Test done.

hawkw: Done (3d29b3f)!

PR #152, opened 2017-10-13
Fix some flow control bugs.

carllerche:
* Release stream capacity back to the connection to avoid capacity
leaks.
* Actually notify waiting tasks when capacity becomes available.

### DIFF
--- a/tests/support/util.rs
+++ b/tests/support/util.rs
@@ -35,8 +35,6 @@ impl Future for WaitForCapacity {
 
         let act = self.stream().capacity();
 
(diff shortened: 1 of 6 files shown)


tests/support/util.rs
```diff
@@ -35,8 +35,6 @@ impl Future for WaitForCapacity {
 
         let act = self.stream().capacity();
 
-        println!("CAP={:?}", act);
```
seanmonstar: SO THAT'S WHY I SOMETIMES SEE `CAP=11`!

PR #155, opened 2017-10-18
API cleanup

carllerche:
This PR provides a bunch of API cleanup

### Unifying `client` & `server` types.

The first step is to rid of the internal `P: Peer` generic.

### Removing `B` generic

All public API types that do not work with the send data frame type are no longer generic.

### `&mut self` for `send_reset`

While calling this function is the last thing that should be done with
the instance, the intent of the h2 library is not to be used directly by
users, but to be used as an implementation detail by other libraries.

Requiring `self` on `send_reset` is pretty annoying when calling the
function from inside a `Future` implementation. Also, all the other fns
on the type take `&mut self`.

### DIFF
--- a/src/proto/streams/state.rs
+++ b/src/proto/streams/state.rs
@@ -337,6 +337,7 @@ impl State {
         // TODO: Is this correct?
         match self.inner {
             Closed(Some(Cause::Proto(reason))) => Err(proto::Error::Proto(reason)),
+            Closed(Some(Cause::Canceled)) => Err(proto::Error::Proto(Reason::CANCEL)),
             Closed(Some(Cause::Io)) => Err(proto::Error::Io(io::ErrorKind::BrokenPipe.into())),
             Closed(None) | HalfClosedRemote(..) => Ok(false),
             _ => Ok(true),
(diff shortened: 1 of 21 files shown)


carllerche:
I have pushed a comment that removes the internal `P: Peer` generic.

carllerche:
I pushed splitting `Respond` from `server::Stream`. This reduces `server::Stream` and `client::Stream` to the same API surface.

carllerche:
Ok, I started removing `B` from public API types.

src/proto/streams/state.rs
```diff
@@ -74,6 +74,11 @@ enum Peer {
 enum Cause {
     Proto(Reason),
     Io,
+
+    /// The user droped all handles to the stream without explicitly canceling.
+    /// This indicates to the connection that a reset frame must be sent out
+    /// once the send queue has been flushed.
+    Canceled,
```
seanmonstar: I think there's a few blanket matches that will match this case, like in `ensure_recv_open` down below...

carllerche: I fixed `ensure_recv_open ` but that is the only one that seemed problematic?

PR #160, opened 2017-10-20
Add test and assertion for idle state handling

### LINKED ISSUE #11: Beware: Idle states
I noticed [the following](http://httpwg.org/specs/rfc7540.html#StreamIdentifiers):

> The first use of a new stream identifier implicitly closes all streams in the "idle" state that might have been initiated by that peer with a lower-valued stream identifier. For example, if a client sends a HEADERS frame on stream 7 without ever sending a frame on stream 5, then stream 5 transitions to the "closed" state when the first frame for stream 7 is sent or received.

We need to be careful that we don't initialize streams in the idle state such that they may be moved into reserved/open out of order. We should probably only ever initialize streams directly to a Reserved or Open (or HalfOpenRemote) state.

hawkw:
I've added a unit test ensuring that we don't initialize streams in the idle state such that they may be moved into reserved/open out of order. I've also added an assertion in `Prioritize::pop_frame` that the stream ID of any headers frame that opens a new stream is greater than all previously sent stream IDs.

Closes #11

### DIFF
--- a/src/proto/streams/prioritize.rs
+++ b/src/proto/streams/prioritize.rs
@@ -628,12 +628,10 @@ impl Prioritize {
                     };
 
                     trace!("pop_frame; frame={:?}", frame);
-                    
-                    if let Frame::Headers(_) = frame {
-                        if stream.state.is_idle() {
-                            debug_assert!(stream.id > self.last_opened_id);
-                            self.last_opened_id = stream.id;
-                        }
+
+                    if stream.state.is_idle() {
+                        debug_assert!(stream.id > self.last_opened_id);
+                        self.last_opened_id = stream.id;
                     }
 
                     if !stream.pending_send.is_empty() || stream.state.is_canceled() {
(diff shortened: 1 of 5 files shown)


src/proto/streams/prioritize.rs
```diff
@@ -613,6 +626,11 @@ impl Prioritize {
 
                     trace!("pop_frame; frame={:?}", frame);
 
+                    if cfg!(debug_assertions) && stream.state.is_idle() {
```
carllerche: If we're going to store a field on the struct, we might as well make it an assert (not debug).

src/proto/streams/prioritize.rs
```diff
@@ -11,9 +11,18 @@ use bytes::buf::Take;
 use std::{cmp, fmt};
 use std::io;
 
+/// # Warning
+///
+/// Queued streans are ordered by stream ID, as we need to ensure that
```
carllerche: `streans`: typo.

src/proto/streams/prioritize.rs
hawkw: I wanted to make the field itself only part of the struct on debug, but since `if` statements can't have `#[cfg(...)]` attrs, the assertion needed to go in its own function, and at that point, I punted.

PR #162, opened 2017-10-20
Add test that recv.next_stream_id should be updated in `open`

hawkw:
As requested in https://github.com/carllerche/h2/issues/38#issuecomment-328254128, I've added a test to `stream_states.rs` to cover the case where we receive a HEADERS frame that gets rejected (due to being malformed), then the same stream ID is received with a valid HEADERS frame which gets accepted.


carllerche:
Oh no! Merge conflict! ⚔️

PR #166, opened 2017-10-21
Expose h2::Error::reason

olix0r:
When an h2 error code is known, applications may care to know about the
specific error. For example, an application may want to report per-code
metrics or handle `ENHANCE_YOUR_CALM` specially.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -34,9 +34,9 @@ impl Error {
     ///
     /// This is either an error received by the peer or caused by an invalid
     /// action taken by the peer (i.e. a protocol error).
-    pub fn reason(&self) -> Option<&Reason> {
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -29,6 +29,19 @@ enum Kind {
...
 // ===== impl Error =====
 
+impl Error {
+    /// If the error was caused by the remote peer, the error reason.
+    ///
+    /// This is either an error received by the peer or caused by an invalid
+    /// action taken by the peer (i.e. a protocol error).
+    pub fn reason(&self) -> Option<Reason> {
+        match self.kind {
+            Kind::Proto(ref reason) => Some(*reason),
```
seanmonstar: The `ref` and `*` aren't necessary, just as an FYI.

PR #172, opened 2017-10-30
client: return user error if uri is missing scheme or authority

seanmonstar:


carllerche:
So, just for completeness:

> To ensure that the HTTP/1.1 request line can be reproduced accurately, this pseudo-header field MUST be omitted when translating from an HTTP/1.1 request that has a request target in origin or asterisk form (see [RFC7230], Section 5.3).

carllerche:
So what do you suggest doing?

PR #173, opened 2017-11-13
reject connection-specific headers

### LINKED ISSUE #36: Strip connection level header fields
> Such intermediaries SHOULD also remove other connection-specific header fields, such as Keep-Alive, Proxy-Connection, Transfer-Encoding, and Upgrade, even if they are not nominated by the Connection header field.

If a header frame is sent that includes a `Connection` field, it should be stripped including all fields referenced by the `Connection` field value.

seanmonstar:
- When receiving, return a PROTOCOL_ERROR.
- When sending, return a user error about malformed headers.

While #36 mentions stripping the headers, this PR actually returns a `UserError` instead.

Closes #36

### DIFF
--- a/src/frame/headers.rs
+++ b/src/frame/headers.rs
@@ -635,7 +635,12 @@ impl HeaderBlock {
                     // Connection level header fields are not supported and must
                     // result in a protocol error.
 
-                    if name == header::CONNECTION {
+                    if name == header::CONNECTION
+                        || name == header::TRANSFER_ENCODING
+                        || name == header::UPGRADE
+                        || name == "keep-alive"
+                        || name == "proxy-connection"
+                    {
                         trace!("load_hpack; connection level header");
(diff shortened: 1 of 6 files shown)


src/frame/headers.rs
```diff
@@ -635,7 +635,12 @@ impl HeaderBlock {
                     // Connection level header fields are not supported and must
                     // result in a protocol error.
 
-                    if name == header::CONNECTION {
+                    if name == header::CONNECTION
+                        || name == header::TRANSFER_ENCODING
+                        || name == header::UPGRADE
+                        || name == "keep-alive"
```
carllerche: It seems like this means it would be worth for `http` to provide consts for these headers, otherwise this will require iterating the string and lowercasing each char (of the RHS).

But this isn't an immediate requirement. I created hyperium/http/issues/142 to track.

PR #174, opened 2017-11-15
ignore received frames on a stream locally reset

### LINKED ISSUE #32: Account for invalid frames against connection flow control window

### LINKED ISSUE #74: After sending RST_STREAM, ignore inbound frames for that stream
When sending a RST_STREAM, the peer may have in-flight frames. In which case, any frames received for the reset stream ID must be ignored for "some period of time".

This bug is exposed by #73.

Relates to #32

seanmonstar:
This adds a map of stream IDs that we have reset locally, since we need to remember even after we've freed the related `Stream`. Upon receiving a data frame, that map is also checked to see if the stream had been reset.

When we do decide to ignore a data frame, the connection flow control is still adjusted, since the sender had to have adjusted their understanding of this connection's window also.

Closes #74 
Closes #32

### DIFF
--- a/src/proto/streams/counts.rs
+++ b/src/proto/streams/counts.rs
@@ -19,6 +19,12 @@ pub(super) struct Counts {
 
     /// Current number of locally initiated streams
     num_recv_streams: usize,
+
+    /// Maximum number of pending locally reset streams
+    max_reset_streams: usize,
+
+    /// Current number of pending locally reset streams
+    num_reset_streams: usize,
 }
 
 impl Counts {
@@ -30,6 +36,8 @@ impl Counts {
             num_send_streams: 0,
             max_recv_streams: config.remote_max_initiated.unwrap_or(usize::MAX),
             num_recv_streams: 0,
+            max_reset_streams: config.local_reset_max,
+            num_reset_streams: 0,
         }
     }
 
@@ -72,6 +80,22 @@ impl Counts {
         self.num_send_streams += 1;
     }
 
+    /// Returns true if the number of pending reset streams can be incremented.
+    pub fn can_inc_num_reset_streams(&self) -> bool {
+        self.max_reset_streams > self.num_reset_streams
+    }
+
+    /// Increments the number of pending reset streams.
+    ///
+    /// # Panics
+    ///
+    /// Panics on failure as this should have been validated before hand.
+    pub fn inc_num_reset_streams(&mut self) {
+        assert!(self.can_inc_num_reset_streams());
+
+        self.num_reset_streams += 1;
+    }
+
     pub fn apply_remote_settings(&mut self, settings: &frame::Settings) {
         if let Some(val) = settings.max_concurrent_streams() {
             self.max_send_streams = val as usize;
@@ -87,19 +111,26 @@ impl Counts {
         F: FnOnce(&mut Self, &mut store::Ptr) -> U,
     {
         let is_counted = stream.is_counted();
+        let is_pending_reset = stream.is_pending_reset_expiration();
 
         // Run the action
         let ret = f(self, &mut stream);
 
-        self.transition_after(stream, is_counted);
+        self.transition_after(stream, is_counted, is_pending_reset);
 
         ret
     }
 
     // TODO: move this to macro?
-    pub fn transition_after(&mut self, mut stream: store::Ptr, is_counted: bool) {
+    pub fn transition_after(&mut self, mut stream: store::Ptr, is_counted: bool, is_reset_counted: bool) {
         if stream.is_closed() {
-            stream.unlink();
+            if !stream.is_pending_reset_expiration() {
+                stream.unlink();
+
+                if is_reset_counted {
+                    self.dec_num_reset_streams();
+                }
+            }
 
             if is_counted {
                 // Decrement the number of active streams.
@@ -115,9 +146,16 @@ impl Counts {
 
     fn dec_num_streams(&mut self, id: StreamId) {
         if self.peer.is_local_init(id) {
+            assert!(self.num_send_streams > 0);
             self.num_send_streams -= 1;
         } else {
+            assert!(self.num_recv_streams > 0);
             self.num_recv_streams -= 1;
         }
+    }
+
+    fn dec_num_reset_streams(&mut self) {
+        assert!(self.num_reset_streams > 0);
+        self.num_reset_streams -= 1;
     }
 }
--- a/src/proto/streams/prioritize.rs
+++ b/src/proto/streams/prioritize.rs
@@ -529,6 +529,11 @@ impl Prioritize {
                 Some(mut stream) => {
                     trace!("pop_frame; stream={:?}", stream.id);
 
+                    // It's possible that this stream, besides having data to send,
+                    // is also queued to send a reset, and thus is already in the queue
+                    // to wait for "some time" after a reset.
+                    //
+                    // To be safe, we just always ask the stream.
                     let is_counted = stream.is_counted();
                     let is_pending_reset = stream.is_pending_reset_expiration();
 
--- a/src/proto/streams/prioritize.rs
+++ b/src/proto/streams/prioritize.rs
@@ -530,6 +530,7 @@ impl Prioritize {
                     trace!("pop_frame; stream={:?}", stream.id);
 
                     let is_counted = stream.is_counted();
+                    let is_pending_reset = stream.is_pending_reset_expiration();
 
                     let frame = match stream.pending_send.pop_front(buffer) {
                         Some(Frame::Data(mut frame)) => {
@@ -651,7 +652,7 @@ impl Prioritize {
                         self.pending_send.push(&mut stream);
                     }
 
-                    counts.transition_after(stream, is_counted);
+                    counts.transition_after(stream, is_counted, is_pending_reset);
 
                     return Some(frame);
                 },
--- a/src/proto/streams/recv.rs
+++ b/src/proto/streams/recv.rs
@@ -245,7 +245,28 @@ impl Recv {
         Ok(())
     }
 
-    /// Releases capacity back to the connection
+    /// Releases capacity of the connection
+    fn release_connection_capacity(
+        &mut self,
+        capacity: WindowSize,
+        task: &mut Option<Task>,
+    ) {
+        trace!("release_connection_capacity; size={}", capacity);
+
+        // Decrement in-flight data
+        self.in_flight_data -= capacity;
+
+        // Assign capacity to connection
+        self.flow.assign_capacity(capacity);
+
+        if self.flow.unclaimed_capacity().is_some() {
+            if let Some(task) = task.take() {
+                task.notify();
+            }
+        }
+    }
+
+    /// Releases capacity back to the connection & stream
     pub fn release_capacity(
         &mut self,
         capacity: WindowSize,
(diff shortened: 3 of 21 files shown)


src/proto/streams/recv.rs
```diff
@@ -411,6 +415,19 @@ impl Recv {
...
+    pub fn consume_connection_window(&mut self, sz: WindowSize) -> Result<(), RecvError> {
+        if self.flow.window_size() < sz {
+            return Err(RecvError::Connection(Reason::FLOW_CONTROL_ERROR));
+        }
+
+        // Update connection level flow control
+        self.flow.send_data(sz);
+
+        // Track the data as in-flight
+        self.in_flight_data += sz;
```
seanmonstar: There is an open question of whether we should immediately release this connection capacity. If we don't, since there is no related stream to release the capacity, the connection capacity is essentially "leaked" forever. At the same time, if we locally reset this stream, but the remote just keeps sending us data anyways, giving capacity back would allow them to just keep flooding the connection with this stream we no longer want.

Perhaps if we add in the "for some time" part, by checking a time elapsed, then the eventual connection error would be a fine deterrent and allow us to free the connection window automatically.

Even without this question answered, it is still better that we count this anyway, because: it **is** correct that the window has been used, it **is not** required that we give the capacity back.

carllerche: IMO we should immediately release capacity. the window updates will be sequenced *after* the  RST_STREAM frame, so the remote should not send any more data.

carllerche: Additionally, we won't ever increment the reset stream's window, so even if the remote keeps sending data, the window of that stream will eventually get depleted.

src/proto/streams/recv.rs
```diff
@@ -368,19 +371,20 @@ impl Recv {
...
         // on the stream.
-        if self.flow.window_size() < sz || stream.recv_flow.window_size() < sz {
-            return Err(RecvError::Connection(Reason::FLOW_CONTROL_ERROR));
-        }
+        self.consume_connection_window(sz)?;
 
-        // Update connection level flow control
-        self.flow.send_data(sz);
+        if stream.recv_flow.window_size() < sz {
+            return Err(RecvError::Stream {
```
carllerche: This is surprising to me. If this check is violated, it indicates a buggy peer. Would you mind adding a comment referencing the spec explaining why this is a stream level error?

src/proto/streams/state.rs
```diff
@@ -57,22 +57,22 @@ enum Inner {
...
-    Closed(Option<Cause>),
+    Open { local: OpenPeer, remote: OpenPeer },
+    HalfClosedLocal(OpenPeer), // TODO: explicitly name this value
+    HalfClosedRemote(OpenPeer),
+    Closed(Cause),
 }
 
 #[derive(Debug, Copy, Clone)]
-enum Peer {
+enum OpenPeer {
```
carllerche: I'd rather avoid this rename (in this commit at least, we can approach it later). Also, see below.

src/proto/streams/state.rs
```diff
@@ -57,22 +57,22 @@ enum Inner {
...
     AwaitingHeaders,
     Streaming,
 }
 
 #[derive(Debug, Copy, Clone)]
 enum Cause {
-    Proto(Reason),
+    EndStream,
+    Proto(Peer, Reason),
```
carllerche: Instead of adding a `Peer` variant, could we add a `Cause::LocallyReset` variant. I believe that is the case that we really care about?

src/proto/streams/store.rs
```diff
@@ -12,6 +12,11 @@ use std::ops;
 pub(super) struct Store {
     slab: slab::Slab<(StoreId, Stream)>,
     ids: OrderMap<StreamId, (usize, StoreId)>,
+    /// Keeps a list of StreamIds that we have reset locally, to allow ignoring
+    /// frames on that stream "for some time".
+    // We could store an `Instant` here perhaps, and upon lookup, if the
+    // elapsed time has been too long, pop it from this Map and return false.
+    reset_streams: OrderMap<StreamId, ()>,
```
carllerche: Ideally we wouldn't introduce a second map. Would it be possible to track the locally reset state as part of the stream state and have a single lookup path?

It also looks like introducing a second map doubles the number of hash lookups for all received data frames (from below).

src/proto/streams/streams.rs
```diff
@@ -172,9 +172,19 @@ where
 
         let id = frame.stream_id();
 
+        if me.store.is_reset(&id) {
```
carllerche: As mentioned above, this doubles all hash lookups for received strams.

Also, I think that the stream window flow should still be validated? We just drop frames that the remote sent before seeing the RST_STREAM, we still want to ensure that the connection state is sane.

src/proto/streams/store.rs
seanmonstar: So, we *could* keep the streams around afterwards, but a `Stream` is rather big in bytes, and we only care about a) the ID, b) that it was reset locally, c) how long ago. Additionally, locally resetting shouldn't be that uncommon, but hopefully receiving frames on a reset stream is less common, and so having a second map is the better situation...

src/proto/streams/streams.rs
seanmonstar: If we do keep a second smaller map, then I can do some stupid syntax crap to fix the borrow checker. Since a mutable stream is taken from the store, even the `None` case can't access the store again...

seanmonstar: Regarding counting the stream window as well: since the stream is reset, the window should be closed.

> Flow-controlled frames (i.e., DATA) received after sending RST_STREAM are counted toward the connection flow-control window.

src/proto/streams/store.rs
carllerche: @seanmonstar As of now, I wouldn't worry about the byte size of `Stream`. We will be shrinking it significantly.

src/proto/streams/streams.rs
carllerche: Right, it needs to be counted against the connection flow control, but unless we keep the stream level flow state around too, we don't know if the received data was a valid size (exceeded the max size the stream would have accepted if it wasn't locally reset).

Basically, we can't differentiate between the remote peer being correct and just not having seen the RST_STREAM vs. the remote being buggy and violating flow control.

carllerche:
@seanmonstar Let me know if you can keep working on this or if I should take it over. I'd like to see this get over the finish line.

seanmonstar:
The changes should be largely all here:

- removed the second `OrderMap`, a new `Queue` has been added to store streams waiting to expire after a reset
- Adds a duration check to determine if a queued stream should be reaped yet
- Adds config to set the duration, and a max allowed pending expirations

I need to expose the config in the server builder, and resolve some merge conflicts...

src/proto/streams/counts.rs
```diff
@@ -87,19 +111,26 @@ impl Counts {
...
 
         ret
     }
 
     // TODO: move this to macro?
-    pub fn transition_after(&mut self, mut stream: store::Ptr, is_counted: bool) {
+    pub fn transition_after(&mut self, mut stream: store::Ptr, is_counted: bool, is_reset_counted: bool) {
         if stream.is_closed() {
-            stream.unlink();
+            if !stream.is_pending_reset_expiration() {
```
carllerche: I **think** that I'm following this logic (and that it is correct), but it would be super helpful to add a comment describing how `transition_after` works now that it is getting pretty involved.

src/proto/streams/prioritize.rs
```diff
@@ -530,6 +530,7 @@ impl Prioritize {
                     trace!("pop_frame; stream={:?}", stream.id);
 
                     let is_counted = stream.is_counted();
+                    let is_pending_reset = stream.is_pending_reset_expiration();
```
carllerche: This line was confusing at first. My initial thought was that, at this point, the stream could never be pending reset expiration. However, a stream transitions to `pending_reset_expiration` immediately when the client resets the stream. At this point, the `RST_STREAM` frame is still queued, and this is probably the point at which we are reading it.

So, in fact, it is actually expected for `is_pending_reset` to be true sometimes.

Could you add some comments describing this?

src/proto/streams/recv.rs
```diff
@@ -531,6 +570,34 @@ impl Recv {
...
+        if stream.is_pending_reset_expiration() {
+            return;
+        }
+
+        if !counts.can_inc_num_reset_streams() {
+            // try to evict 1 stream if possible
+            // if max allow is 0, this won't be able to evict,
+            // and then we'll just bail after
+            if let Some(evicted) = self.pending_reset_expired.pop(stream.store_mut()) {
+                let is_counted = evicted.is_counted();
```
carllerche: This line was confusing at first. My initial thought was that, at this point, the stream could never be "counted" as it has already been reset. However, a stream transitions to `pending_reset_expiration` immediately when the client resets the stream, but the stream transitions to "not counted" once the RST_STREAM frame is sent. So, there is a period in which the stream is in the "pending reset expired" queue while it is still counted. Under a specific workload, this line could be hit while the stream is still counted.

Could you add some comments describing this?

src/proto/streams/recv.rs
```diff
@@ -368,19 +381,33 @@ impl Recv {
 
         // Ensure that there is enough capacity on the connection before acting
         // on the stream.
-        if self.flow.window_size() < sz || stream.recv_flow.window_size() < sz {
-            return Err(RecvError::Connection(Reason::FLOW_CONTROL_ERROR));
+        self.consume_connection_window(sz)?;
+
+        if is_ignoring_frame {
+            trace!("recv_data frame being ignored on locally reset {:?} for some time", stream.id);
+            return Ok(());
```
carllerche: So the connection level window is consumed right above, but I don't see where it is "released". Specifically, how connection level window is not leaked in this case.

It would be nice to have a test for this.

[Related](https://github.com/carllerche/h2/pull/174/files#r151241765)

src/proto/streams/recv.rs
```diff
@@ -368,19 +381,33 @@ impl Recv {
...
+        if stream.recv_flow.window_size() < sz {
+            // http://httpwg.org/specs/rfc7540.html#WINDOW_UPDATE
+            // > A receiver MAY respond with a stream error (Section 5.4.2) or
+            // > connection error (Section 5.4.1) of type FLOW_CONTROL_ERROR if
+            // > it is unable to accept a frame.
+            //
+            // So, for violating the **stream** window, we can send either a
+            // stream or connection error. We've opted to send a stream
+            // error.
+            return Err(RecvError::Stream {
```
carllerche: Similar as above, we need to be very sure that if this code path is hit, the connection level window is not leaked. Again, it would be nice to have a test for this.

[Related](https://github.com/carllerche/h2/pull/174/files#r151241765)

src/proto/streams/streams.rs
```diff
@@ -172,9 +175,21 @@ where
 
         let id = frame.stream_id();
 
+        /*
```
carllerche: Should this comment block be removed?

src/proto/streams/recv.rs
seanmonstar: Looking around this function, it appears there's existing conditions that could cause a stream error after consuming the connection window, and in all of those cases, the window is "leaked".

To fix this, we could add a check outside this function, wherever `streams.recv_data()` is being called, and if a stream error is returned, try to release the connection capacity again...

src/proto/streams/recv.rs
```diff
@@ -385,6 +401,7 @@ impl Recv {
 
         if is_ignoring_frame {
             trace!("recv_data frame being ignored on locally reset {:?} for some time", stream.id);
+            self.release_connection_capacity(sz, &mut None);
```
carllerche: So, just to be extra clear (because, I am very tired and I got pretty confused). Could you add a comment that `self.release_connection_capacity` isn't exactly the opposite of `self.consume_connection_window(sz)?;`, and you still need to perform both fn calls to ensure that a `WINDOW_UPDATE` gets sent out?

carllerche:
While working through this PR, we uncovered some additional flow control issues. Since they are unrelated to this specific change, I created #183 to track.

carllerche:
PR looks good, feel free to merge whenever you are done w/ comments & all that.

Issue #19, opened 2017-08-11
Add hpack tests ensuring ranges are checked


carllerche:
In the HPACK decoder `peek_u8` panics when out of bound. Ensure ranges are checked before calling it.

PR #175, opened 2017-11-21
hpack/decoder: add bound check to peek_u8

lucab:
This updates `peek_u8` in hpack decoder to internally perform bound
checking. This ensures it cannot access bytes out of range once
reaching the end of the buffer.

Fixes: https://github.com/carllerche/h2/issues/19

### DIFF
--- a/src/hpack/decoder.rs
+++ b/src/hpack/decoder.rs
@@ -177,11 +177,11 @@ impl Decoder {
 
         trace!("decode");
 
-        while src.has_remaining() {
+        while let Some(ty) = peek_u8(src) {
             // At this point we are always at the beginning of the next block
             // within the HPACK data. The type of the block can always be
             // determined from the first byte.
-            match Representation::load(peek_u8(src))? {
+            match Representation::load(ty)? {
                 Indexed => {
                     trace!("    Indexed; rem={:?}", src.remaining());
                     can_resize = false;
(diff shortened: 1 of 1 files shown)


carllerche:
Thanks! Would you mind adding an associated test that covers this branch?

lucab:
Ack, added a few simple tests decoding an empty buffer. I am not sure whether this is what you had in mind. Also, I don't know if you want to keep unit-tests in the same file or somewhere else.

lucab:
Gentle ping for a review.

seanmonstar:
Perhaps I'm missing a tiny detail, but after reading over a few times, it seems to me that the code is the same? Both before and after, `has_remaining()` is checked before `bytes()[0]`, right?

lucab:
Right. Unless I misunderstood it, the original request in #19 was to ensure that `peek_u8` doesn't explode in the hands of the caller, whatever the input or the call place is (which sounds reasonable).

This just rearranges existing callsites so that the check and the indexed access are always performed together, even for future callers and reworks.

(I was just skimming through the repo and saw a low hanging fruit to pick. I don't know about the reasons of the original ticket, but I guess somebody made that mistake once and would like to prevent it for the future)

seanmonstar:
As I see. I don't know what exactly @carllerche meant with the original issue...

carllerche:
I created the issue when there *was* a bug caused by `peek_u8` being called w/o a remaining check. I originally just fixed that issue but filed #19 to investigate if any other problems could exist.

This changes `peek_u8` to be harder to misuse, so I'm good with it.

carllerche:
Thanks @lucab

PR #176, opened 2017-11-28
Notify stream refs when connection sees EOF

seanmonstar:


seanmonstar:
Ok, this works, and added a test. Without fix, the test hangs.

src/proto/connection.rs
```diff
@@ -277,8 +277,8 @@ where
                     // TODO: handle
                 },
                 None => {
-                    // TODO: Is this correct?
```
olix0r: :D

src/proto/streams/recv.rs
```diff
@@ -497,6 +497,11 @@ impl Recv {
         stream.notify_recv();
     }
 
+    pub fn recv_eof(&mut self, stream: &mut Stream) {
+        stream.state.recv_eof();
+        stream.notify_recv();
```
carllerche: Should this notify writes as well?

PR #178, opened 2017-12-01
notify stream send task when receiving EOF

seanmonstar:


carllerche:
@olix0r good question. All the various "poll" fns for the stream should be notified then fail.

seanmonstar:
Updated so that if `client.poll_ready()` is called after an EOF occurs, an error is returned.

PR #181, opened 2017-12-11
Client should validate request URI.

### LINKED ISSUE #179: Client sends requests with no scheme
If issuing a request with the URI set to a relative path (`"/foo"`), the client sends out an invalid request (no scheme).

@seanmonstar I vaguely remember we discussed this, but do not remember the conclusion.

carllerche:
This patch adds checks for the request URI and rejects invalid URIs. In
the case of forwarding an HTTP 1.1 request with a path, an "http" pseudo
header is added to satisfy the HTTP/2.0 spec.

`convert_send_message` is removed from the `Peer` trait as it is not
needed to be generic. The client version now returns an error if the request
is invalid.

This PR depends on https://github.com/hyperium/http/pull/147.

Closes #179

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,12 +1,12 @@
 //! HTTP2 client side.
 use {SendStream, RecvStream, ReleaseCapacity};
-use codec::{Codec, RecvError};
+use codec::{Codec, RecvError, SendError, UserError};
 use frame::{Headers, Pseudo, Reason, Settings, StreamId};
 use proto;
 
 use bytes::{Bytes, IntoBuf};
 use futures::{Async, Future, MapErr, Poll};
-use http::{Request, Response};
+use http::{uri, Request, Response, Method, Version};
 use tokio_io::{AsyncRead, AsyncWrite};
 use tokio_io::io::WriteAll;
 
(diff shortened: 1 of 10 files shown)


src/client.rs
```diff
@@ -359,35 +363,58 @@ impl Future for ResponseFuture {
...
+            //
+            if pseudo.authority.is_none() {
+                if version == Version::HTTP_2 {
+                    return Err(UserError::MissingUriSchemeAndAuthority.into());
+                } else {
+                    // This is acceptable as per the above comment. However,
+                    // HTTP/2.0 requires that a scheme is set. Since we are
+                    // forwarding an HTTP 1.1 request, the scheme is set to
+                    // "http".
+                    pseudo.set_scheme(uri::Scheme::HTTP);
```
seanmonstar: I hate the `:scheme` header, it's so worthless. Should we be defaulting to `http` if it's not present? Ah what's it matter anyways...

carllerche: I am only defaulting it to "http" when dealing w/ HTTP/1.1 requests, which really have no way of setting the scheme w/o setting an authority (because it doesn't make sense to?)

If sending an HTTP/2.0 request w/o a scheme + authority, the lib will reject it.

1.1 vs. 2.0 is determined by the version field on the request struct.

carllerche:
I pushed a new `http` crate and am restarting CI.

Issue #139, opened 2017-10-06
Polling trailers before data stream is exhausted panics


seanmonstar:
If you have a `Body`, and call `poll_trailers` while there is still data frames in the queue, it will panic. This may be a common enough thing to do that we may wish to change this behavior. Some things we could do:

- Return a `UserError`. You done goofed. Downside is that returning an `Err` usually means the future is dead, and can't be used anymore. Would that mean we just have to trash the state of this stream altogether? Unhappy unfun times.
- We *could* recover from this, by pushing the DATA frame back to the front of the queue, and return `Async::NotReady`. This might have subtly wrong cases, though. Returning `NotReady` means we should have a way to notify the task when it *is* ready... and when is that? When the user has later polled off all the remaining DATA frames? If they've done that, we still won't have a `Task` to notify...
- We could take this signal to mean the user doesn't care about the DATA frames, and we could just pop off the queue and drop them. If we do this, we need to make sure to release capacity while doing so.

carllerche:
Fixed by #189

PR #189, opened 2017-12-19
Don't panic on poll_trailers called out of order.

carllerche:
This patch removes a panic in `poll_trailers` that is triggered if
`poll_trailers` is called before `poll_data` returns `None`.

This is done by always trying to notify the receive task when
`poll_data` returns None and there already is pending trailers.


carllerche:
Fixes: #139

PR #195, opened 2017-12-19
add test that a window update can be received in reserved state

### LINKED ISSUE #33: Accept stream WINDOW_UPDATE frames in reserved state
In general, the various states in which frames can be accepted should be audited.

http://httpwg.org/specs/rfc7540.html#StreamStates

seanmonstar:
Closes #33


seanmonstar:
Fixed broken test.

PR #198, opened 2017-12-19
Support writing continuation frames.

carllerche:
Large header sets might require being split up across multiple frames.
This patch adds support for doing so.


carllerche:
I know it is surprising, but I think that I wrote a bug.

PR #202, opened 2017-12-22
Misc renames

### LINKED ISSUE #154: Rename `server::Server` -> `Accept` or `Listener`
This type is mostly used to acquire new H2 streams.

This is fairly low priority, but I thought I would propose the change to see what others thought about it.

carllerche:
Misc renames to try to increase consistency across `client` and `server` modules.

* `server::Server` -> `server::Connection`.
* `client::Client` -> `client::SendRequest` (there already is a `client::Connection`).
* `server::Respond` -> `server::SendResponse`.

This also moves the `builder` and `handshake` fns from `client::Client` -> `client::Connection` as well as change the type orders in the tuple returned by the `server::Handshake` future.

Fixes #154

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -150,6 +150,16 @@ where
 // ===== impl Builder =====
 
 impl Builder {
+    /// Creates a `Connection` Builder to customize a `Connection` before binding.
+    pub fn new() -> Builder {
+        Builder {
+            reset_stream_duration: Duration::from_secs(proto::DEFAULT_RESET_STREAM_SECS),
+            reset_stream_max: proto::DEFAULT_RESET_STREAM_MAX,
+            settings: Default::default(),
+            stream_id: 1.into(),
+        }
+    }
+
     /// Set the initial window size of the remote peer.
     pub fn initial_window_size(&mut self, size: u32) -> &mut Self {
         self.settings.set_initial_window_size(Some(size));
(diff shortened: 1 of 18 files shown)


src/client.rs
```diff
@@ -330,7 +325,7 @@ where
     B: IntoBuf,
     B::Buf: 'static,
 {
-    type Item = (Client<B>, Connection<T, B>);
+    type Item = (Connection<T, B>, SendRequest<B>);
```
seanmonstar: Why the change in order? Without this change, it reads as `let (client, conn) = await!(handshake(io))`, where `client` (now a `SendRequest`) is a sender, and `conn` is a thing to push into an executor. It matches somewhat the `let (tx, rx) = channel()` pattern.

With the new change to `SendRequest`, it would be `let (tx, conn) = await!(handshake(io))`.

carllerche: I mostly changed it because I've been following the pattern of `Foo::somefn -> (Self, ...)` where the `Self` type comes first.

Another option would be to move the `handshake` fn off of the `Connection` type, which I think may be more "rusty" for fns that return a tuple.

seanmonstar: I see the benefits of each:

- `let (conn, tx) = Connection::handshake(io)`: Less imports, follow `Foo::foo -> (Self, ...)` pattern
- `let (tx, conn) = Connection::handshake(io)`: Less imports, follow `let (tx, rx) =` pattern.
- `let (tx, conn) = client::handshake(io)`: Possibly an extra import, the `client` module, but doesn't break the previous two patterns.

carllerche: I would do either 1) or 3).

However, I also wouldn't try to match the `(tx, rx)` order much, mostly because the connection handle isn't really a receiver.

seanmonstar: You know, actually, the `conn` part is almost useless anyways. Is there user going to do anything to it because push it onto an executor? ... I guess by returning the future, someone can wrap it more easily than if they had to provide a custom executor... And it does emit a connection error if they occur...

carllerche:
After some discussion, I think we can improve it a bit.

carllerche:
For both client & server:

* Rename `Connection::builder` -> `Builder::new`.
* Rename `Connection::handshake` -> `[mod]::handshake`.

I also switched the items in the tuple yielded by `client::Handshake` back to what it was before.

PR #205, opened 2017-12-23
Add protocol version to response for client

nak3:
Since `convert_poll_message()` does not set protocol version as
HTTP/2.0, responses client received always display `version:
HTTP/1.1`. 
This patch adds `HTTP/2.0` to the response.

Fixes https://github.com/carllerche/h2/issues/204


carllerche:
Thanks. This looks good to me.

Unfortunately, CI is failing right now (due to another issue). I'll merge this once CI is fixed.

PR #206, opened 2017-12-28
SETTINGS_MAX_HEADER_LIST_SIZE

### LINKED ISSUE #18: Improve hpack decoding across continuation frames
Currently, it just buffers all continuation frames. It would be nice if it did something different!

### LINKED ISSUE #191: Ensure the header payload is capped
When a header frame is not final, continuation frames are expected to follow.

In this case, the decoder simply buffers the payload until all continuation frames are received. This buffering should have an upper bound to avoid buffering forever.

seanmonstar:
This, uh, grew into something far bigger than expected, but it turns out, all of it was needed to eventually support this correctly.

- Adds configuration to client and server to set [SETTINGS_MAX_HEADER_LIST_SIZE](http://httpwg.org/specs/rfc7540.html#SETTINGS_MAX_HEADER_LIST_SIZE)
- If not set, a "sane default" of 16 MB is used (taken from golang's http2)
- Decoding header blocks now happens as they are received, instead of buffering up possibly forever until the last continuation frame is parsed.
- As each field is decoded, it's undecoded size is added to the total. Whenever a header block goes over the maximum size, the `frame` will be marked as such.
- Whenever a header block is deemed over max limit, decoding will still continue, but new fields will not be appended to `HeaderMap`. This is also can save wasted hashing.
- To protect against enormous string literals, such that they span multiple continuation frames, a check is made that the combined encoded bytes is less than the max allowed size. While technically not exactly what the spec suggests (counting decoded size instead), this should hopefully only happen when someone is indeed malicious. If found, a `GOAWAY` of `COMPRESSION_ERROR` is sent, and the connection shut down.
- After an oversize header block frame is finished decoding, the streams state machine will notice it is oversize, and handle that.
  - If the local peer is a server, a 431 response is sent, as suggested by the spec.
  - A `REFUSED_STREAM` reset is sent, since we cannot actually give the stream to the user.
- In order to be able to send both the 431 headers frame, and a reset frame afterwards, the scheduled `Canceled` machinery was made more general to a `Scheduled(Reason)` state instead.

Closes #18 
Closes #191

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ unstable = []
 
 [dependencies]
 futures = "0.1"
-tokio-io = "0.1.3"
+tokio-io = "0.1.4"
 bytes = "0.4"
 http = "0.1"
 byteorder = "1.0"
--- a/src/client.rs
+++ b/src/client.rs
@@ -172,6 +172,12 @@ impl Builder {
         self
     }
 
+    /// Set the max size of received header frames.
+    pub fn max_header_list_size(&mut self, max: u32) -> &mut Self {
+        self.settings.set_max_header_list_size(Some(max));
+        self
+    }
+
     /// Set the maximum number of concurrent streams.
     ///
     /// Clients can only limit the maximum number of streams that that the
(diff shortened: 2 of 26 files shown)


seanmonstar:
Well, CI half passes now. The failure is that nightly cannot compile ring 0.12, because of some module naming thing.

PR #210, opened 2018-01-02
send reset CANCEL when SendStream is dropped with no end-of-stream sent

### LINKED ISSUE #138: Handle `Stream` dropped before end-of-stream flag sent
This is a continuation of [a comment](https://github.com/carllerche/h2/pull/109/files#r143109861).

I believe that currently, if a `Stream` is dropped without sending an end-of-stream flag, nothing happens. This would cause the stream to enter a hung state.

seanmonstar:
Closes #138

### DIFF
--- a/src/proto/streams/prioritize.rs
+++ b/src/proto/streams/prioritize.rs
@@ -266,6 +266,27 @@ impl Prioritize {
         Ok(())
     }
 
+    /// Reclaim all capacity assigned to the stream and re-assign it to the
+    /// connection
+    pub fn reclaim_all_capacity(&mut self, stream: &mut store::Ptr) {
+        let available = stream.send_flow.available().as_size();
+        stream.send_flow.claim_capacity(available);
+        // Re-assign all capacity to the connection
+        self.assign_connection_capacity(available, stream);
+    }
+
+    /// Reclaim just reserved capacity, not buffered capacity, and re-assign
+    /// it to the connection
+    pub fn reclaim_reserved_capacity(&mut self, stream: &mut store::Ptr) {
+        // only reclaim requested capacity that isn't already buffered
+        if stream.requested_send_capacity > stream.buffered_send_data {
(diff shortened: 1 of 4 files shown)


src/proto/streams/send.rs
```diff
@@ -155,14 +155,15 @@ impl Send {
...
+        trace!("schedule_cancel; {:?}", stream.id);
         if stream.state.is_closed() {
             // Stream is already closed, nothing more to do
             return;
         }
 
         stream.state.set_canceled();
 
-        self.reclaim_capacity(stream);
+        self.prioritize.reclaim_reserved_capacity(stream);
```
seanmonstar: Without this change, the implicit cancel was clearing the capacity of the buffered data frame, and it was hanging again.

src/proto/streams/prioritize.rs
```diff
@@ -266,6 +266,27 @@ impl Prioritize {
...
+        stream.send_flow.claim_capacity(available);
+        // Re-assign all capacity to the connection
+        self.assign_connection_capacity(available, stream);
+    }
+
+    /// Reclaim just reserved capacity, not buffered capacity, and re-assign
+    /// it to the connection
+    pub fn reclaim_reserved_capacity(&mut self, stream: &mut store::Ptr) {
+        // only reclaim requested capacity that isn't already buffered
+        if stream.requested_send_capacity > stream.buffered_send_data {
```
carllerche: IIRC, `requested_send_capacity` is capacity that has been **requested** but not actually allocated to the stream, i.e., there was not enough connection capacity to provide.

I would think that you would need to check `stream.send_flow.available()` just like above, but then subtracting `buffered_send_data` from that?

If I am correct, we'd probably want to add a test to catch this case.

seanmonstar: I don't think that is correct, it looks like here that `requested >= buffered`: https://github.com/carllerche/h2/blob/master/src/proto/streams/prioritize.rs#L146

PR #212, opened 2018-01-04
Client documentation

carllerche:


src/client.rs
```diff
@@ -1,4 +1,85 @@
...
+//!
+//! # Making requests
+//!
+//! Requests are made using the [`SendRequest`] handle provided by the handshake
+//! future. Once the request is submitted, an HTTP/2.0 stream is initialized and
+//! the request is sent to the server.
+//!
+//! A request body and request trailers are sent using [`SendRequest`] and the
+//! server's response is returned once the [`ResponseFuture`] future completes.
+//! Both the [`SendRequest`] and [`ResponseFuture`] instances are returned by
```
rrichardson: It looks like the [`SendStream`] instance is returned from send_request.

src/client.rs
```diff
@@ -1,4 +1,156 @@
...
+//! details.
+//!
+//! This could be as basic as using Tokio's [`TcpStream`] to connect to a remote
+//! host, but usually it means using either ALPN or HTTP/1.1 protocol upgrades.
+//!
+//! Once a connection is obtained, it is passed to [`handshake`], which will
+//! begin the [HTTP/2.0 handshake]. This returns a future that completes once
+//! the handshake process is performed and HTTP/2.0 streams may be initialized.
+//!
+//! [`handshake`] uses default configuration values. THere are a number of
```
olix0r: `THere` -> `There`

src/client.rs
```diff
@@ -1,4 +1,156 @@
...
+//! Once the the handshake future completes, the caller is provided with a
+//! [`Connection`] instance and a [`SendRequest`] instance. The [`Connection`]
+//! instance is used to drive the connection (see [Managing the connection]).
+//! The [`SendRequest`] instance is used to initialize new streams (see [Making
+//! requests]).
+//!
+//! # Making requests
+//!
+//! Requests are made using the [`SendRequest`] handle provided by the handshake
+//! future. Once the request is submitted, an HTTP/2.0 stream is initialized and
```
olix0r: nit: Once _a_ request is submitted

src/client.rs
```diff
@@ -1,4 +1,156 @@
...
+//!
+//! A request body and request trailers are sent using [`SendRequest`] and the
+//! server's response is returned once the [`ResponseFuture`] future completes.
+//! Both the [`SendStream`] and [`ResponseFuture`] instances are returned by
+//! [`SendRequest::send_request`] and are tied to the HTTP/2.0 stream
+//! initialized by the sent request.
+//!
+//! The [`MAX_CONCURRENT_STREAMS`] setting is enforced by [`SendRequest`]. A
+//! request cannot be sent if the number of currently active streams has reached
+//! the maximum permitted for the connection.
```
olix0r: Does this result in a UserError?  Is the user responsible for ensuring this or does the cilent enforce this concurrency with buffering?

My point is _A request cannot be sent_ is not all that clear, whereas _Users must not send_ etc, would be a bit clearer to me.

olix0r: Perhaps this is all clearer if this note comes after the `poll_ready` section?

src/client.rs
```diff
@@ -1,4 +1,156 @@
...
+//! The [`Connection`] instance is used to manage connection state. The caller
+//! is required to call [`Connection::poll`] in order to advance state.
+//! [`SendRequest::send_request`] and other functions have no effect unless
+//! [`Connection::poll`] is called.
+//!
+//! The [`Connection`] instance should only be dropped once [`Connection::poll`]
+//! returns `Ready`. At this point, the underlying socket has been closed and no
+//! further work needs to be done.
+//!
+//! The easiest is to just submit the [`Connection`] instance to an [executor].
```
olix0r: The easiest what? This sentence needs a subject.

src/client.rs
```diff
@@ -14,25 +166,101 @@ use std::fmt;
...
+///
+/// [module]: index.html
+/// [`Connection`]: struct.Connection.html
 pub struct SendRequest<B: IntoBuf> {
     inner: proto::Streams<B::Buf, Peer>,
     pending: Option<proto::StreamKey>,
 }
 
-/// A future to drive the H2 protocol on a connection.
+/// Manages all state associated with an HTTP/2.0 client connection.
```
olix0r: Each short-description uses a slightly different conjugation/structure -- i think it would be clearer to unify these

```
/// In progress HTTP/2.0 connection handshake future.
```

```
/// Initialize new HTTP/2.0 streams on a connection by sending a request.
```

```
/// Manages all state associated with an HTTP/2.0 client connection.
```

I suspect the first two should be changed to the form:

```
/// Holds the state of a pending HTTP/2.0 connection handshake.
```

```
/// Initializes new HTTP/2.0 streams on a connection by sending a request. 
```

... based on what I've seen in other rustdocs...

src/client.rs
```diff
@@ -43,7 +271,44 @@ pub struct ResponseFuture {
     inner: proto::OpaqueStreamRef,
 }
 
-/// Build a client.
+/// Client connection factory, which can be used in order to configure the
+/// properties of the HTTP/2.0 client before it is created.
```
olix0r: tioli: _Builds client connections..._ (conjugation)

carllerche: I'm not sure what the proposed change is. Substituting "Client connection factory" w/ `Builds client connections"?

carllerche: ah, I see (based on prior comment).

carllerche:
I'd like to get #215 merged to make writing correct examples a bit easier.

src/client.rs
```diff
@@ -87,11 +378,171 @@ where
...
+    ///
+    /// See [module] level docs for more details.
+    ///
+    /// [module]: index.html
     pub fn ready(self) -> ReadySendRequest<B> {
         ReadySendRequest { inner: Some(self) }
     }
 
-    /// Send a request on a new HTTP 2.0 stream
+    /// Send a HTTP/2.0 request to the server.
```
olix0r: Sends

src/client.rs
```diff
@@ -184,7 +635,33 @@ where B: IntoBuf,
 // ===== impl Builder =====
 
 impl Builder {
-    /// Creates a `Connection` Builder to customize a `Connection` before binding.
+    /// Return a new client builder instance initialized with default
```
olix0r: Returns

src/client.rs
```diff
@@ -214,37 +755,185 @@ impl Builder {
 
     /// Set the maximum number of concurrent streams.
```
olix0r: sets

src/client.rs
```diff
@@ -214,37 +755,185 @@ impl Builder {
...
+    /// # }
+    /// #
+    /// # pub fn main() {}
+    /// ```
     pub fn max_concurrent_streams(&mut self, max: u32) -> &mut Self {
         self.settings.set_max_concurrent_streams(Some(max));
         self
     }
 
     /// Set the maximum number of concurrent locally reset streams.
```
olix0r: sets

src/client.rs
```diff
@@ -214,37 +755,185 @@ impl Builder {
...
+    /// # }
+    /// #
+    /// # pub fn main() {}
+    /// ```
     pub fn max_concurrent_reset_streams(&mut self, max: usize) -> &mut Self {
         self.reset_stream_max = max;
         self
     }
 
     /// Set the maximum number of concurrent locally reset streams.
```
olix0r: sets

src/client.rs
```diff
@@ -214,37 +755,185 @@ impl Builder {
...
+    /// #
+    /// # pub fn main() {}
+    /// ```
     pub fn reset_stream_duration(&mut self, dur: Duration) -> &mut Self {
         self.reset_stream_duration = dur;
         self
     }
 
-    /// Enable or disable the server to send push promises.
+    /// Enable or disable server push promises.
```
olix0r: Enables or disables

src/client.rs
```diff
@@ -261,13 +950,67 @@ impl Builder {
         self
     }
 
-    /// Bind an H2 client connection.
+    /// Create a new configured HTTP/2.0 client backed by `io`.
```
olix0r: Creates

src/client.rs
```diff
@@ -284,13 +1027,46 @@ impl Default for Builder {
     }
 }
 
-/// Bind an H2 client connection.
+/// Create a new configured HTTP/2.0 client with default configuration
```
olix0r: Creates

src/server.rs
```diff
@@ -209,7 +214,7 @@ pub struct Connection<T, B: IntoBuf> {
     connection: proto::Connection<T, Peer, B>,
 }
 
-/// Client connection factory, which can be used in order to configure the
+/// Server connection factory, which can be used in order to configure the
 /// properties of the HTTP/2.0 server before it is created.
```
olix0r: should this be more analogous to the client docs?

> Creates a new configured HTTP/2.0 server with default configuration...

or whatever

src/server.rs
```diff
@@ -470,7 +477,7 @@ where
 // ===== impl Builder =====
 
 impl Builder {
-    /// Return a new client builder instance initialized with default
+    /// Return a new server builder instance initialized with default
```
olix0r: Returns

src/server.rs
```diff
@@ -590,9 +597,9 @@ impl Builder {
     /// Set the maximum number of concurrent streams.
```
olix0r: Sets

src/server.rs
```diff
@@ -765,9 +772,9 @@ impl Builder {
     /// # {
     /// // `server_fut` is a future representing the completion of the HTTP/2.0
     /// // handshake.
-    /// let handshake_fut = Builder::new()
+    /// let server_fut = Builder::new()
```
olix0r: i can't comment on the top of this block but the intro maybe better as _Configures the send-payload data type._

src/client.rs
```diff
@@ -1,4 +1,165 @@
...
+//! the request is sent to the server.
+//!
+//! A request body and request trailers are sent using [`SendRequest`] and the
+//! server's response is returned once the [`ResponseFuture`] future completes.
+//! Both the [`SendStream`] and [`ResponseFuture`] instances are returned by
+//! [`SendRequest::send_request`] and are tied to the HTTP/2.0 stream
+//! initialized by the sent request.
+//!
+//! The [`SendRequest::poll_ready`] function returns `Ready` when a new HTTP/2.0
+//! stream can be created, i.e. as long as the current number of active streams
```
hawkw: I believe the "i.e." clause should be parenthesized, as in "when a new HTTP/2.0 stream can be created (i.e. as long as the current number of active streams is below [`MAX_CONCURRENT_STREAMS`].)"

src/client.rs
```diff
@@ -1,4 +1,165 @@
...
+//! The [`Connection`] instance is used to manage connection state. The caller
+//! is required to call [`Connection::poll`] in order to advance state.
+//! [`SendRequest::send_request`] and other functions have no effect unless
+//! [`Connection::poll`] is called.
+//!
+//! The [`Connection`] instance should only be dropped once [`Connection::poll`]
+//! returns `Ready`. At this point, the underlying socket has been closed and no
+//! further work needs to be done.
+//!
+//! The easiest option is to submit the [`Connection`] instance to an
```
hawkw: I feel like "The easiest option..." could be made a little clearer - "The easiest way to `poll` a `Connection`..."?

src/client.rs
```diff
@@ -14,15 +175,47 @@ use std::fmt;
...
     inner: WriteAll<T, &'static [u8]>,
     _marker: PhantomData<B>,
 }
 
-/// Marker type indicating a client peer
+/// Initializes new HTTP/2.0 streams on a connection by sending a request.
+///
+/// This type does no work itself. Instead, it is a handle to the inner
+/// connection state held by [`Connection`]. If the associated connection
+/// instance is dropped, all `SendRequest` functions will error.
```
hawkw: Maybe note what error is produced here?

src/client.rs
```diff
@@ -35,11 +228,65 @@ pub struct ReadySendRequest<B: IntoBuf> {
...
+/// A `Connection` is backed by an I/O resource (usually a TCP socket) and
+/// implements the HTTP/2.0 client logic for that connection. It is responsible
+/// for driving the internal state forward, performing the work requested of the
+/// associated handles ([`SendRequest`], [`ResponseFuture`], [`SendStream`],
+/// [`RecvStream`]).
+///
+/// `Connection` values are created by calling [`handshake`]. Once a
+/// `Connection` value is obtained, the caller must repeatedly call [`poll`]
+/// until `Ready` is returned. The easiest way to do this is to submit the
+/// `Connection` instance to an executor.
```
hawkw: Link `executor` here?

src/client.rs
hawkw: Also, TIOLI, but I'm not sure if it's worth saying that this is the easiest option if you're not going to enumerate any alternatives?

src/client.rs
```diff
@@ -200,7 +201,7 @@ pub struct Handshake<T, B: IntoBuf = Bytes> {
 ///
 /// This type does no work itself. Instead, it is a handle to the inner
 /// connection state held by [`Connection`]. If the associated connection
-/// instance is dropped, all `SendRequest` functions will error.
+/// instance is dropped, all `SendRequest` functions will return [`Error`].
```
hawkw: Is there a specific `Error` subtype returned in this case?

src/client.rs
```diff
@@ -57,9 +57,10 @@
 //! returns `Ready`. At this point, the underlying socket has been closed and no
 //! further work needs to be done.
 //!
-//! The easiest option is to submit the [`Connection`] instance to an
-//! [executor]. The executor will then manage polling the connection until the
-//! connection is complete.
+//! The easiest way to ensure that the [`Connection`] instance gets polled is to
+//! submit the [`Connection`] instance to an [executor]. The executor will then
+//! manage polling the connection until the connection is complete.
+//! Alternatively, the caller can call `poll` manually.
```
hawkw: This might be too fine-grained, so TIOLI, but should we give any reason _why_ you might `poll` manually? I  feel like, when something is typically only done in a few specific cases, enumerating what those cases might be is a good way to discourage users from doing that thing when it's not necessary. If that makes sense.

PR #215, opened 2018-01-11
Implement SendRequest::ready.

carllerche:
This provides a functional future API for waiting for SendRequest
readiness. This follows the pattern found in [tower].

[tower]: https://github.com/tower-rs/tower/blob/master/src/lib.rs#L242


carllerche:
I'm adding this now for #212 to help write correct examples a bit more succinctly.

PR #216, opened 2018-01-11
Start writing docs for shared types

carllerche:


src/share.rs
```diff
@@ -8,13 +8,91 @@ use http::{HeaderMap};
...
 
-/// Receive the body stream and trailers from the peer.
+/// Receives the body stream and trailers from the remote peer.
+///
+/// A `RecvStream` is provided by [`client::ResponseFuture`] and
+/// [`server::Connection`] with the received HTTP/2.0 message head (the response
+/// and request head respectively).
+///
+/// It is used to receive the streaming message body from the remote peer,
+/// receive any trailers, as well as to manage inbound flow control.
```
hawkw: Consider "It is used to receive the streaming message body from the remote peer, _and to_ receive any trailers", or "It is used to receive the streaming message body and any trailers from the remote peer"

hawkw: Also, TIOLI, but  maybe "It is used" -> "A`RecvStream` is used"?

src/share.rs
```diff
@@ -8,13 +8,91 @@ use http::{HeaderMap};
...
+/// **NOTE**: There is no bound on the amount of data that the library will
+/// buffer. If you are sending large amounts of data, you really should hook
+/// into the flow control lifecycle. Otherwise, you risk using up significant
+/// amounts of memory.
+///
+/// To hook into the flow control lifecycle, the caller signals to the library
+/// that it intends to send data by calling [`reserve_capacity`], specifying the
+/// amount of data, in octets, that the caller intends to send. After this,
+/// `poll_capacity` is used to be notified when the requested capacity is
+/// assigned to the stream. Once `poll_capacity` returns `Ready` with the number
```
hawkw: Maybe link`poll_capacity`?

src/error.rs
```diff
@@ -5,7 +5,16 @@ use std::{error, fmt, io};
 
 pub use frame::Reason;
 
-/// The error type for HTTP/2 operations
+/// Represents HTTP/2.0 operation errors.
+///
+/// `Error` covers error cases raised by protocol errors caused by the
+/// peer, I/O (transport) errors, and errors caused by the user of the library.
```
hawkw: TIOLI, but maybe "protocol errors caused by the peer", "I/O (transport) errors", and "errors caused by the user of the library" should mention and/or link to the types that represent those individual error cases?

src/frame/reason.rs
```diff
@@ -1,11 +1,21 @@
...
+/// HTTP/2.0 error codes.
+///
+/// Error codes are used in `RST_STREAM` and `GOAWAY` frames to convey the
+/// reasons for the stream or connection error. For example,
+/// [`SendStream::send_reset`] takes a `Reason` argument. Also, the `Error` type
+/// may contain a `Reason`.
+///
+/// Error codes share a common code space. Some error codes apply only to either
+/// streams or the entire connection and have no defined semantics in the other
+/// context.
```
olix0r: I get what you're trying to say but this sentence is a little confusing.

Maybe clearer:

> Error codes share a common code space. Some error codes apply only to streams, others apply only to connections, and others may apply to either.  See [RFC 7540] for more information.

src/error.rs
carllerche: All of these cases are not differentiated at the user API level. The error cause is pretty opaque as of now.

src/share.rs
```diff
@@ -8,13 +8,91 @@ use http::{HeaderMap};
...
+/// [`SendResponse`]: server/struct.SendResponse.html
+/// [specification]: http://httpwg.org/specs/rfc7540.html#FlowControl
+/// [`reserve_capacity`]: #method.reserve_capacity
 #[derive(Debug)]
 pub struct SendStream<B: IntoBuf> {
     inner: proto::StreamRef<B::Buf>,
 }
 
-/// Receive the body stream and trailers from the peer.
+/// Receives the body stream and trailers from the remote peer.
```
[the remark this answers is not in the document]
carllerche: Added the note

src/share.rs
```diff
@@ -8,13 +8,91 @@ use http::{HeaderMap};
...
+/// further data can be sent on any stream for that connection.
+///
+/// When the remote peer is ready to receive more data, it sends `WINDOW_UPDATE`
+/// frames. These frames increment the windows. See the [specification] for more
+/// details on the principles of HTTP/2.0 flow control.
+///
+/// The implications for sending data are that the caller **should** ensure that
+/// both the stream and the connection has available window capacity before
+/// loading the data to send into memory. The `SendStream` instance provides the
+/// necessary APIs to perform this logic. This, however, is not an oblication.
```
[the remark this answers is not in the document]
carllerche: The APIs are discussed in the following paragraphs. I can still link them if you think it is helpful.

src/share.rs
carllerche: Applied.

Issue #221, opened 2018-02-06
h2::Connection never returns Ready


jwilm:
Running `cargo run --example akamai` should exit as soon as the response is processed, but the program hangs until presumably the peer hangs up.

carllerche:
Closed by #222.

Issue #69, opened 2017-09-08
Provide a way to shutdown a connection on quiescence


carllerche:
Both the server and the client need a way to initiate a clean shutdown. The process should be:

* Stop accepting new streams
* Wait until all existing streams are complete
* Send a GO_AWAY frame
* Close the connection.

teohhanhui:
Isn't this already supported? #245 #250

jorendorff:
This issue is still referenced from the docs. @seanmonstar Is any part of this still open or did you fix all this 5+ years ago?

gumpt:
I have the same question as above. Don't love to be commenting on a thread to say that, but unsure how to raise a racket that would receive a response otherwise.

PR #222, opened 2018-02-06
When Streams are dropped, close Connection (#221)

DarrenTsung:
When all Streams are dropped / finished, the Connection was held
open until the peer hangs up. Instead, the Connection should hang up
once it knows that nothing more will be sent.

To fix this, we notify the Connection when a stream is no longer
referenced. On the Connection poll(), we check that there are no
active, held, reset streams or any references to the Streams
and transition to sending a GOAWAY if that is case.

These changes cause the example to work as expected (exiting when finished getting the response), but break other tests that are not expecting the connection to hang out.

### DIFF
--- a/src/proto/streams/counts.rs
+++ b/src/proto/streams/counts.rs
@@ -47,7 +47,7 @@ impl Counts {
     }
 
     pub fn has_streams(&self) -> bool {
-        self.num_send_streams != 0 || self.num_recv_streams != 0 || self.num_reset_streams != 0
+        self.num_recv_streams != 0 || self.num_reset_streams != 0
     }
 
     /// Returns true if the receive stream concurrency can be incremented
--- a/src/proto/streams/counts.rs
+++ b/src/proto/streams/counts.rs
@@ -46,6 +46,10 @@ impl Counts {
         self.peer
     }
 
+    pub fn has_streams(&self) -> bool {
+        self.num_send_streams != 0 || self.num_recv_streams != 0 || self.num_reset_streams != 0
+    }
+
     /// Returns true if the receive stream concurrency can be incremented
     pub fn can_inc_num_recv_streams(&self) -> bool {
         self.max_recv_streams > self.num_recv_streams
--- a/src/proto/streams/counts.rs
+++ b/src/proto/streams/counts.rs
@@ -47,7 +47,7 @@ impl Counts {
     }
 
     pub fn has_streams(&self) -> bool {
-        self.num_recv_streams != 0 || self.num_reset_streams != 0
+        self.num_send_streams != 0 || self.num_recv_streams != 0
     }
 
     /// Returns true if the receive stream concurrency can be incremented
--- a/src/proto/streams/streams.rs
+++ b/src/proto/streams/streams.rs
@@ -969,7 +969,7 @@ fn drop_stream_ref(inner: &Mutex<Inner>, key: store::Key) {
     // of canceling the stream), we should notify the task
     // (connection) so that it can close properly
     if stream.ref_count == 0 && stream.is_closed() {
-        if let Some(ref task) = actions.task.as_ref() {
+        if let Some(task) = actions.task.take() {
             task.notify();
         }
     }
--- a/src/proto/streams/streams.rs
+++ b/src/proto/streams/streams.rs
@@ -1,10 +1,10 @@
-use super::{Buffer, Config, Counts, Prioritized, Recv, Send, Stream, StreamId};
-use super::recv::RecvHeaderBlockError;
-use super::store::{self, Entry, Resolve, Store};
 use {client, proto, server};
 use codec::{Codec, RecvError, SendError, UserError};
 use frame::{self, Frame, Reason};
 use proto::{peer, Peer, WindowSize};
+use super::{Buffer, Config, Counts, Prioritized, Recv, Send, Stream, StreamId};
+use super::recv::RecvHeaderBlockError;
+use super::store::{self, Entry, Resolve, Store};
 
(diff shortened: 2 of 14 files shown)


carllerche:
The PR looks good at a glance, but there is a complication.

Re the tests, I pushed a commit (7613406d1dd84d0f2c69511687efd35a423b707c) that includes a fix for some of the tests. The issue with the client tests is that they drop the `SendRequest` handle early which causes the `GO_AWAY` frame to be sent. The fix is to hold on to the handle until the end of the test. Hopefully the commit I pushed illustrates that well enough.

Now, the complication.

It turns out that the server half uses the `Connection` to accept new streams, so it should be fine to hold on to the server connection without any outstanding refs.

I think what will need to happen is the logic to actually call `transition_to_go_away` is moved to `server::Connection` and `client::Connection`. On the client side, the same logic that you has applies. However, on the server side, the transition to go away will need to be an explicit call on `Connection`.

This relates to #69

DarrenTsung:
Thanks for the examples on how to fix the tests, I pushed a commit fixing the rest of the tests as well as splitting the logic between `server` / `client` such that `client` closes connection automatically and `server` has a close_connection() call that transitions the `server` to the GOAWAY state.

Let me know what you think!

tests/server.rs
```diff
@@ -204,6 +204,32 @@ fn sends_reset_cancel_when_req_body_is_dropped() {
     srv.join(client).wait().expect("wait");
 }
 
+#[test]
+fn sends_goaway_when_serv_closes_connection() {
```
seanmonstar: I know you didn't originally intend to add graceful shutdown to the server, but it looks like this adds it! It'd be bonus points if there could be another test showing multiple requests being sent and accepted, and then `close_connection()` called before the responses (and their bodies?) are finished, that the outstanding streams can finish before the connection terminates.

(I suppose we'd also need to do something about any new streams sent after sending the goaway, I don't recall what the spec says to do. `REFUSED_STREAM`?)

jwilm: The spec is slightly confusing on this matter. In one case, it says:

> Receivers of a GOAWAY frame MUST NOT open additional streams on the connection

But it doesn't mention a protocol error. It seems to suggest those new streams can simply be discarded:

> After sending a GOAWAY frame, the sender can discard frames for streams initiated by the receiver with identifiers higher than the identified last stream

With a caveat that data frames still need to count towards congestion flow control.

DarrenTsung: I added a test for showing a request being accepted, then the server closing the connection on the next request (and subsequently ignoring the third request).

Not sure how to make sure any properties like congestion flow control still hold true, please let me know if any changes need to be made!

DarrenTsung:
Are there any other changes required?

seanmonstar:
Yikes, I forgot about this, sorry! I'll take a look again.

src/proto/streams/counts.rs
```diff
@@ -46,6 +46,10 @@ impl Counts {
         self.peer
     }
 
+    pub fn has_streams(&self) -> bool {
+        self.num_send_streams != 0 || self.num_recv_streams != 0 || self.num_reset_streams != 0
```
seanmonstar: This probably doesn't need to check `num_reset_streams`, as that's just a count of "stream metadata we are holding on to temporarily to allow us to ignore incoming frames on a stream we just reset". The count is to just allow us to cap how many of those extra resources we keep in memory, but shouldn't affect if the connection is done, I believe.

src/proto/streams/streams.rs
```diff
@@ -951,6 +964,16 @@ fn drop_stream_ref(inner: &Mutex<Inner>, key: store::Key) {
 
     let actions = &mut me.actions;
 
+    // If the stream is not referenced and it is already
+    // closed (does not have to go through logic below
+    // of canceling the stream), we should notify the task
+    // (connection) so that it can close properly
+    if stream.ref_count == 0 && stream.is_closed() {
+        if let Some(ref task) = actions.task.as_ref() {
```
seanmonstar: This should probably be taken (`actions.task.take()`) as well. At least, that's the pattern that has been done through the lib whenever notifying on an `Option<Task>`.

carllerche:
LGTM, I haven't noticed anything besides what @seanmonstar has called out.

DarrenTsung:
Updated with all your feedback, let me know!

src/proto/streams/counts.rs
```diff
@@ -47,7 +47,7 @@ impl Counts {
     }
 
     pub fn has_streams(&self) -> bool {
-        self.num_send_streams != 0 || self.num_recv_streams != 0 || self.num_reset_streams != 0
+        self.num_recv_streams != 0 || self.num_reset_streams != 0
```
carllerche: I think it was `num_reset_streams` that was supposed to be ignored? `num_send_streams` is important. Is that correct @seanmonstar?

src/proto/streams/counts.rs
```diff
@@ -46,6 +46,10 @@ impl Counts {
         self.peer
     }
 
+    pub fn has_streams(&self) -> bool {
+        self.num_recv_streams != 0 || self.num_reset_streams != 0
```
carllerche: I think it was num_reset_streams that was supposed to be ignored? num_send_streams is important. Is that correct @seanmonstar?

seanmonstar: Yes, it is `num_reset_streams` that isn't needed.

DarrenTsung: Oh crap, my bad. Updated.

carllerche:
CI passed 🎊

Issue #224, opened 2018-02-15
self.pending_capacity.is_empty() not always guaranteed


DarrenTsung:
https://github.com/carllerche/h2/blob/master/src/proto/streams/prioritize.rs#L357

I'm getting panics at this assert when trying to load test by sending 4k notifications as fast as possible. Looking at the code, it looks safe to have this assert as `try_assign_capacity` can push 0 or 1 streams into `self.pending_capacity` and can be called from `assign_connection_capacity` which must pop 1 stream from `self.pending_capacity` before calling `try_assign_capacity`.

However, `try_assign_capacity` can also be called from `send_data` (https://github.com/carllerche/h2/blob/master/src/proto/streams/prioritize.rs#L148). 

When opening a bunch of simultaneous streams, I seeing `send_data` getting called multiple times (unable to assign capacity, the streams are put in `self.pending_capacity`). Then a call to `assign_connection_capacity` when there is extra capacity, then the panic because there are still streams waiting in `self.pending_capacity`. 

Do you guys know what might be happening? Maybe there's a bug externally to the assert and the assert should still hold true, or maybe this assert doesn't hold true with code pushed after the fact? 

Thanks!

seanmonstar:
So, a bunch of streams are waiting for capacity, and then when a `WINDOW_UPDATE` is received for the connection, a pending stream is popped from the queue, and `try_assign_capacity` is called with, while there are still other connections pending?

That seems like the assertion is incorrect then. @carllerche?

carllerche:
Yeah, that assertion does not seem correct. I'm not sure what I was thinking originally. That debug_assert was included in the [original PR](https://github.com/carllerche/h2/pull/25) so I was probably still trying to figure out the guarantees and adding as many assertions as possible to see what would fail!

It would be fine to remove I think.

carllerche:
It would also be nice to include a test that demonstrates the panic as it this also is indicative of test coverage lacking. 👍

DarrenTsung:
Oh, just read your message, I'll write up a test for that.

PR #225, opened 2018-02-15
Remove assert around self.pending_capacity.is_empty() (#224)

DarrenTsung:
This assert does not hold as many streams can be pushed into
pending_capacity during a call to send_data(). See issue #224
for more discussion and sign-off.

### DIFF
--- a/src/proto/streams/prioritize.rs
+++ b/src/proto/streams/prioritize.rs
@@ -353,9 +353,6 @@ impl Prioritize {
 
         // First check if capacity is immediately available
         if conn_available > 0 {
-            // There should be no streams pending capacity
-            debug_assert!(self.pending_capacity.is_empty());
-
             // The amount of capacity to assign to the stream
             // TODO: Should prioritization factor into this?
             let assign = cmp::min(conn_available, additional);
--- a/src/proto/streams/prioritize.rs
+++ b/src/proto/streams/prioritize.rs
@@ -353,6 +353,9 @@ impl Prioritize {
 
         // First check if capacity is immediately available
         if conn_available > 0 {
+            // There should be no streams pending capacity
+            debug_assert!(self.pending_capacity.is_empty());
+
             // The amount of capacity to assign to the stream
             // TODO: Should prioritization factor into this?
             let assign = cmp::min(conn_available, additional);
--- a/src/proto/streams/store.rs
+++ b/src/proto/streams/store.rs
@@ -215,10 +215,6 @@ where
         }
     }
 
-    pub fn is_empty(&self) -> bool {
-        self.indices.is_none()
-    }
-
     pub fn take(&mut self) -> Self {
         Queue {
             indices: self.indices.take(),
--- a/src/proto/streams/store.rs
+++ b/src/proto/streams/store.rs
@@ -215,6 +215,10 @@ where
         }
     }
 
+    pub fn is_empty(&self) -> bool {
+        self.indices.is_none()
+    }
+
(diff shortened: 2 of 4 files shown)


DarrenTsung:
Oops, got to remove some unused methods.

DarrenTsung:
Ok, seems good to me. Tests pass with `cargo test --features unstable`.

DarrenTsung:
Currently looking at writing a test for demonstrating the panic..

DarrenTsung:
Ok, added a failing test. Basically given some set of queued requests that are larger than the default window size, send_data should not panic when assigning more capacity to a stream.

Pretty much all of the new data that the mock receives I have no idea how to interpret, maybe you guys could help me out here so the test doesn't look like magic? The base of the test is taken from the test `single_stream_send_body_greater_than_default_window()` which is a couple lines below.

https://github.com/DarrenTsung/h2/commit/4fd36052a03a769e31f5e533bc5adecd72f1c1e6#diff-726773c7c7a823eac63a932c9c0f117bR83

DarrenTsung:
Can anyone take a look the test added in this commit: https://github.com/carllerche/h2/pull/225/commits/4fd36052a03a769e31f5e533bc5adecd72f1c1e6?

I think it demonstrates the failing case correctly, but I'm not sure why some of the data is expected / how to document the test correctly.

carllerche:
It looks like the test is hitting hte assertion. I'm not sure what you mean by:

> but I'm not sure why some of the data is expected / how to document the test correctly.

Also, is there a reason you wrote the raw frame bytes instead of using the [helpers](https://github.com/carllerche/h2/blob/master/tests/ping_pong.rs#L46-L53)?

DarrenTsung:
Sorry for the delay, yes, the test is hitting the assertion which is what I wanted to make sure was covered.

Basically, nearly all of the expected data for the test is taken from another test in the file `single_stream_send_body_greater_than_default_window()` which is a [couple lines below](https://github.com/carllerche/h2/blob/4fd36052a03a769e31f5e533bc5adecd72f1c1e6/tests/prioritization.rs#L251). I added [a couple sections in the expected data with raw frame bytes](https://github.com/carllerche/h2/blob/4fd36052a03a769e31f5e533bc5adecd72f1c1e6/tests/prioritization.rs#L83) in response to the unexpected (new) data from the new `send_request` calls.

Everything works and the test fails as expected, which is great. The only thing is I'm not sure exactly what the new data receiving means. Some of it looks like an ACK for the `send_request` calls I added (my guess), but some of it does not look like that at all. I would love to use the helpers for the data to make it more read-able / comprehensible, just not sure which helpers to use.

Any guidance on interpreting the new data, either through comments or using a helper instead, would be very appreciated!

DarrenTsung:
Can anyone take a look at the test above? If it seems fine I can remove the assert, have the test pass, and the PR should be ready for merge (unless any changes are requested).

seanmonstar:
I admit that looking at the test and seeing raw bytes makes it a lot harder to review and then also be able to debug in the future if we every accidentally regress. For instance, I'm not actually sure what part is unexpected, nor what the `// (???)` mean :D

What do think of tests that are written in this style instead? https://github.com/carllerche/h2/blob/master/tests/flow_control.rs#L1162-L1204

DarrenTsung:
Thanks for pointing out that example @seanmonstar! I was able to get rid of the "hand-wave-y" raw data frames and pushed a commit that fixed the test by removing the debug assertion again.

PR #228, opened 2018-02-28
Normalize HTTP request path.

carllerche:
The HTTP/2.0 specification requires that the path pseudo header is never
empty for requests unless the request uses the OPTIONS method.

This is currently not correctly enforced.

This patch provides a test and a fix.

### DIFF
--- a/src/frame/headers.rs
+++ b/src/frame/headers.rs
@@ -440,10 +440,14 @@ impl Pseudo {
     pub fn request(method: Method, uri: Uri) -> Self {
         let parts = uri::Parts::from(uri);
 
-        let path = parts
+        let mut path = parts
             .path_and_query
             .map(|v| v.into())
-            .unwrap_or_else(|| Bytes::from_static(b"/"));
+            .unwrap_or_else(|| Bytes::new());
+
+        if path.is_empty() && method != Method::OPTIONS {
+            path = Bytes::from_static(b"/");
+        }
 
         let mut pseudo = Pseudo {
             method: Some(method),
--- a/src/frame/headers.rs
+++ b/src/frame/headers.rs
@@ -273,7 +273,6 @@ impl fmt::Debug for Headers {
             .field("stream_id", &self.stream_id)
             .field("stream_dep", &self.stream_dep)
             .field("flags", &self.flags)
-            .field("header_block", &self.header_block)
             // `fields` and `pseudo` purposefully not included
             .finish()
     }
--- a/src/frame/headers.rs
+++ b/src/frame/headers.rs
@@ -273,6 +273,7 @@ impl fmt::Debug for Headers {
             .field("stream_id", &self.stream_id)
             .field("stream_dep", &self.stream_dep)
             .field("flags", &self.flags)
+            .field("header_block", &self.header_block)
             // `fields` and `pseudo` purposefully not included
             .finish()
     }
--- a/tests/client_request.rs
+++ b/tests/client_request.rs
@@ -658,6 +658,44 @@ fn request_without_path() {
     client.join(srv).wait().unwrap();
 }
 
+#[test]
+fn request_options_with_star() {
+    let _ = ::env_logger::try_init();
+    let (io, srv) = mock::new();
+
+    let uri = uri::Uri::from_parts({
+        let mut parts = uri::Parts::default();
+        parts.scheme = Some(uri::Scheme::HTTP);
+        parts.authority = Some(uri::Authority::from_shared("example.com".into()).unwrap());
+        parts.path_and_query = Some(uri::PathAndQuery::from_static("*"));
+        parts
+    }).unwrap();
+
+    let srv = srv.assert_client_handshake()
+        .unwrap()
+        .recv_settings()
+        .recv_frame(frames::headers(1).request("OPTIONS", uri.clone()).eos())
+        .send_frame(frames::headers(1).response(200).eos())
+        .close();
+
+    let client = client::handshake(io)
+        .expect("handshake")
+        .and_then(move |(mut client, conn)| {
+            // Note the lack of trailing slash.
+            let request = Request::builder()
+                .method(Method::OPTIONS)
+                .uri(uri)
+                .body(())
+                .unwrap();
+
+            let (response, _) = client.send_request(request, true).unwrap();
+
+            conn.drive(response)
+        });
+
+    client.join(srv).wait().unwrap();
+}
+
 const SETTINGS: &'static [u8] = &[0, 0, 0, 4, 0, 0, 0, 0, 0];
 const SETTINGS_ACK: &'static [u8] = &[0, 0, 0, 4, 1, 0, 0, 0, 0];
 
--- a/tests/client_request.rs
+++ b/tests/client_request.rs
@@ -663,6 +663,7 @@ fn request_options_with_star() {
     let _ = ::env_logger::try_init();
     let (io, srv) = mock::new();
 
+    // Note the lack of trailing slash.
     let uri = uri::Uri::from_parts({
         let mut parts = uri::Parts::default();
         parts.scheme = Some(uri::Scheme::HTTP);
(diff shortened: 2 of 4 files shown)


src/frame/headers.rs
```diff
@@ -440,10 +440,14 @@ impl Pseudo {
...
         let parts = uri::Parts::from(uri);
 
-        let path = parts
+        let mut path = parts
             .path_and_query
             .map(|v| v.into())
-            .unwrap_or_else(|| Bytes::from_static(b"/"));
+            .unwrap_or_else(|| Bytes::new());
+
+        if path.is_empty() && method != Method::OPTIONS {
```
briansmith: When the URI is "*" it isn't clear to me what `path` is supposed to be here. is it "*" like `Uri::path()` says or is it ""?

When URI is "*" and the method isn't OPTIONS then it seems like we should return an error somewhere here, but that doesn't seem to be done. Or maybe I'm misunderstanding some details of how "*" is handled in the parsing code?

tests/client_request.rs
```diff
@@ -630,6 +630,34 @@ fn recv_too_big_headers() {
...
+        .recv_settings()
+        .recv_frame(frames::headers(1).request("GET", "http://example.com/").eos())
+        .send_frame(frames::headers(1).response(200).eos())
+        .close();
+
+    let client = client::handshake(io)
+        .expect("handshake")
+        .and_then(move |(mut client, conn)| {
+            // Note the lack of trailing slash.
+            let request = Request::get("http://example.com")
```
briansmith: It would be good to see tests for "OPTIONS *" and "GET *" as well as "OPTIONS /*" and "GET /*"

src/frame/headers.rs
carllerche: The relevant [spec section](http://httpwg.org/specs/rfc7540.html#rfc.section.8.1.2.3):

> The :path pseudo-header field includes the path and query parts of the target URI (the path-absolute production and optionally a '?' character followed by the query production (see Sections 3.3 and 3.4 of [RFC3986]). A request in asterisk form includes the value '*' for the :path pseudo-header field.
>
> This pseudo-header field MUST NOT be empty for http or https URIs; http or https URIs that do not contain a path component MUST include a value of '/'. The exception to this rule is an OPTIONS request for an http or https URI that does not include a path component; these MUST include a :path pseudo-header field with a value of '*' (see [RFC7230], Section 5.3.4).

Does this make sense?

I also saw you opened https://github.com/hyperium/http/issues/176, so we can discuss the details of the `http` path API there.

carllerche: So, the path can also be not set of the scheme is not `http` or `https`, but those cases are not really supported yet.

tests/client_request.rs
carllerche: I tried to write a test for OPTIONS, but there seems to be some problems unrelated to this change. So, I will open a new issue to track this.

carllerche: Ok, I figured out how to write a test, but #229 still applies as it is kind of janky.

tests/prioritization.rs
```diff
@@ -68,7 +68,7 @@ fn single_stream_send_large_body() {
 
 #[test]
 fn multiple_streams_with_payload_greater_than_default_window() {
-    let _ = ::env_logger::init();
+    let _ = ::env_logger::try_init();
```
carllerche: This is unrelated to the PR but it can cause tests to fail and it didn't seem worth splitting it out.

briansmith:
I reviewed the initial PR without thinking too much. But, I actually think that the proposed fix is not a good design. We shouldn't default every request path to "/" when the request path isn't provided by the caller. Instead, we should refuse to issue the request. If the caller forgot to provide the request path then that's a logic failure on the caller's part. Defaulting to "/" will result in applications accidentally sending requests to the wrong path, which is dangerous.

In the case of OPTIONS, it is especially confusing: should the default be "*", or should it be "/" like everything else? It seems wrong for it to be inconsistent with the others, but "/" also seems wrong. It's better to require the user to be explicit.

src/frame/headers.rs
```diff
@@ -273,6 +273,7 @@ impl fmt::Debug for Headers {
             .field("stream_id", &self.stream_id)
             .field("stream_dep", &self.stream_dep)
             .field("flags", &self.flags)
+            .field("header_block", &self.header_block)
```
seanmonstar: The comment right after this (while slight outdated) suggests that this field is purposefully not included :)

seanmonstar:
> Defaulting to "/" will result in applications accidentally sending requests to the wrong path, which is dangerous.

Is there an instance where that is true? I cannot think of any.

------

I actually think that perhaps the behavior should be fixed in `http`. If the `Uri` *should* have a path (either `absolute-form` or `origin-form` (aka `?foo=bar` or `http://foo.bar`), then `PathAndQuery` should match the already existing `path()` behavior of including a `/`.

briansmith:
> > Defaulting to "/" will result in applications accidentally sending requests to the wrong path, which is dangerous.
> 
> Is there an instance where that is true? I cannot think of any.

I'm sure there are many cases. One that comes to mind: Some domain ownership verification procedures require one to add a token somewhere in the file at `/.well-known/something-something`. If the request to verify the token is present at `/.well-known/something-something` is accidentally sent to `/` then any user could potentially do the domain verification procedure if they can modify the home page (e.g. by submitting a popular story to reddit, one can get on its home page with arbitrary text).

> PathAndQuery should match the already existing path() behavior of including a /.

That's basically https://github.com/hyperium/http/issues/176.

seanmonstar:
>  One that comes to mind: Some domain ownership verification procedures require one to add a token somewhere in the file at `/.well-known/something-something`. If the request to verify the token is present at `/.well-known/something-something` is accidentally sent to `/` then any user could potentially do the domain verification procedure if they can modify the home page.

I don't see how this bug occurs from assuming that the path of `http://example.com` is `/`.

carllerche:
Also, assuming that `http://example.com` implies a path of `/` is what browsers do.

I would say that constructing a URI from parts should possibly not make that assumption though.

tests/client_request.rs
```diff
@@ -630,6 +630,72 @@ fn recv_too_big_headers() {
...
+        .unwrap()
+        .recv_settings()
+        .recv_frame(frames::headers(1).request("OPTIONS", uri.clone()).eos())
+        .send_frame(frames::headers(1).response(200).eos())
+        .close();
+
+    let client = client::handshake(io)
+        .expect("handshake")
+        .and_then(move |(mut client, conn)| {
+            // Note the lack of trailing slash.
```
seanmonstar: I'd move this comment up to the `let uri = ...` line, since here it just left me confused ("Note? But I don't see any lack of trailing slash here!")

tests/client_request.rs
```diff
@@ -630,6 +630,72 @@ fn recv_too_big_headers() {
...
+#[test]
+fn request_options_with_star() {
+    let _ = ::env_logger::try_init();
+    let (io, srv) = mock::new();
+
+    let uri = uri::Uri::from_parts({
+        let mut parts = uri::Parts::default();
+        parts.scheme = Some(uri::Scheme::HTTP);
+        parts.authority = Some(uri::Authority::from_shared("example.com".into()).unwrap());
+        parts.path_and_query = Some(uri::PathAndQuery::from_static("*"));
```
seanmonstar: Should we add a `PathAndQuery::STAR` const in the http crate?

src/frame/headers.rs
seanmonstar: I think this is fine, but it should also check for `Method::CONNECT.

[RFC7450#CONNECT](http://httpwg.org/specs/rfc7540.html#CONNECT):

> The :scheme and :path pseudo-header fields MUST be omitted.

tests/client_request.rs
carllerche: It might be useful!

carllerche:
@seanmonstar generally, I don't think `CONNECT` is fully supported / implemented yet. I'd recommend punting on any connect work in this PR.

carllerche:
Re the `/` default path discussion, I asked the question [here](https://github.com/hyperium/http/issues/176).

It seems like this is a general question and I am unsure if it merits holding up this PR.

PR #230, opened 2018-03-05
Streams that are RESET by peer don't decrement count towards concurrency limit

DarrenTsung:
Sending 4000 notifications to APNs (Apple Push Notification service) gives this result:
1. Starts 4000 streams because hasn't received MAX_CONCURRENT_STREAMS from settings yet.
2. Receive settings with MAX_CONCURRENT_STREAMS as 1000.
3. Streams past the first 1000 are sent a reset frame by APNs.
4. ... streams are retried with correct limit + eventually everything succeeds.
5. Connection doesn't close automatically (#222) because `num_send_streams` is 3000.

`num_send_streams` is 3000 at the end of this example because streams that are reset do not return true for `is_closed()`. I added a check for this in `is_closed()`.

### DIFF
--- a/src/proto/streams/stream.rs
+++ b/src/proto/streams/stream.rs
@@ -212,6 +212,12 @@ impl Stream {
 
     /// Returns true if the stream is closed
     pub fn is_closed(&self) -> bool {
+        // If the stream has been reset by the peer then it will
+        // still have frames to send, but should count as closed
+        if self.state.is_non_local_reset() {
+            return true;
+        }
+
         // The state has fully transitioned to closed.
         self.state.is_closed() &&
             // Because outbound frames transition the stream state before being
--- a/src/proto/streams/state.rs
+++ b/src/proto/streams/state.rs
@@ -301,6 +301,13 @@ impl State {
         }
     }
 
+    pub fn is_non_local_reset(&self) -> bool {
+        match self.inner {
+            Closed(Cause::Proto(_)) => true,
+            _ => false,
+        }
+    }
+
     /// Returns true if the stream is already reset.
     pub fn is_reset(&self) -> bool {
         match self.inner {
--- a/src/proto/streams/store.rs
+++ b/src/proto/streams/store.rs
@@ -85,6 +85,10 @@ impl Store {
         }
     }
 
+    pub fn contains_key(&mut self, key: &Key) -> bool {
+        self.slab.contains(key.index)
+    }
+
     pub fn find_mut(&mut self, id: &StreamId) -> Option<Ptr> {
         let key = match self.ids.get(id) {
             Some(key) => *key,
--- a/src/proto/streams/store.rs
+++ b/src/proto/streams/store.rs
@@ -85,10 +85,6 @@ impl Store {
         }
     }
 
-    pub fn contains_key(&mut self, key: &Key) -> bool {
-        self.slab.contains(key.index)
-    }
-
     pub fn find_mut(&mut self, id: &StreamId) -> Option<Ptr> {
         let key = match self.ids.get(id) {
             Some(key) => *key,
--- a/src/proto/streams/streams.rs
+++ b/src/proto/streams/streams.rs
@@ -641,13 +641,6 @@ where
         me.actions.send.ensure_next_stream_id()?;
 
         if let Some(key) = key {
-            // Because the key does not hold a reference to this stream
-            // it is possible for the stream not to exist here (stream closed)
-            // therefore, we can return Ok(Async::Ready)
-            if !me.store.contains_key(key) {
-                return Ok(().into());
-            }
-
             let mut stream = me.store.resolve(*key);
             trace!("poll_pending_open; stream = {:?}", stream.is_pending_open);
             if stream.is_pending_open {
--- a/src/proto/streams/streams.rs
+++ b/src/proto/streams/streams.rs
@@ -641,6 +641,13 @@ where
         me.actions.send.ensure_next_stream_id()?;
 
         if let Some(key) = key {
+            // Because the key does not hold a reference to this stream
+            // it is possible for the stream not to exist here (stream closed)
+            // therefore, we can return Ok(Async::Ready)
+            if !me.store.contains_key(key) {
+                return Ok(().into());
+            }
+
             let mut stream = me.store.resolve(*key);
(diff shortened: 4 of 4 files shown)


src/proto/streams/stream.rs
```diff
@@ -212,6 +212,12 @@ impl Stream {
 
     /// Returns true if the stream is closed
     pub fn is_closed(&self) -> bool {
+        // If the stream has been reset by the peer then it will
+        // still have frames to send, but should count as closed
+        if self.state.is_non_local_reset() {
```
seanmonstar: Hm, I kind of wonder if this is bandaiding over another issue, which is that since the remote has reset this stream, then we should have cleared `pending_send` and `buffered_send_data`, right?

jwilm:
I would argue that the h2 API should not support creating streams until after a `SETTINGS` frame is received from the peer. Here's the relevant part of the spec:

> A SETTINGS frame MUST be sent by both endpoints at the start of a connection

seanmonstar:
Also relevant:

> To avoid unnecessary latency, clients are permitted to send additional frames to the server immediately after sending the client connection preface, without waiting to receive the server connection preface.

src/proto/streams/stream.rs
DarrenTsung: Yeah, that makes sense. Just wasn't sure if that was the right thing to do

carllerche:
I've been wanting to add a local limit as well. So the effective limit would be `min(local-limit, remote-limit)`.

Perhaps this could help? In that case, you could configure a max # of in-flight streams equal to what you expect the remote to be?

Of course, you would want to allow more if the remote allows more too. So, another option would be to supply a max that is in effect until you receive the remote settings. If this is set to zero, that would effectively achieve the behavior suggested by @jwilm

src/proto/streams/stream.rs
DarrenTsung: @seanmonstar - I pushed the change for `clearing pending_send` + `buffered_send_data` to a branch: https://github.com/DarrenTsung/h2/commit/6ee879630febb1b0a186054fbdc169e5db251e99 because I'm having some issues with `pop_frame()` getting called with streams that are reset and therefore don't have any frames to pop. The only exception is apparently scheduled reset streams, hitting this `expect("must be scheduled to reset")` (https://github.com/DarrenTsung/h2/blob/clear-send-buffer-on-remote-reset/src/proto/streams/prioritize.rs#L659).

Do you know why the connection is expecting the stream closed by remote reset to pop_frame?

jwilm:
@carllerche the "initial max" variant you suggested might be nice for our use case. The local limit makes sense as well, but doesn't do as much for us. The problem with setting this to the expected value is that the peer could change it at any time, and if they permit more, then we don't necessarily see any improvement without a code change.

All that said, it seems like there's a bug when exceeding the peer's limit at startup which needs to be resolved in any case.

src/proto/streams/stream.rs
seanmonstar: The `pending_send` buffer is always used in conjunction with the `Prioritize.pending_send` queue. So, if there are items in `pending_send`, it is also true the this stream is in `pending_send` queue. Then, in `poll_complete`, when the stream is taken from the queue, it's a logic error that the stream was queued but didn't have anything to write.

DarrenTsung: Ah okay. If this stream which is already in the `pending_send` buffer & queue receives a reset, then we should clear buffer + remove from queue.

Weird thing is that the `clear_queue()` method called by `recv_err()` only clears the buffer + doesn't remove it from the queue. How did streams that receive an error not hit this case where the buffer was cleared, but they were still queued? (Maybe they weren't queued + didn't have anything in the buffer?).

Anyways, I can see two solutions to this: 
1. Remove from queue in the `clear_queue()` (which already clears the buffer). This would require some sort of `remove()` method implemented for `store::Queue`. We would need to traverse the entire queue to find the matching key for the stream, which is not ideal.
2. Check if the stream `is_reset()` when popping a stream from the queue. If it is, we can just `continue;` and move on to the next stream.

Any preferences? I'll move on with implementing 2 first and seeing if it resolves the problem, should be a three-line change.

carllerche: I would say option 2 is best. This is a rare(ish) case, so it does not seem worth to require a linear traversal or change the queue to a doubly linked list.

DarrenTsung:
@carllerche @seanmonstar, what do you think is the right solution for this issue?

Basically, a stream increments `num_send_streams` when being pushed into `self.pending_send`.
https://github.com/carllerche/h2/blob/master/src/proto/streams/prioritize.rs#L700

And when the stream receives a reset, it transitions to closed:
https://github.com/carllerche/h2/blob/master/src/proto/streams/streams.rs#L260

But the stream does not decrement `num_send_streams` because it is not `is_counted()` before transitioning to closed. The reason `is_counted()` returns false is because it is still `self.is_pending_open` which is set when it is pushed into the queue (`self.pending_send`).
https://github.com/carllerche/h2/blob/master/src/proto/streams/stream.rs#L204

So current solution for removing the stream from the queue (when popped, ignore if reset) does not remove this flag pre-transition.

It seems in order to decrement `num_send_streams` properly we would need to remove it from the queue and clear the buffer before transitioning to mark it as `is_counted()`. But it seems a bit weird to change its state and then call `counts.transition`, as the whole point is that the state change happens within the function passed in.

carllerche:
@DarrenTsung Thanks for the analysis.

I'm not following this:

> The reason `is_counted()` returns false is because it is still `self.is_pending_open` which is set when it is pushed into the queue (`self.pending_send`). https://github.com/carllerche/h2/blob/master/src/proto/streams/stream.rs#L204

I don't see where `is_pending_open` is being set when it is pushed into the pending send queue.

The `self.pending_open.pop(store)` call should set `is_pending_open` to false. So, that should also transition the stream to "is_counted".

Also, if you want to talk via IRC/Gitter, to try to work through this faster, that is also an option.

DarrenTsung:
Sorry, what I wrote above was incorrect - pending_send should not be linked with is_pending_open. I actually found that this was the symptom of another issue (streams were being pushed into both pending_send and pending_open).

I'm closing this PR in order to order to open another PR as the work done on this branch is around debugging the root of the issue.

PR #231, opened 2018-03-07
notify send_tasks when there is a connection error

seanmonstar:
If `client::SendRequest` had parked a future with `poll_ready`, and there is a connection error (not eof), that future was never notified.

Also removes `open_task`, as it was dead code.


pimeys:
I've been testing this with real data the whole day, I see no more connection errors and requests getting stuck into limbo.

PR #235, opened 2018-03-07
Prevent pushing a stream into both pending_send + pending_open

DarrenTsung:
Because external users of h2 are allowed to `send_request()` and `send_data()` without blocking, a stream can get pushed into the pending_open queue (waiting for number of streams to go down), and also the pending_send queue (waiting for capacity on the stream / connection). My proposed solution is to prevent streams from being pushed into pending_send queue as streams put in the queue are scheduled to be executed + data should not be sent for a stream that is not open.

There are also newly added asserts that a stream should never be in pending_send and pending_open at the same time.

### DIFF
--- a/src/proto/streams/prioritize.rs
+++ b/src/proto/streams/prioritize.rs
@@ -407,8 +407,9 @@ impl Prioritize {
         // if there is no room according to the concurrency limit (max_send_streams), and we
         // also allow data to be buffered for send with send_data() if there is no capacity for
         // the stream to send the data, which attempts to place the stream in pending_send.
-        // If the stream is not open, we don't want the stream to be
-        // scheduled for execution (pending_send)
+        // If the stream is not open, we don't want the stream to be scheduled for
+        // execution (pending_send). Note that if the stream is in pending_open, it will be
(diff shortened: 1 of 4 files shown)


src/proto/streams/prioritize.rs
```diff
@@ -398,8 +398,18 @@ impl Prioritize {
...
+        // schedule the stream for execution
+        //
+        // Why do we not push into pending_send when the stream is in pending_open?
+        //
+        // We allow users to call send_request() which schedules a stream to be pending_open
+        // if there is no room according to the concurrency limit (max_send_streams), and we
+        // also allow data to be buffered for send with send_data() if there is no capacity for
+        // the stream to send the data, which attempts to place the stream in pending_send.
+        // If the stream is not open, we don't want the stream to be
+        // scheduled for execution (pending_send)
```
carllerche: Really great comment!

Maybe also include that once the stream gets removed from pending_open it will be queued on `pending_send` (so the writes get scheduled at that point).

DarrenTsung: Good point, added!

PR #238, opened 2018-03-12
Streams receiving peer reset clear pending send

DarrenTsung:
So, I thought this issue was solved with #235, but actually what was happening was I had both these fixes + #235 previously, and removed these fixes because they didn't seem to be necessary (test code worked fine). Unfortunately, turns out there a race condition external to the h2 crate that was hiding this issue (adding additional logging causes this issue again).

The issue:
Streams that have pending send data and also receive a peer reset are not counted as `is_closed()` after reset. Because these streams were opened, they increment `num_send_streams` which is an internal tracker of the number of open send streams checked against `max_send_streams`. Therefore the connection will believe streams are still in-flight + won't open new streams.

The fix:
Properly call `prioritize.clear_queue()` when receiving a reset. I added two lines which reset `buffered_send_data` and `requested_send_capacity` to 0 which mimics logic from `pop_frame` (when frames are actually sent, they decrement these as well). This means that the stream is marked as `is_closed()` and decrements `num_send_streams` immediately after receiving a RESET.

Additional things: 
* I removed the Result<()> from `state.recv_reset` as it isn't needed and implies that the stream could fail changing the state + not clear the queue.
* We remove the peer reset stream from the `pending_send` queue when popped because otherwise it would be an O(N) traversal to remove it from the queue when receiving a reset.

Alternatives:
Instead of using `is_peer_reset()` we could alternatively have a variable on stream that is set when the queue is reset. That could be used instead when popping streams off `pending_send` and ignoring them.

### DIFF
--- a/src/proto/streams/prioritize.rs
+++ b/src/proto/streams/prioritize.rs
@@ -555,6 +555,9 @@ impl Prioritize {
         while let Some(frame) = stream.pending_send.pop_front(buffer) {
             trace!("dropping; frame={:?}", frame);
         }
+
+        stream.buffered_send_data = 0;
+        stream.requested_send_capacity = 0;
     }
 
     fn pop_frame<B>(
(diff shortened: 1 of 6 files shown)


carllerche:
This looks good to me. Is there possibly a way to test this?

I'm also not sure why CI is failing. The error doesn't immediately seem related to this PR. I restarted the build to see.

DarrenTsung:
Yes - added a test! Reading all the h2 code really helped, it was a lot easier to write a test this time, haha. Not sure how to prove this, but on master the test will hang because the connection does not send a GO_AWAY frame (the stream is still being held through `num_send_streams`).

PR #242, opened 2018-03-15
Add initial_max_send_streams() as builder option

DarrenTsung:
Problem: Establishing a connection to some provider and using the `poll_ready()` function to determine when there is space to establish new streams means that we can send up to `usize::MAX` streams until receiving a `Settings` frame from the remote peer.

Solution: Allow callers to specify an `initial_max_send_streams` which can be used until receiving the `Settings` frame. This will prevent using extraneous bandwidth and an extraneous amount of REFUSED_STREAM resets and can scale up or down the number of streams without causing unnecessary delays.

The default behavior is to keep the `initial_max_send_streams` at `usize::MAX` to keep functionality the same.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -177,6 +177,7 @@ use tokio_io::io::WriteAll;
 use std::fmt;
 use std::marker::PhantomData;
 use std::time::Duration;
+use std::usize;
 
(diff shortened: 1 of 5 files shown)


carllerche:
LGTM!

PR #244, opened 2018-03-21
Accept HPACK literals greater than max size

### LINKED ISSUE #243: Handle maximum table size of zero in HPACK decoding
According to https://http2.github.io/http2-spec/compression.html#rfc.section.4.2, it's valid to set the HPACK dynamic table size to zero. It is also generally valid to attempt to write more entries into the header table which don't fit in max_size:

> It is not an error to attempt to add an entry that is larger than the maximum size; an attempt to add an entry larger than the maximum size causes the table to be emptied of all existing entries and results in an empty table. 

We currently panic in this case here: https://github.com/carllerche/h2/blob/master/src/hpack/decoder.rs#L503 and should instead return an empty header table.

carllerche:
The spec specifically allows accepting HPACK literals with indexing when
the HPACK literal is greater than the max table size. In this case, the
literal is not inserted in the table.

Fixes #243

### DIFF
--- a/src/hpack/decoder.rs
+++ b/src/hpack/decoder.rs
@@ -482,10 +482,12 @@ impl Table {
 
         self.reserve(len);
 
-        self.size += len;
+        if self.size + len <= self.max_size {
+            self.size += len;
 
-        // Track the entry
-        self.entries.push_front(entry);
+            // Track the entry
+            self.entries.push_front(entry);
+        }
     }
 
     fn set_max_size(&mut self, size: usize) {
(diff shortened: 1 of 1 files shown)


rbtying:
Wanted to confirm that this does in fact fix #243 !

src/hpack/decoder.rs
```diff
@@ -495,14 +497,13 @@ impl Table {
...
-            let last = self.entries
-                .pop_back()
-                .expect("size of table != 0, but no headers left!");
-
-            self.size -= last.len();
+            match self.entries.pop_back() {
+                Some(last) => {
+                    self.size -= last.len();
+                }
+                None => return,
```
olix0r: this return is surprising because there is nothing else in the function? not fundamentally opposed to this, just noting

carllerche: It's for breaking out of the loop.

hawkw: Why not `break`?

carllerche: IMO it makes it more obvious that, at that point, the fn is done vs. having to find the end of the loop and see what happens there.

Just a general opinion. The fn is small.

hawkw: Yeah, that makes sense, just curious!

PR #247, opened 2018-03-27
recv_reset resets closed streams with queued EOS frames

### LINKED ISSUE #246: Panic when receiving a reset on a stream in the process of closing
This is the cause of upstream issue runconduit/conduit#607. 

It appears that when a stream is in the process of transitioning from Half-Closed (remote) to Closed, with a Trailers frame queued to end that stream, and a Reset (Canceled) frame is received, `h2` will hit an `expect` in `prioritize.rs` and panic. 

---

Here's the `expect` that fails:
https://github.com/carllerche/h2/blob/f8baeb7211985834acaf7ecc70548aafd5d9ef6d/src/proto/streams/prioritize.rs#L693-L694

And the relevant logs from @bourquep's panicking Conduit instance:
```rust
E  TRCE h2::proto::streams::state send_close: HalfClosedRemote => Closed
E  TRCE h2::proto::streams::send send_trailers -- queuing; frame=Headers { stream_id: StreamId(107), stream_dep: None, flags: HeadersFlag { end_stream: true, end_headers: true, padded: false, priority: false } }
E  TRCE h2::proto::streams::prioritize schedule_send; StreamId(107)
E  TRCE h2::proto::streams::store Queue::push
E  TRCE h2::proto::streams::store  -> first entry
E  TRCE h2::proto::streams::prioritize reserve_capacity; stream=StreamId(107); requested=0; effective=0; curr=0
E  TRCE h2::proto::settings send_pending_ack; pending=None
E  TRCE h2::codec::framed_read poll
E  TRCE h2::codec::framed_read poll; bytes=13B
E  TRCE h2::codec::framed_read decoding frame from 13B
E  TRCE h2::codec::framed_read     -> kind=Reset
E  DBUG h2::codec::framed_read received; frame=Frame::Reset(Reset { stream_id: StreamId(107), error_code: CANCEL })
E  TRCE h2::proto::connection recv RST_STREAM; frame=Reset { stream_id: StreamId(107), error_code: CANCEL }
E  TRCE h2::proto::streams::prioritize clear_queue; stream-id=StreamId(107)
E  TRCE h2::proto::streams::prioritize dropping; frame=Frame::Headers(Headers { stream_id: StreamId(107), stream_dep: None, flags: HeadersFlag { end_stream: true, end_headers: true, padded: false, priority: false } })
E  TRCE h2::proto::settings send_pending_ack; pending=None
E  TRCE h2::codec::framed_read poll
E  TRCE h2::proto::streams::prioritize try reclaim frame
E  TRCE h2::proto::streams::prioritize poll_complete
E  TRCE h2::proto::streams::prioritize schedule_pending_open
E  TRCE h2::proto::streams::prioritize pop_frame
E  TRCE h2::proto::streams::prioritize pop_frame; stream=StreamId(107)
E  thread 'main' panicked at 'must be scheduled to reset', /checkout/src/libcore/option.rs:891:5
E  note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

More information on the specific case where this issue occurs may be found in runconduit/conduit#607.

hawkw:
Fixes #246. Fixes runconduit/conduit#607.

An issue currently exists where `h2` will panic when a stream transitions from `HalfClosed(remote)` to `Closed(Cause::EndStream)` by enqueuing an EOS frame, and then receives a `RST_STREAM` frame _before_ the queued EOS frame is sent.

In this case, the `expect` in `prioritize::pop_frame` https://github.com/carllerche/h2/blob/f8baeb7211985834acaf7ecc70548aafd5d9ef6d/src/proto/streams/prioritize.rs#L693-L694 is called on a frame which does _not_ have a scheduled reset, because it is in the `Closed(Cause::EndStream)` state. However, the guard which skips this code when the stream is peer reset https://github.com/carllerche/h2/blob/f8baeb7211985834acaf7ecc70548aafd5d9ef6d/src/proto/streams/prioritize.rs#L584-L586 doesn't cause us to skip in this case, because the stream is in the `Closed(Cause::EndStream)` state.

---

I've modified [`streams::State::recv_reset`](
https://github.com/hawkw/h2/blob/14200321d77aaf31e57fe7c40d4223030e113922/src/proto/streams/state.rs#L217-L232) to take a boolean parameter indicating whether the stream receiving a reset currently has a frame queued to send. If the stream is in the `Closed(Cause::EndStream)` state, rather than doing nothing, we check whether there is a queued frame. If so, the stream transitions to `Closed(Cause::Proto)` with the reason for the received reset set as the reason. 

This way, when `prioritize::pop_frame` sees this stream, it will be treated by any other reset stream, draining the queue and transitioning to reset. We no longer hit the `expect` call that was causing panics in this case.

I've also added a unit test that reproduces the crash against master.

### DIFF
--- a/src/proto/streams/recv.rs
+++ b/src/proto/streams/recv.rs
@@ -590,7 +590,7 @@ impl Recv {
     /// Handle remote sending an explicit RST_STREAM.
     pub fn recv_reset(&mut self, frame: frame::Reset, stream: &mut Stream) {
         // Notify the stream
-        stream.state.recv_reset(frame.reason());
+        stream.state.recv_reset(frame.reason(), stream.is_pending_send);
 
         stream.notify_send();
         stream.notify_recv();
--- a/src/proto/streams/state.rs
+++ b/src/proto/streams/state.rs
@@ -218,10 +218,10 @@ impl State {
 
         match self.inner {
             Closed(Cause::EndStream) if queued => {
-                // If the stream has a queued EOS frame, reschedule it to
-                // drain the queue and reset, instead.
+                // If the stream has a queued EOS frame, transition to peer
+                // reset.
                 trace!("recv_reset: reason={:?}; queued=true", reason);
-                self.inner = Closed(Cause::Scheduled(reason));
+                self.inner = Closed(Cause::Proto(reason));
             },
             Closed(..) => {},
             _ => {
--- a/src/proto/streams/state.rs
+++ b/src/proto/streams/state.rs
@@ -214,8 +214,15 @@ impl State {
     }
 
     /// The remote explicitly sent a RST_STREAM.
-    pub fn recv_reset(&mut self, reason: Reason) {
+    pub fn recv_reset(&mut self, reason: Reason, queued: bool) {
+
         match self.inner {
+            Closed(Cause::EndStream) if queued => {
+                // If the stream has a queued EOS frame, reschedule it to
+                // drain the queue and reset, instead.
+                trace!("recv_reset: reason={:?}; queued=true", reason);
+                self.inner = Closed(Cause::Scheduled(reason));
+            },
             Closed(..) => {},
             _ => {
                 trace!("recv_reset; reason={:?}", reason);
@@ -316,6 +323,14 @@ impl State {
             _ => false,
         }
     }
+
+    // /// Returns true if the stream was closed by sending an EOS.
+    // pub fn is_eos(&self) -> bool {
+    //     match self.inner {
+    //         Closed(Cause::EndStream) => true,
+    //         _ => false,
+    //     }
+    // }
 
     /// Returns true if a stream is open or half-closed.
     pub fn is_at_least_half_open(&self) -> bool {
--- a/src/proto/streams/state.rs
+++ b/src/proto/streams/state.rs
@@ -324,14 +324,6 @@ impl State {
         }
     }
 
-    // /// Returns true if the stream was closed by sending an EOS.
-    // pub fn is_eos(&self) -> bool {
-    //     match self.inner {
-    //         Closed(Cause::EndStream) => true,
-    //         _ => false,
-    //     }
-    // }
-
     /// Returns true if a stream is open or half-closed.
     pub fn is_at_least_half_open(&self) -> bool {
         match self.inner {
--- a/tests/stream_states.rs
+++ b/tests/stream_states.rs
@@ -812,7 +812,9 @@ fn exceed_max_streams() {
 
 #[test]
 fn rst_while_closing() {
-    // Test to reproduce panic in issue #246.
+    // Test to reproduce panic in issue #246 --- receipt of a RST_STREAM frame
+    // on a stream in the Half Closed (remote) state with a queued EOS causes
+    // a panic.
     let _ = ::env_logger::try_init();
     let (io, srv) = mock::new();
 
@@ -825,7 +827,11 @@ fn rst_while_closing() {
         )
         .send_frame(frames::headers(1).response(200))
         .send_frame(frames::headers(1).eos())
+        // Idling for a moment here is necessary to ensure that the client
+        // enqueues its TRAILERS frame *before* we send the RST_STREAM frame
+        // which causes the panic.
         .idle_ms(1)
+        // Send the RST_STREAM frame which causes the client to panic.
         .send_frame(frames::reset(1).cancel())
         .recv_frame(frames::headers(1).eos())
         .recv_frame(frames::go_away(0))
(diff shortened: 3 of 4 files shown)


tests/stream_states.rs
```diff
@@ -808,3 +808,65 @@ fn send_data_after_headers_eos() {
...
+        )
+        .send_frame(frames::headers(1).response(200))
+        .send_frame(frames::headers(1).eos())
+        // Idling for a moment here is necessary to ensure that the client
+        // enqueues its TRAILERS frame *before* we send the RST_STREAM frame
+        // which causes the panic.
+        .idle_ms(1)
+        // Send the RST_STREAM frame which causes the client to panic.
+        .send_frame(frames::reset(1).cancel())
+        .recv_frame(frames::reset(1).cancel())
```
carllerche: I believe that, ideally, we would avoid sending a RST_STREAM. While the peer should be able to tolerate this, it is not necessary as a `RST_STREAM` in either direction results in the stream transitioning to closed.

hawkw: Yeah, I think I misunderstood a sentence in the spec saying that we should _allow_ this as saying that we _should_ do this. Had I known that earlier, this PR would have been finished much quicker. My bad!

Corrected this behaviour in 861b10934fb7ac305dbbef020fce9ea3e8c2dda1.

src/proto/streams/state.rs
```diff
@@ -317,6 +324,14 @@ impl State {
         }
     }
 
+    // /// Returns true if the stream was closed by sending an EOS.
+    // pub fn is_eos(&self) -> bool {
```
carllerche: Oops, I meant to mention this in the first review, would it be possible to remove this commented out code?

hawkw: Oh, yeah, missed that when committing. Sorry.

src/proto/streams/recv.rs
```diff
@@ -590,7 +590,7 @@ impl Recv {
     /// Handle remote sending an explicit RST_STREAM.
     pub fn recv_reset(&mut self, frame: frame::Reset, stream: &mut Stream) {
         // Notify the stream
-        stream.state.recv_reset(frame.reason());
+        stream.state.recv_reset(frame.reason(), stream.is_pending_send);
```
seanmonstar: I'd love to eventually make the `state` private. It reads weird that we need to pass in extra information about the stream itself. (Not your fault, I'm just pining for a better future some day :D).

tests/stream_states.rs
```diff
@@ -808,3 +808,65 @@ fn send_data_after_headers_eos() {
...
+        .recv_frame(
+            frames::headers(1)
+                .request("GET", "https://example.com/")
+        )
+        .send_frame(frames::headers(1).response(200))
+        .send_frame(frames::headers(1).eos())
+        // Idling for a moment here is necessary to ensure that the client
+        // enqueues its TRAILERS frame *before* we send the RST_STREAM frame
+        // which causes the panic.
+        .idle_ms(1)
```
seanmonstar: Is this racey at all? Maybe we need a way to tell the mocks to wait on a future, to allow for synchronization?

src/proto/streams/state.rs
```diff
@@ -214,8 +214,15 @@ impl State {
...
     /// The remote explicitly sent a RST_STREAM.
-    pub fn recv_reset(&mut self, reason: Reason) {
+    pub fn recv_reset(&mut self, reason: Reason, queued: bool) {
+
         match self.inner {
+            Closed(Cause::EndStream) if queued => {
+                // If the stream has a queued EOS frame, transition to peer
+                // reset.
+                trace!("recv_reset: reason={:?}; queued=true", reason);
+                self.inner = Closed(Cause::Proto(reason));
```
seanmonstar: So this fixes it because while the stream is still queued, since it's state is changing, the queue later on skips this frame, right?

hawkw: That's correct.

tests/stream_states.rs
hawkw: It could be potentially racy; changing the mocks to allow waiting on a future seems like a better approach long-term.

src/proto/streams/recv.rs
hawkw: As far as I can tell, there doesn't *appear* to be any harm in making the forcing the transition for *all* `Closed(Cause::EndStream)` states on receipt of RST_STREAM regardless of whether or not the stream has pending send; but @carllerche suggested that we only transition to `Closed(Cause::Proto)` if the stream is queued.

carllerche: I prefer to stay on the side of conservative and only make changes that we know are needed.

If there are no queued send frames, then the stream state is actually closed and the received reset should be ignored (I think this is already done)

seanmonstar: I don't think I was clear. I get why we're using the queued information, that's fine! We don't want to trigger a reset error for a stream the user has already closed locally.

I was more concerned that it's weird that we pass information of the stream as an argument onto the state property. I'd want to one day change the methods on the `state` to be methods on `stream` instead.

In other words, ignore me. I'm just sleep deprived. Move along, nothing to see here!

PR #249, opened 2018-03-28
Add initial_connection_window_size to Builders

olix0r:
There is currently no way to configure the initial target window size
for connections. The `Builder::initial_connection_window_size` utility
makes this configurable so that all new server connections have this
target window size set.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1020,7 +1020,8 @@ where
 
 impl<T, B: IntoBuf> Future for Handshake<T, B>
     where T: AsyncRead + AsyncWrite,
-          B: IntoBuf + 'static,
+          B: IntoBuf,
+          B::Buf: 'static,
 {
     type Item = Connection<T, B>;
     type Error = ::Error;
--- a/src/server.rs
+++ b/src/server.rs
@@ -261,6 +261,9 @@ pub struct Builder {
 
     /// Initial `Settings` frame to send as part of the handshake.
     settings: Settings,
+
+    /// Initial target window size for new connections.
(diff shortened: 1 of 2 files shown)


src/server.rs
```diff
@@ -980,7 +1020,7 @@ where
 
 impl<T, B: IntoBuf> Future for Handshake<T, B>
     where T: AsyncRead + AsyncWrite,
-          B: IntoBuf,
+          B: IntoBuf + 'static,
```
hawkw: What necessitated the addition of the static bound?

seanmonstar: Yea, I think this additional bounds would be a breaking change?

olix0r: without it:

```
:; cargo check
   Compiling h2 v0.1.2 (file:///Users/ver/b/rs/h2)
error[E0310]: the associated type `<B as bytes::IntoBuf>::Buf` may not live long enough
    --> src/server.rs:1079:19
     |
1079 |                 c.set_target_window_size(sz);
     |                   ^^^^^^^^^^^^^^^^^^^^^^
     |
     = help: consider adding an explicit lifetime bound `<B as bytes::IntoBuf>::Buf: 'static`...
note: ...so that the type `<B as bytes::IntoBuf>::Buf` will meet its required lifetime bounds
    --> src/server.rs:1079:19
     |
1079 |                 c.set_target_window_size(sz);
     |                   ^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error

```

olix0r: fwiw, the clientside handshake already had `B::Buf: 'static,`

PR #250, opened 2018-03-28
Add Graceful Shutdown support

### LINKED ISSUE #245: Graceful Shutdown
The `GOAWAY` frame can be used for both errors and for closing cleanly.

> The sender of a GOAWAY frame might gracefully shut down a connection by sending a GOAWAY frame, maintaining the connection in an "open" state until all in-progress streams complete.

This issue is about the 3ish ways of a user initiating a shutdown.

1. **Shutdown immediately**: This is what `server::Connection::close_connection` does currently. Once the `GOAWAY` frame is flushed, the rest of the state machine just exits.
2. **Shutdown gracefully**: This would be when a server no longer wants to accept any more streams than the `last_processed_id`, but will allow reading and writing of still open streams before that ID.
3. **Shutdown notice**: The spec mentions that when a peer wants to start shutting down, it should first send a `GOAWAY` of the max stream ID, and then about 1 RTT later, proceed to our number 2. This allows for any streams that were going to be received around the same time the GOAWAY was flushed to still be accepted, and well-behaving peers SHOULDN'T create anymore streams after those. Of course, bad peers will be bad, but a user of this library can always wrap the shutdown future in a timeout.

We need to decide the API and contract exposed by this library. I suggest this:

1. We change `close_connection` to `close_connection(reason: Reason)`. Behavior is the same. Other names are `disconnect`, `close_immediately`, `shutdown`, `goaway`.
2. We add `close_gracefully()`, which sends a `GOAWAY` of the last processed ID, and then the `Connection` future completes after the allowed streams close up.
3. We add `send_graceful_shutdown()`, which just queues a `GOAWAY` of max stream ID, and nothing else. It's up the user to schedule how they'd like to later call `close_gracefully`.

seanmonstar:
If graceful shutdown is initiated, a GOAWAY of the max stream ID - 1 is
sent, followed by a PING frame, to measure RTT. When the PING is ACKed,
the connection sends a new GOAWAY with the proper last processed stream
ID. From there, once all active streams have completely, the connection
will finally close.

Closes #245


olix0r:
note there's a legit test failure in CI:
```
---- rst_while_closing stdout ----
	thread 'rst_while_closing' panicked at 'Idle received unexpected frame on handle; frame=Ok(Ready(Some(Frame::Headers(Headers { stream_id: StreamId(1), stream_dep: None, flags: HeadersFlag { end_stream: true, end_headers: true, padded: false, priority: false } }))))', tests/support/mock.rs:551:17
note: Run with `RUST_BACKTRACE=1` for a backtrace.
failures:
    rst_while_closing
test result: FAILED. 14 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
error: test failed, to rerun pass '--test stream_states'
```

seanmonstar:
/me shakes fist at flaky test XD

PR #253, opened 2018-04-05
Bump version to v0.1.4

seanmonstar:


Cargo.toml
```diff
@@ -33,7 +33,7 @@ unstable = []
 futures = "0.1"
 tokio-io = "0.1.4"
 bytes = "0.4"
-http = "0.1"
+http = "0.1.3"
```
seanmonstar: I suspected that we actually needed something newer than 0.1.0, so I downgraded (`cargo update -p http --precise 0.1.0`) and found that we used `Scheme` constants. Once setting to precisely `0.1.3`, `cargo build` succeeds.

PR #258, opened 2018-04-14
 Reset any queued stream on receipt of remote reset 

### LINKED ISSUE #256: assertion failed: stream.buffered_send_data >= len
I was trying to make a test case and I ran into this panic. Seems like we `clear_queue` when receiving an `RST_STREAM`, which resets `buffered_send_data`, but the next time in `pop_frame` we don't see `is_peer_reset()` because the stream state is already `Closed(Scheduled(CANCEL))`.

Here's the test case that triggers the panic. It's a bit of a mess and it probably shouldn't pass as-is, but it shouldn't panic where it does.

```rust
#[test]
fn rst_while_sending() {

    let mock = mock_io::Builder::new()
        .write(b"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n")
        .write(frames::SETTINGS)
        .read(frames::SETTINGS)
        .wait(Duration::from_millis(10))
        .write(&[
            // POST /
            0, 0, 16, 1, 4, 0, 0, 0, 1, 131, 135, 65, 139, 157, 41,
            172, 75, 143, 168, 233, 25, 151, 33, 233, 132,
        ])
        .write(&[
            // DATA
            0, 64, 0, 0, 0, 0, 0, 0, 1
        ])
        .write(&[
            // DATA
            0; 16384
        ])
        .write(frames::SETTINGS_ACK)
        .read(frames::SETTINGS_ACK)
        .read(&[
            // RST_STREAM
            0, 0, 4, 3, 0, 0, 0, 0, 1,
            0, 0, 0, 2,
        ])
        .wait(Duration::from_millis(1000))
        .write(&[
            // DATA
            0, 64, 0, 0, 0, 0, 0, 0, 1
        ])
        .write(&[
            // DATA
            0; 16384
        ])
        .write(&[
            // DATA
            0, 64, 0, 0, 1, 0, 0, 0, 1
        ])
        .write(&[
            // DATA
            0; 16384
        ])
        .write(&[0, 0, 8, 7, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1])
        .build();

    use futures::Async;
    let client = client::handshake(mock)
        .expect("handshake")
        .and_then(|(mut client, mut conn)| {
            let request = Request::builder()
                .method(Method::POST)
                .uri("https://http2.akamai.com/")
                .body(())
                .unwrap();

            let (_resp, mut stream) = client.send_request(request, false).expect("send_request");
            stream.send_data(vec![0u8; 16384*3].into(), true).unwrap();
            drop(stream);
            drop(_resp);
            futures::future::poll_fn(move || {
                trace!("POLLING CONN");
                match conn.poll() {
                    Ok(Async::NotReady) => trace!("CONN: NOT READY"),
                    r => {
                        trace!("CONN: READY {:?}", r);
                    }
                }
                Ok(Async::NotReady)
            }).map(|()| {
                trace!("DONE");
            })
        });

        client.wait().expect("wait");
}
```

hawkw:
Fixes #256.

This PR changes `state::recv_reset` so any closed stream with queued send is immediately reset (and thus, the queue is cleared) on receipt of a `RST_STREAM` frame from the remote. 

This fixes the panic encountered by the test @goffrie posted in #256, where the stream is scheduled to close, receives a `RST_STREAM` frame, and sets the buffered capacity to 0, but isn't removed from the send queue, so we hit an assertion (or overflow, if debug assertions are disabled) when subtracting a sent frame's size from the buffered size.

I've determined that this does fix the panic in @goffrie's test, but the test instead hangs forever instead of passing. I haven't yet been able to rewrite the test such that it hits the panic on master but doesn't hang when the issue is resolved. I intend to continue working on testing this, but I'm putting this up for review now nonetheless.


src/proto/streams/state.rs
```diff
@@ -217,10 +217,11 @@ impl State {
     pub fn recv_reset(&mut self, reason: Reason, queued: bool) {
 
         match self.inner {
-            Closed(Cause::EndStream) if queued => {
+            Closed(prior_reason) if queued => {
```
olix0r: this looks like the only functional change in this PR... can you explain how this resolves the issue?

carllerche: +1. It would be good to include a comment explaining this as it is definitely subtle.

olix0r: IIUC, the match can be condensed as:

```rust
    pub fn recv_reset(&mut self, reason: Reason, queued: bool) {
        match self.inner {
            Closed(..) if !queued => {},
            state => {
                trace!("recv_reset; {:?} => reason=Closed(Proto({:?})", state, reason);
                self.inner = Closed(Cause::Proto(reason));
            }
        }
}
```

I want to make sure I understand this properly, though.  What are the states where we can be closed and have queued data?

https://github.com/carllerche/h2/blob/b1d14caa2251cfa6316d73ad77409fb0be4a41f1/src/proto/streams/state.rs#L73-L87

In particular, I'm not sure if it makes sense to change the closed type when it's already `Proto` (it's probably harmless? but I'm not sure of the intended behavior).  Also, I'm not sure we'd want to override an `Io` error (though I'm not sure if it's actually possible to receive a reset after an Io error).

hawkw: This is fair, I can make the match more specific. 

My understanding is that we can have queued data in the following cases:
 - when the cause `Cause::Scheduled`,
- when the cause is `Cause::EndStream`, which would indicate that we've enqueued an EOS frame but it has yet to be sent (either because there are other frames ahead of it in the send queue, or we just haven't popped it yet)

AFAIK, `queued` should not be true with any other `Cause` variants, but I'll change the code to actually encode this nonetheless.

hawkw: @olix0r, hopefully 05a6017ba74d7561ad15ddc6d54363687a62db8f is clearer?

goffrie:
Hm, there's one more way that we can call `clear_queue`, which is via `recv_err`; I believe this is able to cause a similar issue wherein the stream's state doesn't necessarily become `is_peer_reset` and so [the check here](https://github.com/carllerche/h2/blob/23090c9fed169985a591b864541a2ba02e87c016/src/proto/streams/prioritize.rs#L585-L587) doesn't trigger. For example I've seen this happen when receiving a GO_AWAY frame, which has caused [this assertion](https://github.com/carllerche/h2/blob/23090c9fed169985a591b864541a2ba02e87c016/src/proto/streams/prioritize.rs#L695) to trip because the frames were all cleared.

hawkw:
@goffrie Ah, thanks for pointing that out --- I think we'll want to add similar logic to `recv_error` for when a stream has frames in the send queue. I'll look into that in an additional PR.

carllerche:
Thanks for getting this done.

PR #260, opened 2018-04-17
Add some missing bounds checks.

goffrie:
Ran a fuzzer and found a few places where we were panicking instead of returning errors.

### DIFF
--- a/src/frame/headers.rs
+++ b/src/frame/headers.rs
@@ -153,6 +153,9 @@ impl Headers {
 
         // Read the padding length
         if flags.is_padded() {
+            if src.len() < 1 {
+                return Err(Error::MalformedMessage);
+            }
             // TODO: Ensure payload is sized correctly
             pad = src[0] as usize;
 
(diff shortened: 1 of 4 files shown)


src/frame/headers.rs
```diff
@@ -153,6 +153,9 @@ impl Headers {
 
         // Read the padding length
         if flags.is_padded() {
+            if src.len() < 1 {
+                return Err(Error::MalformedMessage);
+            }
             // TODO: Ensure payload is sized correctly
```
seanmonstar: Maybe there's more involved, but wanted to check: does this addition essentially complete this TODO?

src/proto/streams/streams.rs
```diff
@@ -322,6 +322,10 @@ where
         let last_stream_id = frame.last_stream_id();
         let err = frame.reason().into();
 
+        if actions.recv.max_stream_id() < last_stream_id {
```
seanmonstar: I got a bit confused with this, until I went digging into `recv` to read the comments about `max_stream_id`. Whatcha think if there was a comment right here just saying to the effect of "if a new GOAWAY has a higher stream id than a previous GOAWAY, that's bad"?

src/frame/headers.rs
goffrie: Seems like it; I removed the comment.

src/proto/streams/streams.rs
goffrie: :+1:

carllerche:
Thanks @goffrie!

Looks like the fuzzing has been submitted in another PR. I'm good with this!

PR #261, opened 2018-04-20
Be more lenient with streams in the `pending_send` queue.

goffrie:
The `is_peer_reset()` check doesn't quite cover all the cases where we call
`clear_queue`, such as when we call `recv_err`. Instead of trying to make the
check more precise, let's gracefully handle spurious entries in the queue.

This fixes the issue I mentioned at https://github.com/carllerche/h2/pull/258#issuecomment-381768145.


hawkw:
At a glance, this seems right to me --- it's similar to the first thing I tried while working on #247, which appeared to work correctly. However, I'd like to hear from @carllerche before approving it.

src/proto/streams/prioritize.rs
```diff
@@ -694,13 +686,20 @@ impl Prioritize {
...
+
+                                let frame = frame::Reset::new(stream.id, reason);
+                                Frame::Reset(frame)
+                            } else {
+                                // If the stream receives a RESET from the peer, it may have
+                                // had data buffered to be sent, but all the frames are cleared
+                                // in clear_queue(). Instead of doing O(N) traversal through queue
+                                // to remove, lets just ignore the stream here.
+                                trace!("removing dangling stream from pending_send");
+                                counts.transition_after(stream, is_counted, is_pending_reset);
```
carllerche: Could you explain why this needs to be called here? I don't *think* this was called before (and I could believe it was another bug).

goffrie: I agree that it was not called before. I think that it should be, otherwise we could leak streams - it's possible that the `pending_send` queue here was holding onto the last reference to the stream.

carllerche: I think it is highly likely that this change is correct, but i'd like to see if we can figure out a failing test case.

carllerche:
I believe that this change is probably the right direction. Panicking in production is no good. It would be good to figure out how to catch any bugs that are let through though. At the very least a `debug_assert` is good.

goffrie:
I added a `debug_assert!`, though I'm not sure what precisely you had in mind - this one's pretty general. (Notably, asserting `is_reset()` doesn't work because the stream can be in the `Closed(EndStream)` state and still get cleared.)

carllerche:
So I spent a bunch of time trying to reproduce the issue described in [the comment](https://github.com/carllerche/h2/pull/258#issuecomment-381768145).

All the paths that I tracked down via `recv_err` would not hit the issue.

Either `recv_err` is called because of a connection level issue. At this point, the code switches to GOAWAY and won't go through `prioritize` anymore. Or, `recv_err` is called when sending an explicit reset, in which case there will always be a frame to pop (the reset frame).

That said, I think that the consensus is that this is still a good change even if it does not fix any bugs.

carllerche:
k, i will try that

carllerche:
Thanks to the fuzzer (#263), I have identified the logic path that results in the panic.

Unfortunately, it relies on a very specific pattern of the underlying I/O's write fn returning `Ready` / `NotReady`, which means that writing a robust test is not really possible. Any change to the `h2` implementation could break the test.

So, given that, I think the best course of action is just going to be to include the fuzzing run and call it a day.

I still need to verify the `counts.transition_after(stream, is_counted, is_pending_reset);` line. I believe that this does fix a bug, but I would like to confirm this first. This will be my next focus.

@goffrie Thanks for your patience. My assumption is that you are currently using a fork that includes this patch, so you are not currently blocked. If this assumption is incorrect, please let me know and we can merge the PR before I complete the work stated above.

goffrie:
That's correct; take your time. Thanks for your work on this :)

carllerche:
So, I haven't reproed the issue w/ the missing count decrement, but I did find & fix a number of other related bugs (#273).

carllerche:
Alright, I have rebased this change.

PR #266, opened 2018-04-26
prevent a leak of 'active streams' if client request has user error

seanmonstar:
A stream is inserted into the store, but `send.send_headers` may notice a `UserError` and return. The stream then is just stuck in the store forever, and the connection assumes it is busy and can't shutdown.

### DIFF
--- a/tests/client_request.rs
+++ b/tests/client_request.rs
@@ -440,9 +440,14 @@ fn request_with_connection_headers() {
     let _ = ::env_logger::try_init();
     let (io, srv) = mock::new();
 
-    let srv = srv.assert_client_handshake()
+    // can't assert full handshake, since client never sends a request, and
+    // thus never bothers to ack the settings...
+    let srv = srv.read_preface()
         .unwrap()
-        .recv_settings()
+        .recv_frame(frames::settings())
+        // goaway is required to make sure the connection closes because
+        // of no active streams
+        .recv_frame(frames::go_away(0))
         .close();
 
     let headers = vec![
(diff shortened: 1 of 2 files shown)


tests/client_request.rs
```diff
@@ -440,9 +440,14 @@ fn request_with_connection_headers() {
     let _ = ::env_logger::try_init();
     let (io, srv) = mock::new();
 
-    let srv = srv.assert_client_handshake()
+    // can't assert full handshake, since client never send a request, and
+    // thus never boths to ack the settings...
```
olix0r: client never _sends_ a request, and thus never _bothers_ ?

tests/client_request.rs
```diff
@@ -440,9 +440,14 @@ fn request_with_connection_headers() {
...
+    let srv = srv.read_preface()
         .unwrap()
-        .recv_settings()
+        .recv_frame(frames::settings())
+        // goaway is just to make sure the connection closes because
+        // of no active streams, not an error
+        .recv_frame(frames::go_away(0))
```
olix0r: Would this test fail without the change? Even with the comments, it's not immediately clear to me how this test exercises the error case.

tests/client_request.rs
seanmonstar: Yea, currently the client includes the ack with the first request.

tests/client_request.rs
seanmonstar: Yep, by adding this goaway receive check, the test will hang, even though all sender handles have been dropped, because the connection thinks there's some 6 active streams.

tests/client_request.rs
olix0r: mostly just noting some typos ;)

PR #272, opened 2018-05-04
Verify that a PUSH_PROMISEd stream ID is not already in use.

goffrie:
One more bug found in fuzzing.

### DIFF
--- a/src/proto/streams/recv.rs
+++ b/src/proto/streams/recv.rs
@@ -535,7 +535,7 @@ impl Recv {
         store: &mut Store,
     ) -> Result<(), RecvError> {
         // First, make sure that the values are legit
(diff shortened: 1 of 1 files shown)


carllerche:
Thanks for this.

It looks like, even with this change, we are not respecting HTTP/2.0. Specifically, stream identifiers should be monotonically increasing ([see here](https://httpwg.org/specs/rfc7540.html#StreamIdentifiers). This change does not enforce the monotonic increase requirement.

I will see if I can work on that fix.

carllerche:
Ok, I opened #275 which covers handles the case fixed by this PR as well as others (related). I will close this PR in favor of that one. Thanks!

PR #273, opened 2018-05-06
Misc bug fixes

carllerche:
In an effort to test #261, I added new assertions checking that stream counts is zero when the connection finalizes and that stream state gets released (the slab is empty on drop).

Adding these assertions uncovered a number of bugs. This PR fixes the discovered bugs.

I would recommend reviewing each commit individually as they are pretty independent.

### DIFF
--- a/src/proto/connection.rs
+++ b/src/proto/connection.rs
@@ -340,7 +340,8 @@ where
                 },
                 None => {
                     trace!("codec closed");
-                    self.streams.recv_eof(false);
+                    self.streams.recv_eof(false)
+                        .ok().expect("mutex poisoned");
                     return Ok(Async::Ready(()));
                 },
             }
@@ -403,6 +404,7 @@ where
     B: IntoBuf,
 {
     fn drop(&mut self) {
-        self.streams.recv_eof(true);
+        // Ignore errors as this indicates that the mutex is poisoned.
+        let _ = self.streams.recv_eof(true);
     }
 }
--- a/src/proto/connection.rs
+++ b/src/proto/connection.rs
@@ -340,7 +340,7 @@ where
                 },
                 None => {
                     trace!("codec closed");
-                    self.streams.recv_eof();
+                    self.streams.recv_eof(false);
                     return Ok(Async::Ready(()));
                 },
             }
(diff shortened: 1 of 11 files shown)


carllerche:
Still looks like there are bugs!

src/proto/connection.rs
```diff
@@ -397,3 +396,13 @@ where
...
     }
 }
+
+impl<T, P, B> Drop for Connection<T, P, B>
+where
+    P: Peer,
+    B: IntoBuf,
+{
+    fn drop(&mut self) {
+        self.streams.recv_eof();
```
seanmonstar: If this is dropped because the thread panicked, this will trigger an abort (`recv_eof` unwraps a lock).

src/proto/streams/counts.rs
```diff
@@ -60,11 +60,13 @@ impl Counts {
     /// # Panics
     ///
     /// Panics on failure as this should have been validated before hand.
-    pub fn inc_num_recv_streams(&mut self) {
+    pub fn inc_num_recv_streams(&mut self, stream: &mut store::Ptr) {
         assert!(self.can_inc_num_recv_streams());
+        assert!(!stream.is_counted);
```
seanmonstar: This question applies for most of the new asserts: do we want to be adding new `assert`s everywhere, or just `debug_assert`? Is it better for this to crash in production?

src/proto/connection.rs
carllerche: Ok, we can have `recv_eof()` return `Err` instead of unwrapping the poisoned mutex.

src/proto/streams/counts.rs
carllerche: I think it depends on the severity of the assertion failure.

If this fails, then we could get into situations where subtractions wrap around. I think this should be guarded again?

carllerche: Any further thoughts on this?

seanmonstar: Hm, so of course, crashing in production is a major bummer, as it destroys **all** in progress requests... In this specific case, it seems like if this assert were to fail, then the subtraction would eventually underflow and crash anyways. So perhaps this is fine...

(It'd probably be neat if we could just encode this in the types somehow instead, but :shrug:)

carllerche: Yeah.... ideally it would only invalidate the connection.

That said, we have a whole bunch of existing `assert!` statements. Maybe we should punt this as an open question: "how to better deal w/ internal errors".

PR #275, opened 2018-05-10
Enforce monotonic stream IDs for push promises

carllerche:
Previously, monotonic stream IDs (spec 5.1.1) for push promises were not
enforced. This was due to push promises going through an entirely
separate code path than normally initiated streams.

This patch unifies the code path for initializing streams via both
HEADERS and PUSH_PROMISE. This is done by first calling `recv.open` in
both cases.

Closes #272

### DIFF
--- a/src/proto/peer.rs
+++ b/src/proto/peer.rs
@@ -61,20 +61,23 @@ impl Dyn {
     }
 
     /// Returns true if the remote peer can initiate a stream with the given ID.
-    pub fn ensure_can_open(&self, id: StreamId) -> Result<(), RecvError> {
-        if !self.is_server() {
-            trace!("Cannot open stream {:?} - not server, PROTOCOL_ERROR", id);
-            // Remote is a server and cannot open streams. PushPromise is
-            // registered by reserving, so does not go through this path.
-            return Err(RecvError::Connection(Reason::PROTOCOL_ERROR));
-        }
+    pub fn ensure_can_open(&self, id: StreamId, push_promise: bool) -> Result<(), RecvError> {
+        if self.is_server() {
+            // Ensure that the ID is a valid client initiated ID
+            if push_promise || !id.is_client_initiated() {
+                trace!("Cannot open stream {:?} - not client initiated, PROTOCOL_ERROR", id);
+                return Err(RecvError::Connection(Reason::PROTOCOL_ERROR));
+            }
 
-        // Ensure that the ID is a valid server initiated ID
-        if !id.is_client_initiated() {
-            trace!("Cannot open stream {:?} - not client initiated, PROTOCOL_ERROR", id);
-            return Err(RecvError::Connection(Reason::PROTOCOL_ERROR));
-        }
+            Ok(())
+        } else {
+            // Ensure that hte ID is a valid server initiated ID
+            if !push_promise || !id.is_server_initiated() {
+                trace!("Cannot open stream {:?} - not server initiated, PROTOCOL_ERROR", id);
+                return Err(RecvError::Connection(Reason::PROTOCOL_ERROR));
+            }
 
-        Ok(())
+            Ok(())
+        }
     }
 }
--- a/src/proto/streams/recv.rs
+++ b/src/proto/streams/recv.rs
@@ -1,5 +1,4 @@
 use super::*;
-use super::store::Resolve;
 use {frame, proto};
 use codec::{RecvError, UserError};
 use frame::{Reason, DEFAULT_INITIAL_WINDOW_SIZE};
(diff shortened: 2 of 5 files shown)


src/proto/peer.rs
```diff
@@ -61,20 +61,23 @@ impl Dyn {
...
+            }
 
-        // Ensure that the ID is a valid server initiated ID
-        if !id.is_client_initiated() {
-            trace!("Cannot open stream {:?} - not client initiated, PROTOCOL_ERROR", id);
-            return Err(RecvError::Connection(Reason::PROTOCOL_ERROR));
-        }
+            Ok(())
+        } else {
+            // Ensure that hte ID is a valid server initiated ID
```
olix0r: typo

src/proto/streams/recv.rs
```diff
@@ -644,17 +620,9 @@ impl Recv {
     }
 
     /// Returns true if the remote peer can reserve a stream with the given ID.
```
olix0r: Is this comment still correct? Does this now apply to either peer?

olix0r: no, nm. i get it now

src/proto/streams/recv.rs
```diff
@@ -121,11 +120,12 @@ impl Recv {
     pub fn open(
         &mut self,
         id: StreamId,
+        push_promise: bool,
```
seanmonstar: In a few cases, passing a boolean argument is pretty clear. In this case, it's not immediately obvious what `recv.open(id, false, counts)` means. This could be an enum instead, so that it's clear at the call-site.

```rust
recv.open(id, Open::Push, counts);
// and ...
recv.open(id, Open::Headers, counts);
```

src/proto/streams/streams.rs
```diff
@@ -402,44 +404,61 @@ where
...
 
-        let res = {
-            let stream = match me.store.find_mut(&id) {
-                Some(stream) => stream.key(),
-                None => return Err(RecvError::Connection(Reason::PROTOCOL_ERROR)),
-            };
+        // First, ensure that the initiating stream is still in a valid state.
+        let parent_key = match me.store.find_mut(&id) {
+            Some(stream) => {
+                if id > me.actions.recv.max_stream_id() {
```
seanmonstar: It took me a minute to go look this up to remember why the check was done. What do you think of a tiny comment hear explaining that if a `GOAWAY` has been started, frames higher than the sent id can be ignored?

src/proto/streams/streams.rs
```diff
@@ -402,44 +404,61 @@ where
...
-                //
-                // Otherwise, this MUST be an connection error.
-                assert!(!err.is_stream_error());
-                Err(err)
-            }
-        } else {
-            res
+        // Next, open the stream.
+        if me.actions.recv.open(promised_id, true, &mut me.counts)?.is_none() {
+            return Ok(());
```
seanmonstar: I also had to go look up why `open` might return `None`. Maybe a comment in here about the stream being refused and it is already scheduled?

Maybe we should someday also make the return value of `open` clearer in the first place...

src/proto/streams/streams.rs
```diff
@@ -402,44 +404,61 @@ where
...
+                    me.actions.send.init_window_sz(),
+                    me.actions.recv.init_window_sz())
+            });
+
+            let actions = &mut me.actions;
+
+            me.counts.transition(stream, |counts, stream| {
+                let res = actions.recv.recv_push_promise(frame, stream);
+
+                let mut send_buffer = self.send_buffer.inner.lock().unwrap();
```
seanmonstar: We don't need to take this lock if there wasn't an error. The previous code checked before locking.

carllerche: I don't disagree with you, but I opted to unconditionally take the lock because:

* There are plenty of other places in the code that unconditionally take the lock, and any fixes should be applied globally.
* The function being called takes the `res` and conditionally does work on error, it seems like it would be better to pass the mutex in vs. the buffer. Either way, it felt weird conditionally calling a fn based on the logic that the fn is supposed to deal with for you.
* There are no measurements that unconditionally taking the lock there matters and it is an edge case.

Given this, I opted for simpler code in favor of fixing it universally at a later time. If you disagree, I can revert back.

src/proto/streams/streams.rs
```diff
@@ -652,7 +652,7 @@ where
 
         let key = match me.store.find_entry(id) {
             Entry::Occupied(e) => e.key(),
-            Entry::Vacant(e) => match me.actions.recv.open(id, false, &mut me.counts) {
+            Entry::Vacant(e) => match me.actions.recv.open(id, Open::PushPromise, &mut me.counts) {
```
carllerche: Here is the bug!

PR #277, opened 2018-05-21
Wakeup any waiting tasks when transitioning a stream from pending_ope…

johnklai1:
…n to pending_send.

We were noticing cases where `SendRequest::poll_ready` would return `NotReady`, but than it would never get notified.  It looks like this was triggering when we had streams that were `pending_open`.  This fixes the issue locally but it might not be the exact right fix.

### DIFF
--- a/src/proto/streams/prioritize.rs
+++ b/src/proto/streams/prioritize.rs
@@ -801,6 +801,7 @@ impl Prioritize {
 
                 counts.inc_num_send_streams(&mut stream);
                 self.pending_send.push(&mut stream);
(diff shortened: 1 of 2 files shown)


carllerche:
looks great to me. I would like a test before merging. I can try to work on that too.

johnklai1:
Thanks for taking a look!  I won't be able to get to writing a test until later today but if you can take a look that would be great.

carllerche:
Ok, I added a test. This looks good to me. @seanmonstar ?

PR #279, opened 2018-05-29
add server::SendResponse::poll_reset to listen for reset streams

seanmonstar:
Allows for listening for `RST_STREAM`s that may be sent before a `Response` has been generated, and cancel everything.

### DIFF
--- a/src/proto/streams/send.rs
+++ b/src/proto/streams/send.rs
@@ -1,10 +1,16 @@
 use codec::{RecvError, UserError};
-use codec::UserError::*;
 use frame::{self, Reason};
-use http;
-use super::*;
+use proto;
+use super::{
+    store, Buffer, Codec, Config, Counts, Frame, Prioritize,
+    Prioritized, Store, Stream, StreamId, StreamIdOverflow, WindowSize,
+};
 
 use bytes::Buf;
+use http;
+use futures::{Async, Poll};
+use futures::task::Task;
+use tokio_io::AsyncWrite;
 
 use std::{cmp, io};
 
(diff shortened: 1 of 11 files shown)


src/server.rs
```diff
@@ -972,6 +972,22 @@ impl<B: IntoBuf> SendResponse<B> {
         self.inner.send_reset(reason)
     }
 
+    /// Polls to be notified when the client resets this stream.
+    ///
+    /// If stream is still open, this returns `Ok(Async::NotReady)`, and
```
carllerche: Maybe add a section above this: "This function allows the caller to detect a RST_STREAM before sending the response, providing an opportunity to cancel work. This function must only be called before calling `send_response`." or something like that. Basically, emphasize the intent / requirements.

src/proto/streams/send.rs
```diff
@@ -21,6 +26,14 @@ pub(super) struct Send {
     prioritize: Prioritize,
 }
 
+/// A value to detect which public API has called `poll_reset`.
+#[derive(Debug)]
+pub(crate) enum PollReset {
+    AwaitingHeaders,
+    #[allow(unused)]
```
carllerche: Why is this annotation needed? It seems to be used below.

src/server.rs
seanmonstar: Both of those ideas are in the following lines. Are they not clear?

src/proto/streams/send.rs
seanmonstar: Ah, woops, I implemented just for `SendResponse` first, and so that variant was unused.

src/server.rs
carllerche: It’s covered by the errors section but who reads that?

PR #285, opened 2018-06-14
fix tight loop on aborted connection

aep:
when the underlying IO returns 0 on read, we must stop polling it since
it's closed. Otherwise we'll be stuck in a tight loop.

this fixes https://github.com/sfackler/rust-openssl/issues/949

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1045,6 +1045,12 @@ where
 
         while rem > 0 {
             let n = try_nb!(self.inner_mut().read(&mut buf[..rem]));
+            if n == 0 {
+                return Err(::std::io::Error::new(
+                    ::std::io::ErrorKind::ConnectionReset,
+                    "connection closed unexpectedly",
+                ).into());
+            }
 
             if PREFACE[self.pos..self.pos + n] != buf[..n] {
                 // TODO: Should this just write the GO_AWAY frame directly?
--- a/src/server.rs
+++ b/src/server.rs
@@ -140,7 +140,7 @@ use proto::{self, Config, Prioritized};
 use bytes::{Buf, Bytes, IntoBuf};
 use futures::{self, Async, Future, Poll};
 use http::{Request, Response};
-use std::{convert, fmt, mem};
+use std::{convert, fmt, io, mem};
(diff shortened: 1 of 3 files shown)


src/frame/reason.rs
```diff
@@ -59,6 +59,8 @@ impl Reason {
     pub const INADEQUATE_SECURITY: Reason = Reason(12);
     /// The endpoint requires that HTTP/1.1 be used instead of HTTP/2.
     pub const HTTP_1_1_REQUIRED: Reason = Reason(13);
+    /// The connection was closed
+    pub const CONNECTION_CLOSED : Reason = Reason(14);
```
carllerche: These reasons are mapped to http 2.0 reasons from the spec. In this case, it would either be a protocol error or an I/o error (connection reset by peer).

aep:
>  or an I/o error

sounds more appropriate than protocol error but couldnt find it in the spec

> Also, any way to add a test that covers this case?

sounds like a lot of effort, since we'd have to make an io that deliberatly returns 0 at a specific point.

sfackler:
Seems like it wouldn't be that bad to make a test. Start a server, open up a connection, write `PRI *` and close the connection, and make sure the server handshake returns an error.

aep:
not sure how to do this using the test framework.  this never returns:

```
#[test]
fn server_error_on_unclean_shutdown() {
    use std::io::Write;

    let (io, mut client) = mock::new();

    let mut settings = frame::Settings::default();
    settings.set_max_concurrent_streams(Some(1));

    let mut builder = server::Builder::new();
    builder.max_concurrent_streams(1);


    let srv = server::Builder::new()
        .handshake::<_, Bytes>(io)
        .map_err(|e|println!("{}", e))
        .and_then(|srv| {
            println!("{:?}", srv);
            Ok(())
        });


    client.write_all(b"PRI *").expect("write");
    client.shutdown().expect("shutdown");

    srv.wait().expect("wait");
}

```

aep:
ok i changed mock to just ignore write if its closed, this works. the test now passes with this fix and never returns without it.

PR #290, opened 2018-06-20
Do not distribute the test fixtures in published crates

shepmaster:
### Before

Unpacked: 59M
Compressed: 5.4M

### After

Unpacked: 788K
Compressed: 133K


carllerche:
This will make people emo. Distro packagers want to be able to run tests
from the crate

seanmonstar:
That's a fair point... I'm partial to making the crates.io experience better, which in this case, it's a big difference!

PR #292, opened 2018-07-09
Add `stream_id` accessors to public API types

### LINKED ISSUE #289: Expose stream ID to application
Problem:

Applications may want to access the underlying h2 stream ID for diagnostics, etc. As far as I know, stream IDs are not exposed in the application-facing h2 APIs.

Proposal:

In h2 servers, inbound `http::Request`s should have an `h2::StreamId` extension.

In h2 clients, `h2::client::ResponseFuture` and `h2::SendStream` should both expose a `pub fn stream_id(&self) -> h2::StreamId`; and responses satisfied from this future should have an `h2::StreamId` extension.

I'd expect `h2::StreamId` to implement at least copy and comparison traits.

hawkw:
Problem:

Applications may want to access the underlying h2 stream ID for
diagnostics, etc. Stream IDs were not previously exposed in public APIs.

Solution:

The `frame::StreamId` type is now publically exported and has RustDoc
comments. The public API types `SendStream`, `RecvStream`, 
`ReleaseCapacity`, `client::ResponseFuture`, and `server::SendResponse` 
now all have `stream_id` methods which return the stream ID of the 
corresponding stream.

Closes #289.

### DIFF
--- a/src/frame/stream_id.rs
+++ b/src/frame/stream_id.rs
@@ -20,8 +20,10 @@ pub struct StreamIdOverflow;
 const STREAM_ID_MASK: u32 = 1 << 31;
 
 impl StreamId {
+    /// Stream ID 0.
     pub const ZERO: StreamId = StreamId(0);
 
+    /// The maximum allowed stream ID.
     pub const MAX: StreamId = StreamId(u32::MAX >> 1);
 
     /// Parse the stream ID
@@ -49,6 +51,7 @@ impl StreamId {
         id != 0 && id % 2 == 0
     }
 
+    /// Return a new `StreamId` for stream 0.
     #[inline]
     pub fn zero() -> StreamId {
         StreamId::ZERO
--- a/src/frame/stream_id.rs
+++ b/src/frame/stream_id.rs
@@ -1,6 +1,16 @@
 use byteorder::{BigEndian, ByteOrder};
 use std::u32;
 
+/// A stream identifier, as described in [Section 5.1.1] of RFC 7540.
+///
+/// Streams are identified with an unsigned 31-bit integer. Streams
+/// initiated by a client MUST use odd-numbered stream identifiers; those
+/// initiated by the server MUST use even-numbered stream identifiers.  A
+/// stream identifier of zero (0x0) is used for connection control
+/// messages; the stream identifier of zero cannot be used to establish a
+/// new stream.
+///
+/// [Section 5.1.1]: https://tools.ietf.org/html/rfc7540#section-5.1.1
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub struct StreamId(u32);
 
@@ -25,11 +35,15 @@ impl StreamId {
         (StreamId(unpacked & !STREAM_ID_MASK), flag)
     }
 
+    /// Returns true if this stream ID corresponds to a stream that
+    /// was initiated by the client.
     pub fn is_client_initiated(&self) -> bool {
         let id = self.0;
         id != 0 && id % 2 == 1
     }
 
+    /// Returns true if this stream ID corresponds to a stream that
+    /// was initiated by the server.
     pub fn is_server_initiated(&self) -> bool {
         let id = self.0;
         id != 0 && id % 2 == 0
@@ -40,10 +54,14 @@ impl StreamId {
         StreamId::ZERO
     }
 
+    /// Returns true if this stream ID is zero.
     pub fn is_zero(&self) -> bool {
         self.0 == 0
     }
 
+    /// Returns the next stream ID initiated by the same peer as this stream
+    /// ID, or an error if incrementing this stream ID would overflow the
+    /// maximum.
     pub fn next_id(&self) -> Result<StreamId, StreamIdOverflow> {
         let next = self.0 + 2;
         if next > StreamId::MAX.0 {
--- a/src/proto/streams/streams.rs
+++ b/src/proto/streams/streams.rs
@@ -946,6 +946,10 @@ impl<B> StreamRef<B> {
     {
         self.opaque.clone()
     }
+
+    pub fn stream_id(&self) -> StreamId {
+        self.opaque.stream_id()
+    }
 }
 
 impl<B> Clone for StreamRef<B> {
@@ -1017,6 +1021,13 @@ impl OpaqueStreamRef {
         me.actions
             .recv
             .release_capacity(capacity, &mut stream, &mut me.actions.task)
+    }
+
+    pub fn stream_id(&self) -> StreamId {
+        self.inner.lock()
+            .unwrap()
+            .store[self.key]
+            .id
     }
 }
 
--- a/src/share.rs
+++ b/src/share.rs
@@ -96,6 +96,19 @@ pub struct SendStream<B: IntoBuf> {
     inner: proto::StreamRef<B::Buf>,
 }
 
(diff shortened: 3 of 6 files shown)


src/frame/stream_id.rs
```diff
@@ -10,8 +20,10 @@ pub struct StreamIdOverflow;
 const STREAM_ID_MASK: u32 = 1 << 31;
 
 impl StreamId {
+    /// Stream ID 0.
     pub const ZERO: StreamId = StreamId(0);
 
+    /// The maximum allowed stream ID.
     pub const MAX: StreamId = StreamId(u32::MAX >> 1);
 
     /// Parse the stream ID
```
hawkw: I'm not sure if this function should be publicly exposed or not (perhaps it should be `pub(crate)` instead?)

src/frame/stream_id.rs
```diff
@@ -25,25 +37,34 @@ impl StreamId {
...
 
+    /// Returns true if this stream ID is zero.
     pub fn is_zero(&self) -> bool {
         self.0 == 0
     }
 
+    /// Returns the next stream ID initiated by the same peer as this stream
+    /// ID, or an error if incrementing this stream ID would overflow the
+    /// maximum.
     pub fn next_id(&self) -> Result<StreamId, StreamIdOverflow> {
```
hawkw: Again, I'm not sure whether or not we want users of the library to be able to construct their own `StreamId`s. It seems harmless but I'm not sure if there's any use-case for it. Perhaps this should be `pub(crate)`?

src/proto/streams/streams.rs
```diff
@@ -1018,6 +1022,13 @@ impl OpaqueStreamRef {
             .recv
             .release_capacity(capacity, &mut stream, &mut me.actions.task)
     }
+
+    pub fn stream_id(&self) -> StreamId {
+        self.inner.lock()
+            .unwrap()
```
hawkw: I thought `unwrap`ping this was probably reasonable, rather than returning a `Result`, since if the mutex on the stream store is poisoned, we're in a pretty bad state (and the other functions on `StreamRef` also `unwrap`).

seanmonstar:
Yea, I'm also a little uncertain about exposing the `StreamId` type directly... Would it be more conservative to simply expose a `u32` instead?

hawkw:
@seanmonstar 
> Yea, I'm also a little uncertain about exposing the `StreamId` type directly... Would it be more conservative to simply expose a `u32` instead?

I'm unsure...I think the newtype better expresses intent, and the `is_client_initiated`/`is_server_initiated` functions might potentially be useful. Personally, I think if we're concerned about exposing this type, restricting the constructors to `pub(crate)` is the right thing to do, but I could be convinced otherwise.

carllerche:
@seanmonstar Exposing a new type would be more conservative than `u32`. It could be another `StreamId` type that doesn't have any additional fns on it?

seanmonstar:
True, a public `StreamId` that isn't the same as `frame::StreamId`. Then, we don't have to worry about any traits that are implemented for `frame::StreamId`, and down the line, its backwards compatible to just change to directly exporting `frame::StreamId`.

hawkw:
@seanmonstar WDYT of 557435caa4c68248c0373f9f4b955fefbbac2964?

hawkw:
@seanmonstar 
> Looks good to me, if we also put `pub(crate)` on the internal one, to save us from any accidental exposure.

Will do. Note that making `frame::StreamId` `pub(crate)` means that we also have to change all `pub` functions taking one (including those which are in private modules and thus not exported) explicitly `pub(crate)` as well, to placate the compiler:

```
error[E0446]: private type `frame::stream_id::StreamId` in public interface
  --> src/frame/data.rs:28:5
   |
28 | /     pub fn new(stream_id: StreamId, payload: T) -> Self {
29 | |         assert!(!stream_id.is_zero());
30 | |
31 | |         Data {
...  |
36 | |         }
37 | |     }
   | |_____^ can't leak private type
```
...and so on.

hawkw:
Actually, nevermind; the fact that a lot of these types _are_ publically-exposed if the `unstable` feature is set complicates this a bit. If we want the internal `frame::StreamId` to be `pub(crate)`, then we'll have to have separate `#[cfg("unstable")]` `pub` versions of pretty much every function that takes a `frame::StreamId`. I'm not sure whether or not having it be `pub(crate)` is worth that, WDYT?

seanmonstar:
Hmph. I'm not even particularly sure what the `unstable` feature is really for, so I'll defer the final decision to @carllerche.

hawkw:
@seanmonstar 
> I'm not even particularly sure what the `unstable` feature is really for

I believe it's mainly used so that the test support crate can access things that would otherwise not be public.

carllerche:
Yes, it is for the test stuff.

That being said, with the introduction of tests in a sub crate, there is most likely a way to get rid of that feature flag.

hawkw:
> That being said, with the introduction of tests in a sub crate, there is most likely a way to get rid of that feature flag.

Okay. That seems out of scope for this branch, though?

carllerche:
yes

carllerche:
To comment on this issue, I'm OK exposing the stream ID where appropriate, but it should be a new type `StreamId` that has as few fns on it as possible.

hawkw:
@carllerche 
> To comment on this issue, I'm OK exposing the stream ID where appropriate, but it should be a new type `StreamId` that has as few fns on it as possible.

That's what this branch does as of commit 557435caa4c68248c0373f9f4b955fefbbac2964 --- that commit adds a separate `share::StreamId` type for externally exposing stream IDs, distinct from `frame::StreamId`. 

The conversation @seanmonstar and I were having was about the fact that the `frame::StreamId` type (the internal `StreamId` type) is currently declared as `pub` (and was before I opened this branch), but the module it's declared in is usually not exported, unless the `unstable` feature is set. @seanmonstar requested that it be declared as `pub(crate)` instead so that it's not accidentally exposed, but doing so would add a lot of additional complexity if we still wanted that type to be publicly exposed when the feature flag is set.

src/share.rs
```diff
@@ -96,6 +96,19 @@ pub struct SendStream<B: IntoBuf> {
...
+///
+/// Streams are identified with an unsigned 31-bit integer. Streams
+/// initiated by a client MUST use odd-numbered stream identifiers; those
+/// initiated by the server MUST use even-numbered stream identifiers.  A
+/// stream identifier of zero (0x0) is used for connection control
+/// messages; the stream identifier of zero cannot be used to establish a
+/// new stream.
+///
+/// [Section 5.1.1]: https://tools.ietf.org/html/rfc7540#section-5.1.1
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
```
carllerche: I would recommend removing `Copy`. I'd also remove `Ord` and `PartialOrd` unless there is a case to keep them.

hawkw: The issue prompting this change (#289) states that:
> I'd expect `h2::StreamId` to implement at least copy and comparison traits.

I'm not sure whether or not @olix0r has any specific rationale or use-case behind implementing those traits, though?

carllerche: `Clone` would make more sense. `Copy` is pretty strong and adds a forwards compat hazard.

hawkw:
Build failure for 2d63ed47ac7c0d83e09a5c597f3522f7366b97b9 was due to a connection issue downloading `h2spec`, so I've restarted it.

PR #299, opened 2018-08-02
Fix the handling of incoming SETTINGS_INITIAL_WINDOW_SIZE.

goffrie:
From my understanding of the code and the HTTP/2 spec, we shouldn't try to reclaim `dec` from the `available` flow, but rather only take away any excess if it would exceed the new `window_size`.

This fixes a deadlock we observed: if we manage to send out a a new stream's HEADERS frame just before receiving a SETTINGS frame from the server that decreases the initial window size, `available` would be dropped down to zero and never requeued, causing us to never send `DATA` and then eventually time out. This is probably easy enough to write a test for but I don't have the time for it at the moment.

### DIFF
--- a/src/proto/streams/send.rs
+++ b/src/proto/streams/send.rs
@@ -11,7 +11,7 @@ use futures::{Async, Poll};
 use futures::task::Task;
 use tokio_io::AsyncWrite;
 
-use std::{cmp, io};
+use std::io;
 
 /// Manages state transitions related to outbound frames.
 #[derive(Debug)]
@@ -370,8 +370,8 @@ impl Send {
             self.init_window_sz = val;
 
             if val < old_val {
+                // We must decrease the (remote) window on every open stream.
                 let dec = old_val - val;
-
                 trace!("decrementing all windows; dec={}", dec);
 
                 let mut total_reclaimed = 0;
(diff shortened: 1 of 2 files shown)


carllerche:
Thanks for all the fixes 👍 I'm just doing a drive by comment and hoping that @seanmonstar or others can do a review.

Any chance for a test or thoughts on how to repro / add a test?

src/proto/streams/send.rs
```diff
@@ -380,15 +380,24 @@ impl Send {
 
                     stream.send_flow.dec_window(dec);
 
+                    let window_size = stream.send_flow.window_size();
                     let available = stream.send_flow.available().as_size();
-                    let reclaim = cmp::min(dec, available);
-                    stream.send_flow.claim_capacity(reclaim);
-                    total_reclaimed += reclaim;
+
+                    let reclaimed = if available > window_size {
```
seanmonstar: I admittedly no longer remember exactly what should be done. I wish we had let more comments relating back to the spec about what is done here. I know you're just fixing broken things, but since you're in here and seem to have more context, would it be possible to add a few lines of comments reminding Future Us what is supposed to happening? If not, not a blocker at all!

goffrie:
I added some comments and a test that hangs without this change.

PR #315, opened 2018-09-24
Update http-dependency to 0.1.13

theikkila:
Hello!

I'd like to update http (https://github.com/hyperium/http/blob/master/CHANGELOG.md) to latest version (0.1.13) because it contains fixes for bugs in urls that caused panics upstream.


carllerche:
Is there a reason this must be done in `h2?`

The dependency versions in `Cargo.toml` are intended to describe the *minimum* version required at an API level. However, Cargo will pull in the most up to date versions. So, you should be getting `http` 0.1.13 by default w/o needing to change `h2`.

theikkila:
yeah, you are right. Closing this! thx

PR #316, opened 2018-09-25
Update examples to use new Tokio

hawkw:
The old `tokio-core` crate is deprecated in favour of the `tokio` crate,
which also provides the new multithreaded Tokio runtime. Although `h2`
is generic over the runtime, the examples do depend on `tokio`. 

This branch update the example code to use the new `tokio` library
rather than `tokio-core`.

For the most part, this was pretty trivial --- simply replacing
`handle.spawn(...)` with `tokio::spawn(...)` and `core.run(...)` with
`tokio::run(...)`. There were a couple of cases where error and item
types from spawned futures had to be mapped to `(), ()`. Alternatively,
this could have been avoided by using the single-threaded runtime and
calling `block_on` instead to await the value of those futures, but I
thought it was better to reduce the amount of tokio-specific code in the
examples.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,7 +62,7 @@ serde = "1.0.0"
 serde_json = "1.0.0"
 
 # Akamai example
-tokio = "0.1"
+tokio = "0.1.8"
 env_logger = { version = "0.5.3", default-features = false }
 rustls = "0.12"
 tokio-rustls = "0.5.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,7 +62,7 @@ serde = "1.0.0"
 serde_json = "1.0.0"
 
 # Akamai example
-tokio-core = "0.1"
+tokio = "0.1"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -62,7 +62,7 @@ serde = "1.0.0"
 serde_json = "1.0.0"
 
 # Akamai example
-tokio-core = "0.1"
+tokio = "0.1"
```
carllerche: Would you mind just setting this to `0.1.8`? I'm trying to move towards minimal versions compiling. I'm not sure what the actual minimal version here that works is, but just going with `0.1.8` is fine.

hawkw: Done! 60d4d02bddd4b0db1b087e856d38224968482f75

hawkw:
Looks like the reason CI is failing is because there are some doctests for examples in the RustDoc that also use `tokio-core`, and this branch removed the dependency. I'll update those as well.

PR #319, opened 2018-10-11
Avoid prematurely unlinking streams in `send_reset`, in some cases.

goffrie:
Because `send_reset` called `recv_err`, which calls `reclaim_all_capacity`,
which eventually calls `transition(stream, ..)` -- all of which happens _before_
the RESET frame is enqueued -- it was possible for the stream to get unlinked
from the store (if there was any connection-level capacity to reassign). This
could then cause the stream to get "leaked" on drop/EOF since it would no longer
be iterated.

Fix this by delaying the call to `reclaim_all_capacity` _after_ enqueueing the
RESET frame.

A test demonstrating the issue is included.

### DIFF
--- a/src/proto/streams/send.rs
+++ b/src/proto/streams/send.rs
@@ -159,12 +159,17 @@ impl Send {
             return;
         }
 
-        self.recv_err(buffer, stream, counts);
+        // Clear all pending outbound frames.
+        // Note that we don't call `self.recv_err` because we want to enqueue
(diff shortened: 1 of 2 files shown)


carllerche:
Unfortunately, the build is broken due to an unrelated problem. I will work on fixing that.

carllerche:
I submitted #321 to fix the build.

PR #320, opened 2018-10-11
Reassign capacity from reset streams.

goffrie:
I believe this was an oversight - a stream that is reset can still have some
capacity assigned to it (e.g. if said capacity was assigned in the same poll as
the reset), which should be redistributed.

Note that `Send::recv_reset` is now identical to `Send::recv_err`; I'm not sure if they should be unified.

### DIFF
--- a/src/proto/streams/send.rs
+++ b/src/proto/streams/send.rs
@@ -320,17 +320,6 @@ impl Send {
         Ok(())
     }
 
-    pub fn recv_reset<B>(
-        &mut self,
-        buffer: &mut Buffer<Frame<B>>,
-        stream: &mut store::Ptr,
-        counts: &mut Counts,
-    ) {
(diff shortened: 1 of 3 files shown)


carllerche:
I submitted #321 to fix the build.

carllerche:
Yeah, it looks like these two functions *should* be unified (to `recv_err` as reset is how errors are provided).

carllerche:
Would you mind making that change in this PR? Thanks so much 👍

carllerche:
@goffrie I'll just make the change, I'm in the code anyway 👍 Thanks!

goffrie:
thanks! :+1:

carllerche:
@goffrie np, thanks for the QA / bug fixes. A release w/ all these fixes should be out shortly.

PR #321, opened 2018-10-16
Disable length_delimited deprecation warning.

carllerche:
Until tokio-rs/tokio#680 is resolved, we should allow using deprecated
APIs in the codec module.


carllerche:
r+? @seanmonstar @hawkw

carllerche:
@seanmonstar there are a lot of specific locations... this is a short term fix, so I"m not super worried about being more specific.

PR #324, opened 2018-10-16
Bump version to v0.1.13

carllerche:


src/lib.rs
```diff
@@ -84,7 +84,7 @@
 //! [`server::handshake`]: server/fn.handshake.html
 //! [`client::handshake`]: client/fn.handshake.html
 
-#![doc(html_root_url = "https://docs.rs/h2/0.1.12")]
+#![doc(html_root_url = "https://docs.rs/h2/0.1.13")]
```
seanmonstar: Care that this still points at docs.rs?

PR #327, opened 2018-10-21
Add server support for push

### LINKED ISSUE #185: Provide a way for the server to send push promises to the client
This will require an additional function on `server::Response`.

### LINKED ISSUE #291: State of HTTP/2 push
What is the state of HTTP/2 server push in h2? I couldn't find anything about it in the documentation. Are there any plans to implement it in the future?

michaelbeaumont:
Closes #291, closes #185


michaelbeaumont:
The test I added doesn't pass.
What happens:
```
.queue_frame(PushPromise(1, 2), .. , Stream(1) ..)
  Stream(1).pending_send.push_back(.., PushPromise(1,2))
  .schedule_send(Stream(1))
    .pending_send.push(Stream(1))
.queue_frame(PushPromise(1, 4), .. , Stream(1) ..)
  Stream(1).pending_send.push_back(.., PushPromise(1,4))
  .schedule_send(Stream(1))
    .pending_send.push(Stream(1))
.queue_frame(Headers(4), .. , Stream(4) ..)
  Stream(4).pending_send.push_back(.., Headers(4))
  .schedule_send(Stream(4))
    .pending_send.push(Stream(4))
```

`pending_send` doesn't add duplicates, so we have:
`pending_send: [1,4]`

in `pop_frame` `loop`:
```
1: .pending_send.pop() -> Stream(1)
      Stream(1).pending_send.pop_front -> PushPromise(1,2)
2: .pending_send.pop() -> Stream(2)
      Stream(1).pending_send.pop_front -> Headers(4)
```
So we write `[PushPromise(1,2), Headers(4)]` to our buffer.

When what we really want is `[PushPromise(1,2), PushPromise(1,4), Headers(4)]`

Obvious option would I guess be to only call `pop` from `Prioritize.pending_send` when there's no more `Frame`s left on `Stream.pending_send`, otherwise `peek`. Not sure what that effect that might have though.

carllerche:
@michaelbeaumont thanks for doing this <3 I just got home after traveling. I'm going to review this in more depth shortly.

Regarding the issue you described, it sounds tricky indeed.

The required behavior is to not send any frames for a push promise stream until the push promise is sent. It *sounds* like this behavior needs to be coded in.

One option would be to not schedule a push promise stream for send if the push promise frame has not been written. This would be tracked in the stream state.

So, if the headers frame is pushed, but the push promise was not written, then the stream is not scheduled for send. Once the push promise frame is written, at that point, the stream's pendign frames is checked to see if there is anything. If there is a frame, then the stream is scheduled for send at that point.

Thoughts?

src/server.rs
```diff
@@ -950,6 +949,22 @@ impl<B: IntoBuf> SendResponse<B> {
...
+
+    /// Push a request and response to the client
+    ///
+    /// On success, a [`SendResponse`] instance is returned.
+    ///
+    /// [`SendResponse`]: #
+    pub fn push_request(
+        &mut self,
+        request: Request<()>,
+    ) -> Result<SendResponse<B>, ::Error> {
```
carllerche: Since this returns a `SendResponse`, it looks like we will have a similar problem as with the client half.

Only "root" streams may have push promises. Calling `push_request` on the returned `SendResponse` should fail. Is there a test for this?

Or should this function return `SendPushedResponse`, a type similar to `SendResponse` but without a `push_request` function?

src/frame/headers.rs
```diff
@@ -283,9 +283,86 @@ impl fmt::Debug for Headers {
     }
 }
 
+// ===== util =====
+
+pub fn parse_u64(src: &[u8]) -> Result<u64, ()> {
```
carllerche: Since this is used in multiple locations now (I believe), I would consider creating a private `util` module at the root and moving this there.

src/frame/headers.rs
```diff
@@ -283,9 +283,86 @@ impl fmt::Debug for Headers {
...
+                fields,
+                is_over_size: false,
+                pseudo,
+            },
+            promised_id,
+            stream_id,
+        }
+    }
+
+    pub fn validate_request(req: &Request<()>) -> bool {
```
carllerche: This cleanup work is good standalone. It might be worth extracting it to a dedicated PR in order to simplify this PR (and make reviewing easier). Up to you though! Whatever makes it easier for you.

carllerche:
Unrelated, I'm thinking that, when writing frames, `prioritize` should keep writing frames from the same stream until after a data frame is reached...

src/server.rs
michaelbeaumont: My plan was definitely the `SendPushedResponse`! I think that's the cleanest option.

michaelbeaumont:
> So, if the headers frame is pushed, but the push promise was not written, then the stream is not scheduled for send. Once the push promise frame is written, at that point, the stream's pendign frames is checked to see if there is anything. If there is a frame, then the stream is scheduled for send at that point.

I've actually fixed the issue I mentioned in my first comment, in the second commit. By pushing the `PushPromise(1,4)` (which technically belongs to `Stream(1)`) onto `Stream(4)`'s `pending_send` (i.e. before `Headers(4)`) it's ensured that they don't get sent out of order.

But then there's unexpected behavior, as evidened by the test (which actually runs successfully, not sure why travis didn't build that commit).
https://github.com/carllerche/h2/blob/287c36bb547e20edb2c7f8e760d86956f8827a7e/tests/h2-tests/tests/server.rs#L145-L151

Although we send the headers for `Stream(4)` "first", the headers for `Stream(1)` are received first. Definitely not the best ordering for most use cases of push.

https://github.com/carllerche/h2/blob/287c36bb547e20edb2c7f8e760d86956f8827a7e/tests/h2-tests/tests/server.rs#L126-L129

> Unrelated, I'm thinking that, when writing frames, prioritize should keep writing frames from the same stream until after a data frame is reached...

Exactly, I think this is the crux of the issue. This seems the only way to enforce the ordering of frames like this and would fix the issue with the 2nd commit.

carllerche:
> I've actually fixed the issue I mentioned in my first comment, in the second commit. By pushing the PushPromise(1,4) (which technically belongs to Stream(1)) onto Stream(4)'s pending_send (i.e. before Headers(4)) it's ensured that they don't get sent out of order.

It seems to me that it would be possible for `Stream(1)` to be closed before the push promise is sent out? The order in which frames are sent out for any given stream should not be assumed. The prioritization logic is free to change.

michaelbeaumont:
@carllerche  do you have time to take another look?

src/frame/headers.rs
```diff
@@ -283,9 +283,86 @@ impl fmt::Debug for Headers {
...
+        // The spec has some requirements for promised request headers
+        // [https://httpwg.org/specs/rfc7540.html#PushRequests]
+
+        // A promised request "that indicates the presence of a request body
+        // MUST reset the promised stream with a stream error"
+        if let Some(content_length) = req.headers().get(header::CONTENT_LENGTH) {
+            match parse_u64(content_length.as_bytes()) {
```
hawkw: Nit: it seems like this can just be
```rust
if parse_u64(content_length.as_bytes()) != Ok(0) {
    return false;
}
```
which IMO is a little clearer.

src/server.rs
```diff
@@ -284,6 +283,31 @@ pub struct SendResponse<B: IntoBuf> {
...
+/// the HTTP/2.0 stream will be reset.
+///
+/// See [module] level docs for more details.
+///
+/// [module]: index.html
+pub struct SendPushedResponse<B: IntoBuf> {
+    inner: SendResponse<B>,
+}
+
+impl<B: IntoBuf + fmt::Debug> fmt::Debug for SendPushedResponse<B>
```
hawkw: If there's a reason this can't be derived, a comment explaining that would be good...

src/proto/streams/streams.rs
```diff
@@ -866,6 +866,67 @@ impl<B> StreamRef<B> {
...
+        request: Request<()>,
+    ) -> Result<StreamRef<B>, UserError> {
+        let mut me = self.opaque.inner.lock().unwrap();
+        let me = &mut *me;
+
+        let mut send_buffer = self.send_buffer.inner.lock().unwrap();
+        let send_buffer = &mut *send_buffer;
+
+        let promised_id = {
+            let actions = &mut me.actions;
```
hawkw: this seems weird --- is the subscope really necessary? it seems like you should be able to just say
```rust
let send_buffer = ...

let actions = &mut me.actions;
actions.send.reserve_local()?;

let child_key = ...
```
unless I'm missing something?

src/server.rs
michaelbeaumont: Yep, and I'm really not sure that's the best way to solve the issue.
It fails because `#[derive(Debug)]` on `SendResponse` requires `<B as bytes::IntoBuf>::Buf: fmt::Debug`

I _think_ this is because of https://github.com/rust-lang/rust/issues/26925

Actually `#[derive(Debug)]` introduces these too strict bounds from `SendResponse` through `StreamRef`, `SendBuffer`, `Buffer` and `Slab`

Another option would be to include the `fmt::Debug` bound on the `SendPushedResponse` type itself.

Any thoughts?

src/proto/streams/streams.rs
michaelbeaumont: probably was necessary at some point, fixed it

src/frame/headers.rs
michaelbeaumont: fixed

src/server.rs
hawkw: In that case, I think the manual impl is correct. Let's leave it as-is, and add a comment stating why it can't be derived so that this is documented for future readers.

ivanceras:
I need to use HTTP/2 server push. How much of the work is done for a bare minimum functional server push? Is resolving the conflicts here still applicable?

le-invoker:
Any progress on this? I also needed the HTTP/2 server_push extension. What's good for an HTTP/2 without support for the server_push extension.

michaelbeaumont:
Tests still need to be rewritten, I'll work on it

AlexandroPixel22:
Any news on this? I think thousands of people are expecting this crate to finally work since months but still nothing :/ Please anyone fork it and make it work, because HTTP2 is only useful because of SERVERPUSH... Save the rust community please

LimitlessCloud:
> Any news on this? I think thousands of people are expecting this crate to finally work since months but still nothing :/ Please anyone fork it and make it work, because HTTP2 is only useful because of SERVERPUSH... Save the rust community please

It's kinda mean to say that this is useless with this option...
It looks like the developer just abandoned the library that's it, I might work on a fork guys to help you, might be done in a few weeks.

seanmonstar:
Looks like there's some merge conflicts?

michaelbeaumont:
@seanmonstar The branch is up to date as far as I can tell

seanmonstar:
itshappening.gif

NN-Binary:
Wow amazing job guys! Congrats, you are the first Rust crate to handle HTTP2 Server push!

le-invoker:
Thank you all for the hardwork. 
Now, if someone can pinpoint me the exact function to use for me to tell the webserver to create a `server_push` response for each corresponding future `request`.

le-invoker:
hi @michaelbeaumont , there is a posted [example/issue](https://github.com/hyperium/h2/issues/412) on how to use h2 to do server_push. I have the same problem, can you pinpoint what the code does wrong here?

PR #333, opened 2018-11-29
Avoid locking when printing

carllerche:
It is not obvious that attempting to print the
value of a struct could cause a deadlock. To avoid
this, this patch does not lock the mutex when generating
a debug representation of the h2 struct.


seanmonstar:
If this debug output is useful, there's `Mutex::try_lock`...

carllerche:
Updated

carllerche:
@seanmonstar you good w/ this now?

PR #334, opened 2018-12-05
Release closed streams capacity back to connection

seanmonstar:
Previously, any streams that were dropped or closed while not having
consumed the inflight received window capacity would simply leak that
capacity for the connection. This could easily happen if a `RecvStream`
were dropped before fully consuming the data, and therefore a user would
have no idea how much capacity to release in the first place. This
resulted in stalled connections that would never have capacity again.

### DIFF
--- a/src/proto/streams/recv.rs
+++ b/src/proto/streams/recv.rs
@@ -334,7 +334,11 @@ impl Recv {
         capacity: WindowSize,
         task: &mut Option<Task>,
     ) {
-        trace!("release_connection_capacity; size={}", capacity);
+        trace!(
+            "release_connection_capacity; size={}, connection in_flight_data={}",
+            capacity,
+            self.in_flight_data,
+        );
 
         // Decrement in-flight data
         self.in_flight_data -= capacity;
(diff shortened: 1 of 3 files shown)


src/proto/streams/recv.rs
```diff
@@ -544,6 +567,12 @@ impl Recv {
             }
         }
 
+        // Update stream level flow control
+        stream.recv_flow.send_data(sz);
+
+        // Track the data as in-flight
+        stream.in_flight_recv_data += sz;
+
```
olix0r: just for my own understanding, was this change required subtly? It seems that the effect is just that we don't update the stream state when the stream protocol errors, which seems fine. Anything else here?

seanmonstar: Yes it was. Without the change, another test was failing (yay having tests!). Before the change, the size was added to the stream's `in_flight_recv_data`, but then the returned error already returns the window size for this data frame to the connection: https://github.com/carllerche/h2/blob/master/src/proto/streams/streams.rs#L240-L247

In this PR, with the `in_flight_recv_data` being returned to the connection when `ref_count` becomes zero meant there was window capacity returned twice, causing an underflow.

PR #346, opened 2019-02-13
Add user PING support

seanmonstar:
- Add `share::PingPong`, which can send `Ping`s, and poll for the `Pong`
  from the peer.


src/share.rs
```diff
@@ -193,6 +195,22 @@ pub struct ReleaseCapacity {
     inner: proto::OpaqueStreamRef,
 }
 
+/// dox
```
seanmonstar: Will write up docs, just wanted to get the API in front of more eyes.

seanmonstar:
cc @olix0r

src/proto/ping_pong.rs
```diff
@@ -29,14 +41,27 @@ pub(crate) enum ReceivedPing {
...
         PingPong {
             pending_ping: None,
             pending_pong: None,
+            user_pings: None,
         }
     }
 
-    pub fn ping_shutdown(&mut self) {
+    pub(crate) fn take_user_pings(&mut self) -> Arc<UserPings> {
+        assert!(self.user_pings.is_none(), "user ping pong taken more than once");
```
olix0r: Why not just return the cloned `self.user_pings` value in this case?  In general, the "take" part of this isn't really clear to me, since it seems that UserPings is intended to be shareable?  Or if it's not intended to be shareable, perhaps we should return `Result` types so that this contract is clearer to callers?

carllerche: It is `spsc` due to the task.

seanmonstar: The `PingPong` type can only be taken once (otherwise calling `poll_pong` on clones would result in stalled tasks). This was "just" an internal method to get the type out of the `proto` module and into `share`. But still, it was a little confusing, so I've tried to make it clearer:

- Some comments here that its a 1-time thing.
- Returns an `Option` instead of panicking, panic is now up to the user.

Without a way to check if some other possibly unrelated code had already taken the `PingPong`, there was no way to protect yourself against a panic. Now, you can optionally use `PingPong`, or choose to panic anyways.

seanmonstar:
I realized there were a couple bugs with just using 2 `AtomicBool`s, so I've switched them out for a single `AtomicUsize` as state. This deals with some edge cases, like receiving a `PING` but the user had never actually sent one.

Also wrote up some docs.

PR #356, opened 2019-05-15
Header encoding fixes

seanmonstar:
- Fixes panic if a "resume" had to pause at a header with a repeat name.
- Fixes erroneous name index usage when repeated header used an index of a header that had evicted itself.

### DIFF
--- a/src/hpack/encoder.rs
+++ b/src/hpack/encoder.rs
@@ -133,10 +133,13 @@ impl Encoder {
                 // which case, we skip table lookup and just use the same index
                 // as the previous entry.
                 Err(value) => {
-                    let index = last_index.as_ref().unwrap_or_else(|| {
-                        panic!("encoding header without name, but not previous index to use for name");
-                    });
-                    let res = self.encode_header_without_name(index, &value, dst);
+                    let res = self.encode_header_without_name(
+                        last_index.as_ref().unwrap_or_else(|| {
+                            panic!("encoding header without name, but no previous index to use for name");
+                        }),
+                        &value,
+                        dst,
+                    );
 
                     if res.is_err() {
                         dst.truncate(len);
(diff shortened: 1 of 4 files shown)


src/hpack/encoder.rs
```diff
@@ -103,10 +105,9 @@ impl Encoder {
                 dst.truncate(len);
                 return Encode::Partial(resume);
             }
+            last_index = Some(resume.index);
```
seanmonstar: This fixes the first issue.

src/hpack/table.rs
```diff
@@ -125,7 +125,7 @@ impl Table {
             Indexed(idx, ..) => idx,
             Name(idx, ..) => idx,
             Inserted(idx) => idx + DYN_OFFSET,
-            InsertedValue(idx, _) => idx,
+            InsertedValue(_name_idx, slot_idx) => slot_idx + DYN_OFFSET,
```
seanmonstar: This makes the next header (that has no name, so using this index) point at the new index, not at the old name index, which may have been evicted.

src/hpack/encoder.rs
```diff
@@ -443,6 +446,23 @@ mod test {
...
+        let res = encode(&mut encoder, vec![header("content-language", "foo")]);
+
+        assert_eq!(res[0], 0b01000000 | 27); // Indexed name
+        assert_eq!(res[1], 0x80 | 2); // header value w/ huffman coding
+
+        assert_eq!("foo", huff_decode(&res[2..4]));
+
+        // Same name, new value should still use incremental
+        let res = encode(&mut encoder, vec![header("content-language", "bar")]);
+        assert_eq!(res[0], 0b01000000 | 27); // Indexed name
```
seanmonstar: The seed I was using to replay the fuzz had this indexed name, and I noticed that new values were pointing at the dynamic index when it could be pointing at the static index. The HPACK spec has [an example](https://httpwg.org/specs/rfc7541.html#n-third-response_1) doing exactly this (well, the `date` header...).

olix0r:
fyi, ci failure:

```
error[E0505]: cannot move out of `last_index` because it is borrowed
   --> src/hpack/encoder.rs:144:36
    |
136 |                     let index = last_index.as_ref().unwrap_or_else(|| {
    |                                 ---------- borrow of `last_index` occurs here
...
144 |                             index: last_index.unwrap(), // checked just above
    |                                    ^^^^^^^^^^ move out of `last_index` occurs here
error: aborting due to previous error
```

PR #360, opened 2019-05-29
Fix write loop when a header is too big to encode

seanmonstar:
If any single header value was bigger than the write buffer, we'd write a CONTINUATION frame, and then get ready to write the header after. But since the write buffer doesn't grow in size, we'd loop on that condition forever.

- The immediate fix checks if the `CONTINUATION` frame that was written is empty. Instead of looping forever, this will panic, as it's now considered an internal bug.
- So as to not reach the panic, `HEADERS` frames are now checked that they don't contain a header that would be larger than the write buffer.
- Lastly, since this was found with a 6000 length header, the write buffer size has been increased from 4kb to 16kb.

### DIFF
--- a/src/codec/framed_write.rs
+++ b/src/codec/framed_write.rs
@@ -197,6 +197,17 @@ where
                 Some(Next::Continuation(frame)) => {
                     // Buffer the continuation frame, then try to write again
                     if let Some(continuation) = frame.encode(&mut self.hpack, self.buf.get_mut()) {
+
+                        // We previously had a CONTINUATION, and after encoding
+                        // it, we got *another* one? Let's just double check
+                        // that at least some progress is being made...
+                        if self.buf.get_ref().len() == frame::HEADER_LEN {
+                            // If *only* the CONTINUATION frame header was
+                            // written, and *no* header fields, we're stuck
+                            // in a loop...
+                            panic!("CONTINUATION frame write loop; header value too big to encode");
+                        }
+
                         self.next = Some(Next::Continuation(continuation));
                     }
                 },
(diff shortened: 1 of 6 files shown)


src/codec/framed_write.rs
```diff
@@ -197,6 +200,17 @@ where
...
                     if let Some(continuation) = frame.encode(&mut self.hpack, self.buf.get_mut()) {
+
+                        // We previously had a CONTINUATION, and after encoding
+                        // it, we got *another* one? Let's just double check
+                        // that at least some progress is being made...
+                        if self.buf.get_ref().len() == frame::HEADER_LEN {
+                            // If *only* the CONTINUATION frame header was
+                            // written, and *no* header fields, we're stuck
+                            // in a loop...
+                            panic!("CONTINUATION frame write loop; header value too big to encode");
```
hawkw: Since the expectation is that the check in `Send` should return an error before we ever hit this, I think this should probably be an `unreachable!` and/or make it clear that this is an internal implementation bug?

src/frame/headers.rs
```diff
@@ -844,6 +852,46 @@ impl HeaderBlock {
             .map(|(name, value)| decoded_header_size(name.as_str().len(), value.len()))
             .sum::<usize>()
     }
+
+    /// Iterator all pseudos and headers to see if any individual pair
```
hawkw: Nit: I think there's a typo here:
```suggestion
    /// Iterate over all pseudos and headers to see if any individual pair
```

src/codec/framed_write.rs
seanmonstar: >  make it clear that this is an internal implementation bug

It seems clear to me that it's a bug, I'm not 100% certain it's completely unreachable...

PR #365, opened 2019-05-31
Refactor proto::streams::store indices

seanmonstar:
- Removes incrementing counter, instead just using the StreamId as a
  slab ABA guard.
- Adjusts Ptr::deref to use Store::index, as before it was skipping to
  check the ABA guard.
- Rename fields and types to clarify it actually is an ABA guard.
- Improve panic message in case a dangling Ptr is accessed.

### DIFF
--- a/src/proto/streams/store.rs
+++ b/src/proto/streams/store.rs
@@ -11,9 +11,8 @@ use std::ops;
 /// Storage for streams
 #[derive(Debug)]
 pub(super) struct Store {
-    slab: slab::Slab<(StoreId, Stream)>,
-    ids: IndexMap<StreamId, (usize, StoreId)>,
-    counter: StoreId,
+    slab: slab::Slab<Stream>,
+    ids: IndexMap<StreamId, SlabIndex>,
 }
 
 /// "Pointer" to an entry in the store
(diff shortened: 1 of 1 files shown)


src/proto/streams/store.rs
```diff
@@ -178,17 +178,23 @@ impl ops::Index<Key> for Store {
...
 
     fn index(&self, key: Key) -> &Self::Output {
-        let slot = self.slab.index(key.index);
-        assert_eq!(slot.0, key.store_id);
-        &slot.1
+        self.slab
+            .get(key.index.0)
+            .filter(|s| s.id == key.stream_id)
+            .unwrap_or_else(|| {
+                panic!("dangling store key for stream_id={:?}", key.stream_id);
```
hawkw: 👍 this is significantly clearer than the old assert

src/proto/streams/store.rs
```diff
@@ -82,35 +83,32 @@ impl Store {
...
             ids: IndexMap::new(),
-            counter: 0,
         }
     }
 
     pub fn find_mut(&mut self, id: &StreamId) -> Option<Ptr> {
-        let key = match self.ids.get(id) {
+        let index = match self.ids.get(id) {
             Some(key) => *key,
             None => return None,
```
hawkw: This is unrelated, but while we're refactoring, this can probably just use `?`.

src/proto/streams/store.rs
```diff
@@ -82,35 +83,32 @@ impl Store {
...
-        let key = self.slab.insert((store_id, val));
-        assert!(self.ids.insert(id, (key, store_id)).is_none());
+        let idx = SlabIndex(self.slab.insert(val));
+        assert!(self.ids.insert(id, idx).is_none());
 
         Ptr {
             key: Key {
-                index: key,
-                store_id,
+                index: idx,
```
hawkw: TIOLI: why not just call the local var `index` so this can use local var shorthand?

src/proto/streams/store.rs
```diff
@@ -140,12 +137,15 @@ impl Store {
 
         while i < len {
             // Get the key by index, this makes the borrow checker happy
-            let key = *self.ids.get_index(i).unwrap().1;
+            let (stream_id, index) = {
```
hawkw: Nit: I think this can be simplified to just
```rust
let (&stream_id, &index) = self.ids.get_index(i).unwrap();
```

seanmonstar: Does that require a more recent compiler? I worried the lifetime would cause a conflict with passing `self` into `f()`...

PR #368, opened 2019-06-04
Clear recv buffer when stream refs are dropped

### LINKED ISSUE #367: Server memory leak
A memory leak exists in the current `h2` server implementation. 

It can be be reproduced quite simply by running the `server` example, pointing an HTTP/2 load generator at it, and observing the server process's resident set size increase over time. When the request payload is larger, the server's RSS grows much more rapidly.

For example,
```
$ cargo run --example server &
...
$ h2load --requests=10000 --clients=1 --threads=1 http://localhost:5928 -d target/debug/examples/server
```

Since larger request payloads cause the server to leak memory faster, I suspect this issue is related to payload buffer management in some way.

seanmonstar:
Closes #367

### DIFF
--- a/src/share.rs
+++ b/src/share.rs
@@ -448,6 +448,17 @@ impl fmt::Debug for RecvStream {
     }
 }
 
+impl Drop for RecvStream {
+    fn drop(&mut self) {
+        // Eagerly clear any received DATA frames now, since its no longer
+        // possible to retrieve them. However, this will be called
+        // again once *all* stream refs have been dropped, since
+        // this won't send a RST_STREAM frame, in case the user wishes to
+        // still *send* DATA.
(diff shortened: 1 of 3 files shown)


src/share.rs
```diff
@@ -448,6 +448,17 @@ impl fmt::Debug for RecvStream {
     }
 }
 
+impl Drop for RecvStream {
+    fn drop(&mut self) {
+        // Eagerly clear any received DATA frames now, since its no longer
+        // possible to retrieve them. However, this will be called
```
olix0r: is _this_ `drop` or `clear_recv_buffer`? I assume the latter, but it's not clear

seanmonstar: The call to `clear_recv_buffer`.

PR #371, opened 2019-06-14
Log protocol errors

hawkw:
Currently, there are many cases where `h2` will fail a connection or
stream with a PROTOCOL_ERROR, without recording why the protocol error
occurred. Since protocol errors may result from a bug in `h2` or from a
misbehaving peer, it is important to be able to debug the cause of
protocol errors.

This branch adds a log line to almost all cases where a protocol error
occurs. I've tried to make the new log lines consistent with the
existing logging, and in some cases, changed existing log lines to make
them internally consistant with other log lines in that module. I have
_not_ tried to enforce a single style for log lines across all modules,
however, as the debug-level in some modules uses a different style than
the trace logging elsewhere. Presumably this is because it is expected
to be more user-facing.

### DIFF
--- a/src/codec/framed_read.rs
+++ b/src/codec/framed_read.rs
@@ -63,7 +63,7 @@ impl<T> FramedRead<T> {
         let head = frame::Head::parse(&bytes);
 
         if self.partial.is_some() && head.kind() != Kind::Continuation {
-            trace!("connection error PROTOCOL_ERROR -- expected CONTINUATION, got {:?}", head.kind());
+            proto_err!(conn: "expected CONTINUATION, got {:?}", head.kind());
             return Err(Connection(Reason::PROTOCOL_ERROR));
         }
 
@@ -81,7 +81,7 @@ impl<T> FramedRead<T> {
                 let (mut frame, mut payload) = match frame::$frame::load($head, $bytes) {
                     Ok(res) => res,
                     Err(frame::Error::InvalidDependencyId) => {
-                        debug!("stream error PROTOCOL_ERROR -- invalid HEADERS dependency ID");
+                        proto_err!(stream: "invalid HEADERS dependency ID");
                         // A stream cannot depend on itself. An endpoint MUST
                         // treat this as a stream error (Section 5.4.2) of type
                         // `PROTOCOL_ERROR`.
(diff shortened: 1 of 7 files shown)


hawkw:
@olix0r I used `warn` because I thought a situation likely resulting from a misbehaving client ought to result in a warning, but upon further consideration, those logs probably ought to be at the `trace` level, since almost all logging in `h2` is at `trace`. I used `debug` in `framed_read` since that module was already logging protocol errors at the `debug` level.

seanmonstar:
Yea, I'd put logs about errors at debug level.

hawkw:
Okay, will do!

src/server.rs
```diff
@@ -1268,9 +1270,10 @@ impl proto::Peer for Peer {
 
         let mut request = match b.body(()) {
             Ok(request) => request,
-            Err(_) => {
+            Err(e) => {
                 // TODO: Should there be more specialized handling for different
                 // kinds of errors
+                trace!("request body error: {}, PROTOCOL_ERROR", e);
```
olix0r: note that this isn't necessarily a _body errror_ -- it's any error building the request.

hawkw: ah, you're right, will rephrase.

hawkw:
I've updated this branch so that all protocol errors are logged at the debug level, and changed all the log lines to use a consistent formatting based on the protocol error log lines in `framed_read`, using a helper macro to reduce boilerplate.

src/codec/framed_read.rs
```diff
@@ -171,18 +171,25 @@ impl<T> FramedRead<T> {
             },
             Kind::Reset => {
                 let res = frame::Reset::load(head, &bytes[frame::HEADER_LEN..]);
-                res.map_err(|_| Connection(Reason::PROTOCOL_ERROR))?.into()
+                res.map_err(|e| {
+                    debug!("connection error PROTOCOL_ERROR -- failed to load RESET frame; err={:?}", e);
```
olix0r: should these use `proto_err!` as well?

src/proto/streams/recv.rs
```diff
@@ -586,6 +597,11 @@ impl Recv {
 
     pub fn consume_connection_window(&mut self, sz: WindowSize) -> Result<(), RecvError> {
         if self.flow.window_size() < sz {
+            trace!(
```
olix0r: Should these other error logs be emitted at debug!? it seems any time where the server fails a stream without the application taking action, we should at least log at debug...

hawkw: We can log these at debug as well --- I was scoping this change to protocol errors specifically, but I can change these to debug as well.

PR #372, opened 2019-06-17
allow servers to receive requests without an :authority header

### LINKED ISSUE #345: Requests missing an :authority pseudo-header return PROTOCOL_ERROR
We are using nghttpx to upgrade requests from HTTP to H2. nghttpx omits the `:authority` pseudo-header when creating the request to forward on via H2, this appears to be consistent with [the spec](https://httpwg.org/specs/rfc7540.html#rfc.section.8.1.2.3):

> The :authority pseudo-header field includes the authority portion of the target URI ([RFC3986], Section 3.2). The authority MUST NOT include the deprecated userinfo subcomponent for http or https schemed URIs.

> To ensure that the HTTP/1.1 request line can be reproduced accurately, this pseudo-header field MUST be omitted when translating from an HTTP/1.1 request that has a request target in origin or asterisk form (see [RFC7230], Section 5.3). Clients that generate HTTP/2 requests directly SHOULD use the :authority pseudo-header field instead of the Host header field. An intermediary that converts an HTTP/2 request to HTTP/1.1 MUST create a Host header field if one is not present in a request by copying the value of the :authority pseudo-header field. 

However, it appears that h2 incorrectly rejects the request with `PROTOCOL_ERROR` when this happens. It also seems like there should also be some logging around this code path, but I couldn't get it to trigger.

seanmonstar:
The code before *tried* to allow these requests, but then the `request::Builder` would error trying to construct a `Uri` with only the scheme and path. Now, if there is no authority, the scheme is dropped.

Closes #345

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1236,20 +1236,32 @@ impl proto::Peer for Peer {
         // Convert the URI
         let mut parts = uri::Parts::default();
 
-        if let Some(scheme) = pseudo.scheme {
-            let maybe_scheme = uri::Scheme::from_shared(scheme.clone().into_inner());
-            parts.scheme = Some(maybe_scheme.or_else(|why| malformed!(
-                "malformed headers: malformed scheme ({:?}): {}", scheme, why,
-            ))?);
-        } else {
-            malformed!("malformed headers: missing scheme");
-        }
 
+        // A request translated from HTTP/1 must not include the :authority
+        // header
         if let Some(authority) = pseudo.authority {
             let maybe_authority = uri::Authority::from_shared(authority.clone().into_inner());
             parts.authority = Some(maybe_authority.or_else(|why| malformed!(
                 "malformed headers: malformed authority ({:?}): {}", authority, why,
             ))?);
+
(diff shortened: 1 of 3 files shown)


src/server.rs
```diff
@@ -1236,20 +1236,32 @@ impl proto::Peer for Peer {
...
-            let maybe_scheme = uri::Scheme::from_shared(scheme.clone().into_inner());
-            parts.scheme = Some(maybe_scheme.or_else(|why| malformed!(
-                "malformed headers: malformed scheme ({:?}): {}", scheme, why,
-            ))?);
-        } else {
-            malformed!("malformed headers: missing scheme");
-        }
 
+        // A request translated from HTTP/1 must not include the :authority
+        // header
```
olix0r: If a `host` header is present do we have any obligation to validate it?

seanmonstar: I can't find anything in the [HTTP2 spec](https://httpwg.org/specs/rfc7540.html) stating to do anything with a received `host` header besides handling it like any other.

PR #374, opened 2019-06-17
Update listed version numbers

atouchet:


atouchet:
I just noticed that the

[dependencies]
h2 = "0.1.15"

section hasn't been updated either. What should that be changed to?

seanmonstar:
That can also just be "0.1".

PR #375, opened 2019-06-24
Send RST_STREAM of STREAM_CLOSED instead of GOAWAY

seanmonstar:
In some cases, we may cancel and/or reset a stream, and then proceed to forget it from our store. Normally, we'll add it to an expiring queue to allow any frames that were in-transit to be silently ignored, as per the spec. However, in some cases, such as if many streams are reset, some may end up forgotten before the peer could have reacted. In those cases, we would send a GOAWAY, tearing down the entire connection.

Inspired by other HTTP2 implementations, if we receive a frame for a stream that we *might* have seen before, we now send a RST_STREAM of STREAM_CLOSED, and no longer trash the connection. While this might be more permissive of actual bad peers, in practice this should mostly protect from latency/memory limits that cause the connection to die.

### DIFF
--- a/src/proto/streams/recv.rs
+++ b/src/proto/streams/recv.rs
@@ -516,27 +516,18 @@ impl Recv {
             stream.recv_flow.window_size()
         );
 
-        // Ensure that there is enough capacity on the connection before acting
-        // on the stream.
-        self.consume_connection_window(sz)?;
 
         if is_ignoring_frame {
             trace!(
                 "recv_data; frame ignored on locally reset {:?} for some time",
                 stream.id,
             );
-            // we just checked for enough connection window capacity, and
-            // consumed it. Since we are ignoring this frame "for some time",
-            // we aren't returning the frame to the user. That means they
-            // have no way to release the capacity back to the connection. So
-            // we have to release it automatically.
-            //
-            // This call doesn't send a WINDOW_UPDATE immediately, just marks
-            // the capacity as available to be reclaimed. When the available
-            // capacity meets a threshold, a WINDOW_UPDATE is then sent.
-            self.release_connection_capacity(sz, &mut None);
-            return Ok(());
+            return self.ignore_data(sz);
         }
+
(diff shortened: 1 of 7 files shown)


seanmonstar:
Hrm, so now test explicitly checking that sending on an old stream ID is a protocol error is failing. That's because this change explicitly made that a stream error instead... I guess modify the test?

src/proto/streams/recv.rs
```diff
@@ -599,6 +590,22 @@ impl Recv {
         Ok(())
     }
 
+    pub fn ignore_data(&mut self, sz: WindowSize) -> Result<(), RecvError> {
+        // Ensure that there is enough capacity on the connection...
+        self.consume_connection_window(sz)?;
```
olix0r: Is the thought that this window underflow is important to surface, even in cases when we are going to totally ignore the frame?  I haven't thought deeply, but if we can't trust the client to understand the state of these streams, might it better to just ignore the window in this case?

seanmonstar: This is being lenient over whether the peer has received our `RST_STREAM` frame yet. I'd still expect a peer to respect the connection flow control.

seanmonstar:
Now `h2spec` is complaining that we should sent a GOAWAY... Hm...

carllerche:
@seanmonstar didn't we discuss at some point adding a PING to track when the peer should know the stream is no longer valid?

Also... strict mode? :'(

seanmonstar:
I don't remember a ping... We use that for GOAWAY... But yea, this breaks strict mode and h2spec.

Looking at other implementations: Netty just ignores that test case. Golang enforces it on the server, and ignores it on the client. Nghttp2 does similar.

An alternative to this I'm going to look into is if we just need to improve our cancellation stuff, and/or if we should just increase the retained reset streams limit from 10 to something a little bigger...
-

seanmonstar:
Trying the alternatives for now.

PR #378, opened 2019-06-26
tests: reduce boilerplate of sending GET requests

seanmonstar:
This adds a `SendRequestExt` trait to h2-support, with a `get` method
that does a lot of the repeated request building stuff many test cases
were doing.

As a first step, the cleans up stream_states tests to use it.


tests/h2-support/src/future_ext.rs
```diff
@@ -72,20 +72,11 @@ pub trait FutureExt: Future {
     ///
     /// This allows the executor to poll other futures before trying this one
     /// again.
-    fn yield_once(self) -> Box<Future<Item = Self::Item, Error = Self::Error>>
+    fn yield_once(self) -> Box<dyn Future<Item = Self::Item, Error = Self::Error>>
```
hawkw: Do we care about supporting Rust versions that don't have the `dyn` keyword (such as 1.26.0?)

tests/h2-tests/tests/stream_states.rs
```diff
@@ -202,16 +202,9 @@ fn errors_if_recv_frame_exceeds_max_frame_size() {
...
-            .uri("https://example.com/")
-            .body(())
-            .unwrap();
-
         let req = client
-            .send_request(request, true)
-            .unwrap()
-            .0.unwrap()
+            .get("https://example.com/")
+            .expect("response")
```
hawkw: <3

tests/h2-tests/tests/stream_states.rs
```diff
@@ -829,7 +753,7 @@ fn rst_while_closing() {
         .recv_settings()
         .recv_frame(
             frames::headers(1)
-                .request("GET", "https://example.com/")
+                .request("POST", "https://example.com/")
```
hawkw: Is the change from GET to POST here related to the rest of this change?

tests/h2-support/src/future_ext.rs
seanmonstar: No, it's too much of a pain. hyper already uses this, so onward!

tests/h2-tests/tests/stream_states.rs
seanmonstar: I was searching for GET, and then part way into replacing this, realized it wants the `SendStream`, so POST seemed clearer.

PR #379, opened 2019-06-26
Clear pending_reset_expiration queue if receiving an EOS frame

seanmonstar:
The `pending_reset_expiration` queue keeps locally reset streams around
for "some time" to allow us to ignore frames on those streams, in case
the remote sent them before receiving our RST_STREAM. If we receive a
HEADERS or DATA frame for one of those streams, while still ignoring it,
if the frame has the EOS flag, we can now assume the remote won't send
any more frames on that stream, and can cleanup the queue.

This helps reduce the likelihood of reaching the "max concurrent reset
streams" limit.

### DIFF
--- a/src/proto/streams/recv.rs
+++ b/src/proto/streams/recv.rs
@@ -334,6 +334,23 @@ impl Recv {
         Ok(())
     }
 
+    /// Transition the stream even though we're ignoring frames "for some time".
+    pub fn recv_ignored_trailers(
+        &mut self,
+        frame: frame::Headers,
+        stream: &mut store::Ptr,
+    ) {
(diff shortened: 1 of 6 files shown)


seanmonstar:
I've re-opened https://github.com/hyperium/h2/pull/380, which *might* mean this isn't needed... Though, this does separately just help with *remembering* only the reset streams that are still relevant...

carllerche:
What is the status of this?

seanmonstar:
I'm undecided if this should be merged. We already fixed the original bug, this is just a possible optimization. It could reduce the memory usage of keeping track of unacknowledged reset streams.

PR #380, opened 2019-06-27
Send RST_STREAM instead of GOAWAY if stream may have been forgotten

seanmonstar:
Re-opening of #375, with an adjustment:

The forgotten check for HEADERS is only done on the client side. The rationale is that the *client* can send a request, and then later a RST_STREAM and then forget about the stream while the server HEADERS were in transit. A *server* however will **never** reset a stream before receiving the request HEADERS.

This distinction is also done in golang-http2, grpc-go, and nghttp2. I'm hopeful this makes h2spec still pass.


seanmonstar:
Yep, h2spec is part of CI, and it's green!

PR #389, opened 2019-07-27
Update to std::future

### LINKED ISSUE #385: Update to std::future

95th:
Closes #385


95th:
For some reason, CI is using old version of Rust Nightly. Multiple lifetime issue has been resolved a month now. @seanmonstar How do I make it use latest Nightly?

seanmonstar:
I'm on vacation for a couple weeks with no computer and spotty internet. If no one else looks while I'm gone, I'll take a look when I get back.

Ping @carllerche @sgrif

carllerche:
And @sgrif is sick. I def want to pull in one of the seans...

95th:
@carllerche Okay.. Meanwhile I'll finish up what's remaining. Current status:

Works:
- Code builds
- Internal h2 crate tests pass
- h2spec passes
- h2-test compiles

In progress:
- h2-tests fail.
- doc tests needs to be updated - I am thinking of doing it in a separate PR.
- h2-fuzz - needs to updated - will update it after fixing the h2-tests

95th:
Since integrated CI doesnt use correct rust version, here is a build run against my fork:

https://travis-ci.org/95th/h2/builds/567710136

95th:
@carllerche
Most of the failing tests have similar errors:
```
expected SETTINGS; actual=Headers { stream_id: StreamId(1), flags: (0x5: END_HEADERS | END_STREAM) }
expected SETTINGS; actual=Headers { stream_id: StreamId(1), flags: (0x4: END_HEADERS) }
```

The error comes in `assert_client_handshake` when trying to assert settings ACK.

Any ideas?

cc @LucioFranco @sgrif

src/client.rs
```diff
@@ -64,7 +64,7 @@
 //!
 //! # Example
 //!
-//! ```rust
+//! ```rust, no_run
```
95th: Previously, any panics inside `tokio::run` didn't fail the test. Now it does. So, make this compile only.

sgrif:
Is there any way for you to break this up into smaller pieces? As I've been working through this my plan was to start with just `h2-support` and the tests, since those can be changed without touching the main crate, while the inverse is not true. 6k lines changed is a lot to review at once

95th:
Looks like my own PR https://github.com/rust-lang/rust/issues/63514 has come back to bite me

95th:
@sgrif I rewrote commit history where each commit updates one area. Is that sufficient or do you want me to open separate PRs?

src/client.rs
```diff
@@ -182,9 +172,9 @@ use std::usize;
 ///
 /// [module]: index.html
 #[must_use = "futures do nothing unless polled"]
-pub struct Handshake<T, B = Bytes> {
+pub struct Handshake<'a, T, B = Bytes> {
     builder: Builder,
-    inner: WriteAll<T, &'static [u8]>,
+    inner: Pin<Box<dyn Future<Output = io::Result<T>> + 'a>>,
```
seanmonstar: With this holding a `Box<dyn Future>`, the type is no longer `Send` or `Sync`, and to add those bounds to the trait object would prevent using this with a `!Send` transport type. It's unfortunate, but that's the reason the `WriteAll` type was there initially. Can we still name the future instead of needing a trait object?

src/codec/framed_write.rs
```diff
@@ -267,23 +268,18 @@ impl<T, B> FramedWrite<T, B> {
...
 }
 
-impl<T: io::Read, B> io::Read for FramedWrite<T, B> {
-    fn read(&mut self, dst: &mut [u8]) -> io::Result<usize> {
-        self.inner.read(dst)
-    }
-}
-
-impl<T: AsyncRead, B> AsyncRead for FramedWrite<T, B> {
-    fn read_buf<B2: BufMut>(&mut self, buf: &mut B2) -> Poll<usize, io::Error>
```
seanmonstar: We also want to forward `poll_read_buf`, to allow for gathering reads.

src/proto/connection.rs
```diff
@@ -292,52 +289,60 @@ where
...
-                debug_assert_eq!(reason, Reason::NO_ERROR, "graceful GOAWAY should be NO_ERROR");
+                Some(Err(e)) => return Poll::Ready(Err(e.into())),
+                None => (),
             }
-            try_ready!(self.poll_ready());
+            ready!(self.poll_ready(cx))?;
 
-            match try_ready!(self.codec.poll()) {
-                Some(Headers(frame)) => {
+            match ready!(Pin::new(&mut self.codec).poll_next(cx)) {
```
seanmonstar: I thought there was a `Try` impl for `Poll<Option<Result<T, E>>` so all the `Ok(..)` patterns could be skipped. If not, oh well.

tests/h2-support/src/notify.rs
```diff
@@ -21,22 +19,22 @@ impl MockNotify {
         }
     }
 
-    pub fn with<F: FnOnce() -> R, R>(&self, f: F) -> R {
-        use futures::Async::Ready;
-        use futures::future::poll_fn;
+    pub fn with<F: FnOnce() -> R, R>(&self, _f: F) -> R {
+        unimplemented!();
```
seanmonstar: No longer used?

tests/h2-tests/tests/client_request.rs
```diff
@@ -84,431 +90,429 @@ fn recv_invalid_server_stream_id() {
...
-        .and_then(|(mut client, h2)| {
-            // we send a simple req here just to drive the connection so we can
-            // receive the server settings.
-            let request = Request::builder()
-                .method(Method::POST)
-                .uri("https://example.com/")
-                .body(())
-                .unwrap();
+        .await;
+        srv.send_frame(frames::headers(1).response(200).eos()).await;
```
seanmonstar: This isn't too big of a deal, but thought I'd bring it up: I'm partial to the chaining API that existed before, as it seemed a little more succinct. Is it difficult to keep?

cc @carllerche @hawkw @sgrif

Cargo.toml
```diff
@@ -41,8 +41,9 @@ members = [
 ]
 
 [dependencies]
-futures = "0.1"
-tokio-io = "0.1.4"
+futures-preview = "0.3.0-alpha.18"
+tokio-io = { git = "https://github.com/tokio-rs/tokio" }
```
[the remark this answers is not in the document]
95th: Done 👍

examples/client.rs
```diff
@@ -12,76 +18,70 @@ struct Process {
 }
 
 impl Future for Process {
-    type Item = ();
-    type Error = h2::Error;
+    type Output = Result<(), h2::Error>;
 
-    fn poll(&mut self) -> Poll<(), h2::Error> {
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
```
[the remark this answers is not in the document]
95th: Done 👍

src/codec/framed_write.rs
95th: Done

src/proto/connection.rs
95th: Done

tests/h2-support/src/notify.rs
95th: Yes. `MockNotify` is not used anymore.

src/client.rs
95th: The problem is that the WriteAll struct is private now. One solution I have in mind is make the handshake fns async. The handshake path has only one asynchronous element and that is `write_all`. We can just await it right there in the `handshake2` function. WDYT?

tests/h2-tests/tests/client_request.rs
95th: I think the tests reflect the future usage. IMO async-await is much clearer than chaining. In fact combinators based chaining API is main reason I abandoned a project I started 2 years back. Async-await has finally made it look like regular code. It is just my opinion though. 

Also, Tests were the major part of this work. The actual lib update was trivial.

Please let me know if you'd like me to revert it.

src/client.rs
95th: In hyper we store the handshake future in a state enum.

https://github.com/hyperium/hyper/blob/a503f8b3369ff24ce39ccf1a1da0d1d56366dabb/src/proto/h2/client.rs#L34-L37

Making handshake functions async will return an opaque future which we won't be able to store without Boxing. Will we update that area to async functions in hyper 0.13?  From what I understand, Tokio is also moving towards async fns where possible. (recent examples: `poll_accept` vs `accept`, `poll_ref` vs `recv_ref`)

tests/h2-tests/tests/client_request.rs
seanmonstar: It's alright, we can keep the new way. Thinking about in isolation, it does seem like one would expose these as async fns now. I just noticed what appeared to be noise adding all the `await`s.

src/client.rs
seanmonstar: @carllerche thoughts here?

LucioFranco: Is it fine for now to provide a custom Writeall impl within h2?

carllerche: The idea is that APIs that use `write_all` would be written w/ `async fn`. Can this happen here?

carllerche: This would obviously bubble into hyper, would it be acceptable for hyper to also use `async fn` or is there a blocker somewhere?

seanmonstar: It could, though that would make it more difficult to allow the client `Executor` to be `Send` or `!Send`, since we couldn't name that spawned future...

95th: Done 👍

LucioFranco:
@95th awesome job! 🎉

PR #392, opened 2019-08-03
Remove byteorder-dependency

lukaslueg:
This removes the `byteorder`-dependency, which is not needed since 1.32, and gets pulled in _a lot_ due to `h2`.

Running `rustfmt` kind of screwed the branch up, I can reset if so desired. The actual change is just a couple of lines.


carllerche:
Thanks!

PR #402, opened 2019-08-30
Prune futures-* dependencies

seanmonstar:


src/lib.rs
```diff
@@ -91,6 +91,15 @@ macro_rules! proto_err {
     };
 }
 
+macro_rules! ready {
```
taiki-e: FYI: futures_core/futures_util provides this: [futures_core::ready](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.18/futures_core/macro.ready.html)/[futures_util::ready](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.18/futures_util/macro.ready.html)

seanmonstar: Ah yea, it was the first thing I added in this branch, and then realized I still needed futures_util for `poll_fn`... :cry:

PR #408, opened 2019-09-13
fixing the akamai example to work with tokio-0.2

rrichardson:
This PR brings the rustls and webpki current (resolving some version conflicts with ring)  
It also leverages the tokio-0.2 branch to make the akamai example work^H^H^H^Hcompile again.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,7 +71,7 @@ serde_json = "1.0.0"
 # Akamai example
 tokio = "=0.2.0-alpha.4"
 env_logger = { version = "0.5.3", default-features = false }
-rustls = "0.12"
-tokio-rustls = "0.5.0"
-webpki = "0.18"
-webpki-roots = "0.14"
+rustls = "0.16"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -71,7 +71,7 @@ serde_json = "1.0.0"
 # Akamai example
 tokio = "=0.2.0-alpha.4"
 env_logger = { version = "0.5.3", default-features = false }
-rustls = "0.12"
-tokio-rustls = "0.5.0"
-webpki = "0.18"
-webpki-roots = "0.14"
+rustls = "0.16"
+tokio-rustls = { git = "https://github.com/quininer/tokio-rustls", branch = "tokio-0.2" }
```
LucioFranco: I believe there is a release of this on crates?

rrichardson: @lucab - I think you're right.. it looks like the release is from that branch, in fact.

rrichardson:
@LucioFranco - Can you review?  I have addressed the dependency issue.

LucioFranco:
@carllerche @seanmonstar ^ I don't have merge perms here.

seanmonstar:
Thanks :D

PR #410, opened 2019-09-20
Update to tokio 0.2 alpha.5

seanmonstar:


seanmonstar:
Woops, slightly too slow: https://github.com/hyperium/h2/pull/409

PR #421, opened 2019-10-04
Add ability to adjust INITIAL_WINDOW_SIZE setting on an existing connection

seanmonstar:


src/proto/streams/flow_control.rs
```diff
@@ -146,6 +146,22 @@ impl FlowControl {
...
+    pub fn dec_recv_window(&mut self, sz: WindowSize) {
+        log::trace!(
+            "dec_recv_window; sz={}; window={}, available={}",
+            sz,
+            self.window_size,
+            self.available
+        );
+        // This should not be able to overflow `window_size` from the bottom.
+        self.window_size -= sz;
+        self.available -= sz;
```
carllerche: I have forgotten a lot about how things work. Isn't it OK for the available window to go in the negative when updating the window size from a settings frame?

PR #422, opened 2019-10-06
Migrate CI to GitHub Actions

### LINKED ISSUE #415: Move to Github Actions
Travis is very very slow. It would be nice to move to github actions.

taiki-e:
Closes #415


.github/workflows/CI.yml
```diff
@@ -0,0 +1,72 @@
...
+    name: Publish Documentation
+    needs: [style, test]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Install Rust
+        run: rustup update beta && rustup default beta
+      - name: Build documentation
+        run: cargo doc --no-deps --all-features
+      - name: Publish documentation
```
taiki-e: This is based on [the script used in stdarch.](https://github.com/rust-lang/stdarch/blob/master/.github/workflows/main.yml#L32-L39)

.github/workflows/CI.yml
```diff
@@ -0,0 +1,71 @@
...
+  test:
+    name: Test
+    needs: [style]
+    runs-on: ubuntu-latest
+    env:
+      RUSTFLAGS: -Dwarnings
+    strategy:
+      matrix:
+        rust:
+          - nightly
```
taiki-e: GitHub Actions don't have allow_failures, so we probably need to use a way like https://stackoverflow.com/questions/57850553/github-actions-check-steps-status.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,71 @@
+name: CI
+on: [push, pull_request]
+
+jobs:
+  style:
+    name: Check Style
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Install Rust
```
LucioFranco: what about using a rust-toolchain file?

LucioFranco:
@carllerche @seanmonstar Are you all +1 on this?

.github/workflows/CI.yml
taiki-e: H2 has some checks that require nightly...
https://github.com/hyperium/h2/blob/3cfcab016e17f36884ee77ff85947402535468ee/.travis.yml#L47-L51

PR #425, opened 2019-10-11
A patch for server_push and a working example

le-invoker:
A `server_push` and `client_push` example is added to show the usage of `server_push`

### DIFF
--- a/src/proto/streams/streams.rs
+++ b/src/proto/streams/streams.rs
@@ -984,6 +984,8 @@ impl<B> StreamRef<B> {
         let mut me = self.opaque.inner.lock().unwrap();
         let me = &mut *me;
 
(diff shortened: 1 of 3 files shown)


src/proto/streams/streams.rs
```diff
@@ -984,6 +984,8 @@ impl<B> StreamRef<B> {
         let mut me = self.opaque.inner.lock().unwrap();
         let me = &mut *me;
 
+        me.refs += 1;
```
seanmonstar: This will need to go down lower, perhaps right after `let opaque = OpaqueStreamRef::new(..)`. That's because there's some checks in the middle that might not allow the push promise, and so we shouldn't increase the `refs` then.

PR #428, opened 2019-11-26
Update to Tokio 0.2

### LINKED ISSUE #427: Update to tokio 0.2

seanmonstar:
Closes #427


seanmonstar:
Status: the `hpack` encoder was depending on the non-growing nature of `BytesMut`, and so tests are breaking now that `BytesMut` will grow.

I'm working on adjusting it to take a `Limit` to prevent the growing.

seanmonstar:
Looks like `minimal-versions` is failing because `tokio-util` doesn't use an up-to-date enough `log`...

carllerche:
we can remove the log dep from tokio-util. Probably easiest.

PR #429, opened 2019-11-28
Re-enable tokio-rustls now that it has a stable release

alex:


alex:
(Happy Thanksgiving! 🦃🦃🦃)

martell:
After getting this merged it would be amazing if we could get another alpha, though maybe this means that 0.2 final can be released? ^^

I see that @bluejekyll  https://github.com/bluejekyll/trust-dns/pull/932 is blocked by this.
And I am blocked by that https://github.com/actix/actix-net/pull/66

PR #434, opened 2019-12-03
Prepare v0.2.0

seanmonstar:


Cargo.toml
```diff
@@ -5,11 +5,14 @@ name = "h2"
...
+version = "0.2.0"
 license = "MIT"
-authors = ["Carl Lerche <me@carllerche.com>"]
+authors = [
+  "Carl Lerche <me@carllerche.com>",
+  "Sean McArthur <sean@seanmonstar.com>",
+]
 description = "An HTTP/2.0 client and server"
-documentation = "https://docs.rs/h2/0.2.0-alpha.3/h2/"
+documentation = "https://docs.rs/h2/0.2.0/h2/"
```
LucioFranco:
```suggestion
documentation = "https://docs.rs/h2/0.2.0"
```
I don't think this needs that at the end?

seanmonstar: It's not needed, but without it redirects to with it, so 🤷‍♂ .

Issue #234, opened 2018-03-07
Support CONNECT requests


carllerche:
`CONNECT` requests have a number of special requirements. Currently, no focus has been put on making this work.

This is a continuation of #228.

nox:
@kornelski did a PR to support some details at the end of January, does anyone know what's missing still?

nox:
AFAIK, this can be closed thanks to #534, I don't think any other requirement is related to this crate itself, just proxies using it.

PR #445, opened 2020-01-31
Handle :scheme and :path in CONNECT

kornelski:
If the :method is CONNECT the :scheme and :path pseudo-header fields MUST be omitted

This is for #234

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -121,7 +121,7 @@ use crate::proto::{self, Config, Prioritized};
 use crate::{FlowControl, PingPong, RecvStream, SendStream};
 
 use bytes::{Buf, Bytes};
-use http::{HeaderMap, Request, Response};
+use http::{HeaderMap, Method, Request, Response};
 use std::future::Future;
 use std::pin::Pin;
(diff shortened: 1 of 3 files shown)


kornelski:
I've verified that Node.js's HTTP/2 can make a request to an h2 server from examples.

```js
const http2 = require('http2');

const client = http2.connect('http://127.0.0.1:5928');

// Must not specify the ':path' and ':scheme' headers
// for CONNECT requests or an error will be thrown.
const req = client.request({
  ':method': 'CONNECT',
  ':authority': `127.0.0.1:5928`,
});

req.on('response', (headers) => {
  console.log("resp", headers[http2.constants.HTTP2_HEADER_STATUS], headers);
});

req.on('error', (err) => console.error("boo", err));
req.setEncoding('utf8');
let data = '';
req.on('data', (chunk) => data += chunk);
req.on('end', () => {
  console.log(`The server says: ${data}`);
  client.close();
});
req.end('Jane');
```

kornelski:
I've added a test and rebased on master.

PR #446, opened 2020-01-31
Read body in the example server

kornelski:
Expands the example to demonstrate reading of the body.

### DIFF
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -28,17 +28,22 @@ async fn handle(socket: TcpStream) -> Result<(), Box<dyn Error + Send + Sync>> {
     println!("H2 connection bound");
 
     while let Some(result) = connection.accept().await {
-        let (request, mut respond) = result?;
+        let (mut request, mut respond) = result?;
         println!("GOT request: {:?}", request);
+
+        let body = request.body_mut();
+        while let Some(data) = body.data().await {
+            let data = data?;
+            println!("<<<< recv {:?}", data);
+        }
+
         let response = http::Response::new(());
-
         let mut send = respond.send_response(response, false)?;
-
-        println!(">>>> sending data");
-        send.send_data(Bytes::from_static(b"hello world"), true)?;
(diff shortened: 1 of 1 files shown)


kornelski:
Is this code correct? 

```rust
       let body = request.body_mut();
        while let Some(data) = body.data().await {
            let data = data?;
            println!("<<<< recv {:?}", data);
        }
```

This loop gets stuck after receiving 65535 bytes.

```
GOT request: Request { method: POST, uri: http://127.0.0.1:5928/, version: HTTP/2.0, headers: {"user-agent": "curl/7.54.0", "accept": "*/*", "content-length": "100000", "content-type": "application/x-www-form-urlencoded"}, body: RecvStream { inner: FlowControl { inner: OpaqueStreamRef { stream_id: StreamId(1), ref_count: 2 } } } }
<<<< recv 16384 bytes
<<<< recv 16384 bytes
<<<< recv 16384 bytes
<<<< recv 16383 bytes
```

I'm testing with `curl`

```sh
yes | head -n 100000 > f
curl --http2-prior-knowledge http://127.0.0.1:5928  -d @f
```

Tonyce:
also have the problem
截屏2021-03-12 下午8 20 10

nox:
I tried this with `body.flow_control().release_capacity(data.len())` in the loop but that didn't help either.

nox:
I am almost certain the issue is on curl's side.

nox:
Hah, never mind, I managed to repro that with the client example connecting to the server example 😅

nox:
Superseded by #544, turns out we were not driving the connection at all anymore.

kornelski:
In any one is wondering why this didn't work — it needs a call to `poll_close` 

https://docs.rs/h2/0.3.3/h2/server/index.html#managing-the-connection

nox:
Oh `poll_close` is supposed to be deprecated (in favour of `poll_closed`), I will update those docs.

Issue #448, opened 2020-02-19
TCP backpressure?


goffrie:
From what I can tell, h2 will buffer send data unboundedly. This is mitigated using `poll_capacity` if the receiver is using HTTP/2 flow control, but if they aren't (e.g. if they advertise a massive window) this doesn't really help. Would it be reasonable to add some mechanism to wait for h2's internal buffers to drain (to the underlying connection) before continuing to write?

Relatedly, I'm also interested in particular in knowing when a particular stream's data has been fully written to the socket.

goffrie:
We realized that this can sort of be tracked already by changing the `B: Buf` type - h2 will drop it once the data has been sent off. This is a bit hacky but it does neatly cut through intermediate abstractions like hyper.

PR #449, opened 2020-02-20
[wip] Add an API to track and wait on the amount of data in h2's internal buffers.

goffrie:
This is my stab at #448. Does this seem like a reasonable API?

I guess that technically this still isn't fully accurate in that data could be buffered inside of the `Codec`. Not sure if that's a big deal though.


goffrie:
I can also imagine an API where we offer a "cap" on the available window and then roll it all into `.capacity()` / `.poll_capacity()`.

PR #450, opened 2020-02-25
Enable support for multiple user pings

seanmonstar:
This expands the user `PING` support from only having a single outstanding ping to multiple.

To keep limit allocations, the single `AtomicUsize` is still used to track all the pings, but it is treated like a bit-array to track state of each ping. The original user ping payload is used as a prefix for the first 7 bytes, and then a "ping id" is used to determine the 8th byte (this restriction is to prevent collisions with our internally reserved ping payloads).

An unresolved question is what API to expose to construct different pings?

1. Multiple constructors, like `Ping::user_0()`, `Ping::user_1()`, etc. This completely removes any error conditions with specifying an "id" higher than the amount of pings we support.
2. Provide `Ping::new(u8)`, which asserts that the ID is less than the max allowed.
3. Same `Ping::new(u8)`, but return an error instead of asserting in `send_ping(ping)`.


seanmonstar:
@LucioFranco This is a prerequisite for adding keep-alive ping support in Tonic.

seanmonstar:
Hm, I actually think I may not need this. hyper can use the `Ping::opaque()` for both BDP and keep-alive, since they are in-flight in different circumstances...

LucioFranco:
Sorry I never got this, happy to review this if you want to move forward with this change @seanmonstar

LucioFranco:
@seanmonstar ping, do you still want a review on this?

seanmonstar:
Nah, probably not, thanks!

PR #462, opened 2020-04-30
Implement `Error::source` for `h2::Error`.

goffrie:
This makes it easier to detect an IO error at the bottom of an error
cause chain (e.g. behind a `hyper::Error`).

Also, don't implement `std::error::Error` for the private error types,
to avoid confusion.


src/error.rs
```diff
@@ -132,4 +132,11 @@ impl fmt::Display for Error {
     }
 }
 
-impl error::Error for Error {}
+impl error::Error for Error {
+    fn source(&self) -> Option<&(dyn error::Error + 'static)> {
+        match self.kind {
+            Kind::Io(ref e) => Some(e),
```
seanmonstar, line 138: I always find this part a little tricky. I *think* since the `h2::Error` pretty much *is* the `io::Error`, this should instead be `e.source()`, similar to how `io::Error` works with a custom error.

goffrie: Hm. The specific reason why I want this is so that I can traverse a `&dyn Error`'s cause chain and discover an `io::Error`; right now I have to also check for an `h2::Error` explicitly, even if that error is an IO error.

alanhe:
Can we push this forward?

In my case, when a client crashes, the server gets an error similar to this.
```
// dbg!(&source);

 &source = hyper::Error(
    Body,
    Error {
        kind: Io(
            Kind(
                BrokenPipe,
            ),
        ),
    },
)
```

I want to match the error kind to record the event, but recursively downcasting the error to `std::io::Error` couldn't work because `h2::Error` does not implement `std::error::Error#source()`.

This is what I end up with:
```rust
fn match_broken_pipe_error(err: &(dyn std::error::Error + 'static)) -> bool {
    if let Some(e) = err.downcast_ref::<h2::Error>() {
        if let Some(e) = e.get_io() {
            if e.kind() == std::io::ErrorKind::BrokenPipe {
                return true;
            }
        }
    }

    if let Some(source) = err.source() {
        match_broken_pipe_error(source)
    } else {
        false
    }
}
```
The code is much more awkward.

kaifastromai:
oof...any update on this? Ran into this myself

mkatychev:
This is still an issue esp while matching on a stream disconnect in tonic: https://github.com/hyperium/tonic/issues/1579
@seanmonstar what is needed to move this PR forward, happy to start a separate feature branch? This will still affect grpc disonnects since `tonic`  uses `h2` 0.3 (so it would have to be backported as well).

mkatychev:
@seanmonstar could we get some feedback on this PR? This would greatly simplify some tonic error handling cases:
https://github.com/hyperium/tonic//blob/f7090c24813041dc62052894dee119520d603d88/examples/src/streaming/server.rs#L23-L29

kaifastromai:
Any update this?? @seanmonstar

PR #475, opened 2020-07-07
feat(lib): switch from log to tracing

davidbarsky:
Like https://github.com/hyperium/hyper/pull/2204, this PR switches from `log` to `tracing`. It's a slightly noisy diff because I didn't want to make a _big_ change to `h2` in this PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ tokio-util = { version = "0.3.1", features = ["codec"] }
 tokio = { version = "0.2", features = ["io-util"] }
 bytes = "0.5.2"
 http = "0.2"
-log = "0.4.1"
+tracing = "0.1"
 fnv = "1.0.5"
 slab = "0.4.0"
 indexmap = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ tokio-util = { version = "0.3.1", features = ["codec"] }
 tokio = { version = "0.2", features = ["io-util"] }
 bytes = "0.5.2"
 http = "0.2"
-tracing = { version = "0.1", default-features = false, features = ["std", "log"] }
+tracing = { version = "0.1.13", default-features = false, features = ["std", "log"] }
 fnv = "1.0.5"
 slab = "0.4.0"
 indexmap = "1.0"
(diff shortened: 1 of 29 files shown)


Cargo.toml
```diff
@@ -47,7 +47,7 @@ tokio-util = { version = "0.3.1", features = ["codec"] }
 tokio = { version = "0.2", features = ["io-util"] }
 bytes = "0.5.2"
 http = "0.2"
-log = "0.4.1"
+tracing = "0.1"
```
hawkw, line 50: We should probably enable the `log` feature for downstream users who expect `log` records to be emitted.
```suggestion
tracing = { version = "0.1", default-features = false, features = ["std", "log"] }
```

src/codec/framed_read.rs
```diff
@@ -74,7 +74,7 @@ impl<T> FramedRead<T> {
 
         let kind = head.kind();
 
-        log::trace!("    -> kind={:?}", kind);
+        tracing::trace!("    -> kind={:?}", kind);
```
hawkw, line 77: stuff like this seems like it would be pretty trivial to turn into a structured field...but we can do that in a follow up

davidbarsky:
I've rebased against mainline and included @hawkw's commit.

seanmonstar:
Hm, looks like CI is failing with `-Zminimal-versions`. Do we need to pin to a later version of `tracing`?

davidbarsky:
> Hm, looks like CI is failing with `-Zminimal-versions`. Do we need to pin to a later version of tracing?

Possibly. I'm thinking that we select _at least_ the latest stable `tracing` release?

seanmonstar:
It's forcing cargo to download the minimum versions allowed by the `Cargo.toml`'s involved. It seems it's complaining that `tracing-attributes` doesn't compile on Rust 1.39, but I thought we disabled it. Perhaps the minimum version here still included it?

hawkw:
@seanmonstar aah, I see --- `tracing-attributes` was made an optional dependency in `tracing` 0.1.13, before then it was always downloaded even with `default-features = false` (yes, this was an unintentional breaking change for `default-features = false` users, it's one of my bigger regrets). We probably want to depend on 0.1.13 and later.

Side note, we should probably add minimal versions checks for tracing CI (cc @davidbarsky).

seanmonstar:
Ah, then this CI failure is kind of dumb. We don't really want to require tracing >0.1.13 for anything inside `h2`, but we have to in order to make the Rust 1.39 + minimal-versions check pass... Oh well, probably not bad to require a newer version.

PR #478, opened 2020-07-31
[WIP] start adding `tracing` spans to internals

hawkw:
We've adopted `tracing` for diagnostics, but currently, it is just being
used as a drop-in replacement for the `log` crate. Ideally, we would
want to start emitting more structured diagnostics, using `tracing`'s
`Span`s and structured key-value fields.

A lot of the logging in `h2` is already written in a style that imitates
the formatting of structured key-value logs, but as textual log
messages. Migrating the logs to structured `tracing` events therefore is
pretty easy to do. I've also started adding spans, mostly in the read
path.

Finally, I've updated the tests to use `tracing` rather than
`env_logger`. The tracing setup happens in a macro, so that a span for
each test with the test's name can be generated and entered. This will
make the test output easier to read if multiple tests are run
concurrently with `--nocapture`.

### DIFF
--- a/src/codec/framed_read.rs
+++ b/src/codec/framed_read.rs
@@ -61,6 +61,8 @@ impl<T> FramedRead<T> {
 
     fn decode_frame(&mut self, mut bytes: BytesMut) -> Result<Option<Frame>, RecvError> {
         use self::RecvError::*;
+        let span = tracing::trace_span!("decode_frame", offset = bytes.len());
+        let _e = span.enter();
 
         tracing::trace!("decoding frame from {}B", bytes.len());
 
@@ -74,7 +76,7 @@ impl<T> FramedRead<T> {
 
         let kind = head.kind();
 
-        tracing::trace!("    -> kind={:?}", kind);
+        tracing::trace!(frame.kind = ?kind);
 
         macro_rules! header_block {
             ($frame:ident, $head:ident, $bytes:ident) => ({
@@ -338,6 +340,8 @@ where
     type Item = Result<Frame, RecvError>;
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let span = tracing::trace_span!("framed_read");
+        let _e = span.enter();
         loop {
             tracing::trace!("poll");
             let bytes = match ready!(Pin::new(&mut self.inner).poll_next(cx)) {
@@ -346,9 +350,9 @@ where
                 None => return Poll::Ready(None),
             };
 
-            tracing::trace!("poll; bytes={}B", bytes.len());
+            tracing::trace!(read.bytes = bytes.len());
             if let Some(frame) = self.decode_frame(bytes)? {
-                tracing::debug!("received; frame={:?}", frame);
+                tracing::debug!(?frame, "received;");
                 return Poll::Ready(Some(Ok(frame)));
             }
         }
--- a/src/codec/framed_read.rs
+++ b/src/codec/framed_read.rs
@@ -61,7 +61,7 @@ impl<T> FramedRead<T> {
 
     fn decode_frame(&mut self, mut bytes: BytesMut) -> Result<Option<Frame>, RecvError> {
         use self::RecvError::*;
-        let span = tracing::trace_span!("decode_frame", offset = bytes.len());
+        let span = tracing::trace_span!("FramedRead::decode_frame", offset = bytes.len());
         let _e = span.enter();
 
         tracing::trace!("decoding frame from {}B", bytes.len());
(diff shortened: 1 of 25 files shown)


hawkw:
this is currently a draft because i haven't had the chance to add spans in the write/encode path as much as i have in the read/decode path. wanted to put it up now for folks to start checking out.

src/hpack/decoder.rs
```diff
@@ -191,14 +194,14 @@ impl Decoder {
             // determined from the first byte.
             match Representation::load(ty)? {
                 Indexed => {
-                    tracing::trace!("    Indexed; rem={:?}", src.remaining());
+                    tracing::trace!(rem = src.remaining(), "    Indexed;");
```
hawkw, line 197: we may want to remove the indentation in these event messages --- it's essentially an ad-hoc way of showing the same information that's better captured by the span context. however, i thought it might be worth leaving in because some folks are still consuming these events with `log` rather than `tracing`.

seanmonstar: The indentation is purely for the sanity of the pure soul consuming the `trace` level logs. If a span provides the same thing, that seems fine (or probably even better).

hawkw: Yeah, I just thought that _some_ users might still be consuming those logs through the `log` crate, so they wouldn't see the span context.

davidbarsky: Not sure how many people would be broken up by _this_ change. I'd maybe get rid of the extra spaces?

hawkw:
Added a bit more to the write side. Here's a quick demo:
Screenshot_20200731_171017

There's more to do, of course. I haven't touched flow control & prioritization at all yet, and it would be nice to get a single top-level span for each stream in a connection, which I haven't done yet...but it might be worth moving this out of draft so we can roll forward?

src/codec/framed_read.rs
```diff
@@ -346,9 +350,9 @@ where
                 None => return Poll::Ready(None),
             };
 
-            tracing::trace!("poll; bytes={}B", bytes.len());
+            tracing::trace!(read.bytes = bytes.len());
             if let Some(frame) = self.decode_frame(bytes)? {
-                tracing::debug!("received; frame={:?}", frame);
+                tracing::debug!(?frame, "received;");
```
seanmonstar, line 355: What's the difference with having some lines where the message is first, and other lines where the values are first?

src/codec/framed_write.rs
```diff
@@ -105,8 +105,10 @@ where
     pub fn buffer(&mut self, item: Frame<B>) -> Result<(), UserError> {
         // Ensure that we have enough capacity to accept the write.
         assert!(self.has_capacity());
+        let span = tracing::trace_span!("FramedWrite::buffer", frame = ?item);
```
seanmonstar, line 108: I noticed some of the spans are prefixed with the struct name, and others aren't. Is one preferred over the other? Should we be consistent?

src/proto/connection.rs
```diff
@@ -200,9 +202,12 @@ where
 
     /// Advances the internal state of the connection.
     pub fn poll(&mut self, cx: &mut Context) -> Poll<Result<(), proto::Error>> {
+        let span = tracing::trace_span!("connection", peer = %P::NAME);
```
seanmonstar, line 205: Should this be something like `connection::poll`? `connection` sounds *to me* like a span that measures the full connection...

src/codec/framed_read.rs
davidbarsky: `frame` is recorded as a key/value pair where the key is `frame` and the value is the `fmt::Debug` output of `frame`.

src/codec/framed_write.rs
hawkw: I've seen span names that are prefixed, and span names that are not prefixed. I think it depends somewhat on whether the name of the function is unambiguous on its own, or if the name of the type is necessary to make the meaning clear. e.g. if there is only one `handshake` function, it doesn't need a prefix, but a `poll` function definitely needs a prefix, because there are lots of `poll`s. 

I've tried to err on the side of brevity and used un-prefixed names when they don't seem strictly necessary, and only use prefixed names when they do. Brevity seemed worthwhile, because at the highest verbosity levels, the span contexts can get quite long. However, if folks prefer, I'm happy to err on the side of consistency rather than brevity.

src/proto/connection.rs
hawkw: Good call. Perhaps we want to add a span to the connection struct, either instead of or in addition to this span.

src/codec/framed_read.rs
seanmonstar: Yea, I just noticed some are like this one, `debug!(?thing, "message")`, and others are `debug!("message", ?thing)`. If we can do either, we should probably be consistent (and if I can vote, I like the message first, since that tells me "why" there's a message).

hawkw:
> LGTM, squash and merge whenever you want.

will do as soon as CI works!

PR #487, opened 2020-09-16
Do not require authority for UNIX domain sockets

saschagrunert:
This fixes connections where paths to UNIX domain sockets (`*.sock`) are provided or the authority is empty.

Refers to https://github.com/hyperium/tonic/issues/243, https://github.com/cri-o/cri/issues/34

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1395,14 +1395,20 @@ impl proto::Peer for Peer {
         // A request translated from HTTP/1 must not include the :authority
         // header
         if let Some(authority) = pseudo.authority {
-            let maybe_authority = uri::Authority::from_maybe_shared(authority.clone().into_inner());
-            parts.authority = Some(maybe_authority.or_else(|why| {
-                malformed!(
-                    "malformed headers: malformed authority ({:?}): {}",
-                    authority,
-                    why,
-                )
-            })?);
+            // When connecting to a UNIX Domain Socket (UDS), then we might get a path for the
+            // authority field. If it's a local path and exists, then we do not error in that case
+            // and assume an UDS.
+            if !authority.is_empty() && !authority.ends_with(".sock") {
+                let maybe_authority =
+                    uri::Authority::from_maybe_shared(authority.clone().into_inner());
+                parts.authority = Some(maybe_authority.or_else(|why| {
+                    malformed!(
+                        "malformed headers: malformed authority ({:?}): {}",
+                        authority,
+                        why,
+                    )
+                })?);
+            }
         }
 
         // A :scheme is required, except CONNECT.


saschagrunert:
Test are failing now for sure. In case of an unix socket, we would only get a path and nothing else. Do you see any feasible way to implement this in H2?

seanmonstar:
The description mentions not requiring an authority, but it looks like this changes it to stop checking if the provided authority is valid. So, in the Unix socket case, does it send an `:authority` that is not a normal looking domain name?

saschagrunert:
> The description mentions not requiring an authority, but it looks like this changes it to stop checking if the provided authority is valid. So, in the Unix socket case, does it send an `:authority` that is not a normal looking domain name?

Unfortunately not, the authority would be the complete path, like `/path/to/my/socket.sock`

saschagrunert:
Would it be okay to early exit the validation path if the parsed URI is a path? Not sure how to classify a “path” though…

saschagrunert:
@seanmonstar I checked for a path. If this exists locally, then we assume that the connection through the UDS should work.

saschagrunert:
Is there anyone over here to give this a review? :innocent:

src/server.rs
```diff
@@ -1393,14 +1394,20 @@ impl proto::Peer for Peer {
...
-                malformed!(
-                    "malformed headers: malformed authority ({:?}): {}",
-                    authority,
-                    why,
-                )
-            })?);
+            // When connecting to a UNIX Domain Socket (UDS), then we might get a path for the
+            // authority field. If it's a local path and exists, then we do not error in that case
+            // and assume an UDS.
+            if !Path::new(authority.as_str()).exists() {
```
seanmonstar, line 1400: This will do a blocking IO call, which we wouldn't want to do in an async environment.

seanmonstar (commented):
As I mentioned in https://github.com/grpc/grpc-go/issues/3854#issuecomment-704950114, I think it's fine to ignore the `:authority` if it's a blank value. I would expect a client to not send an *invalid* authority.

saschagrunert:
> As I mentioned in [grpc/grpc-go#3854 (comment)](https://github.com/grpc/grpc-go/issues/3854#issuecomment-704950114), I think it's fine to ignore the `:authority` if it's a blank value. I would expect a client to not send an _invalid_ authority.

Yes, but in case of a unix domain socket the authority is not blank. It's the socket path.

src/server.rs
saschagrunert: Removed in favor of a check for a `.sock` suffix.

saschagrunert:
PTAL @seanmonstar 

I changed the check to either exclude `*.sock` authorities or if the authority is empty.

Ruilkyu:
Has 0.4.0 solved the problem？

arthurlm:
Just want to add that I am facing a similar issue using h2 to process gRPC requests using [tonic](https://github.com/hyperium/tonic).
I have also open an issue https://github.com/hyperium/h2/issues/525

I am using h2 v0.3.1.

I have fixed the problem by just sending a warning when `:authority` is invalid instead of rejecting the requests.
See my fix here: https://github.com/hyperium/h2/compare/master...arthurlm:master

Maybe we should find another more elegant solution than just ignoring `*.sock` or other _allowed list_ `:authority` pattern ?

nox:
Didn't grpc-go change to not emit those malformed authority headers?

BogdanIonesq:
Any idea when this could be merged? Calling gRPC methods from Go clients or grpcurl on unix domain sockets still does not work due to this issue.

noboruma:
For anyone trying to get that working, the workaround is to simply send a dummy authority field:
- grpcurl client, add `-authority "dummy"`
- grpc-go client, use `grpc.Dial("unix:///blah.sock", grpc.WithAuthority("dummy"), grpc.WithInsecure())`

bachp:
Changing this on the client side is currently not an option as I can't patch every kubelet.

Currently this issue is preventig me from properly implementing a [Kubernetes Device plugin](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/) in Rust using Tonic.

arthurlm:
I am exactly in the same situation: kubelet is the client and patch it is not an option for me either.

This is why few month ago I have fork this repository to disable panic if authority field is `Err` (ie. It has failed to be parsed).

Moreover, from k8s point of view, authority field is not invalid. It represent namespace from where the request came from. This is not just a valid http authority but it still contains meaningful network information.

In my case, forking unlock me and I have been able to make k8s extension works correctly. But from my point of view this is too bad request fail because an optional field cannot be parsed.

bachp:
@saschagrunert I think only checking for `.sock` is kind of brittle, there is nothing forcing a socket to end with `.sock`. A better heuristic might be to check for a leading `/`

@seanmonstar What would be the harm if we just ignore an invalid authority instead of failing with an error?

arthurlm:
Totally agree on the fact there is nothing that enforce a socket file ending with `.sock`.

As example, there is a lot of perfectly valid socket file in every `/tmp` workstation not matching this pattern:
```
❯ find /tmp -type s  -ls 2> /dev/null
      109      0 srwxrwxrwx   /tmp/.X11-unix/X0
```

A better check for that would be to check file descriptor exists and check it is a valid socket file (but it will still not works for unnamed unix socket).

The patch you have made is exactly the same I have suggest around 1 year ago :wink: https://github.com/hyperium/h2/commit/2c26edaa257b934744b13e7c8a8333c94f956d28 .
I have also edit it later to keep log in case authority is not a valid http host (see https://github.com/hyperium/h2/compare/master...arthurlm:master).

If any maintainers want I would be happy to make a PR for this or you could take the solution of @bachp without the log.
It would avoid more people forking for this and let us use `h2` with `tonic` to create nice kubernetes rust extension :grin:.

bachp:
@arthurlm I think logging it makes a lot of sense.

arthurlm:
Before creating any PR, I have done more investigation to understand clearly what is happening.

In my extension service k8s case the `:authority` field received looks like `extension/<target namespace>/<target service>`.

---

Looking at RFCs:

1. [RFC 7540: HTTP V2@section 8.1.2.3. Request Pseudo-Header Fields](https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.3).

> The ":authority" pseudo-header field includes the authority
      portion of the target URI ([[RFC3986], Section 3.2](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2)).  The authority
      MUST NOT include the deprecated "userinfo" subcomponent for "http"
      or "https" schemed URIs.

> All HTTP/2 requests MUST include exactly one valid value for the
   ":method", ":scheme", and ":path" pseudo-header fields, unless it is
   a CONNECT request ([Section 8.3](https://datatracker.ietf.org/doc/html/rfc7540#section-8.3)).  An HTTP request that omits
   mandatory pseudo-header fields is malformed ([Section 8.1.2.6](https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.6)).

2. [RFC 3986: Uniform Resource Identifier@section 3.2. Authority](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2)

> Many URI schemes include a hierarchical element for a naming
   authority so that governance of the name space defined by the
   remainder of the URI is delegated to that authority (which may, in
   turn, delegate it further).

> The authority component is preceded by a double slash ("//") and is
   terminated by the next slash ("/"), question mark ("?"), or number
   sign ("#") character, or by the end of the URI.

> Non-validating parsers (those that merely separate a URI reference into
   its major components) will often ignore the subcomponent structure of
   authority, treating it as an opaque string from the double-slash to
   the first terminating delimiter, until such time as the URI is
   dereferenced.

So, from my understanding:

* Following (1):
  * `:authority` field is not mandatory in HTTP2
  * It should be a valid URI without scheme and userinfo (so k8s might be sending valid content)
  * I have deliberately ignored the CONNECT case since it does not seems relevant to me here

* Following (2):
  * k8s is sending invalid content in `:authority` 
  * data should be split as `:authority = extension` and `:path = <target namespace>/<target service>`
  * `:authority` might be treated as opaque string by URI parser

**NOTE**: I have not read the whole RFCs but just read in details the mentioned sections. If anyone have better understanding of this RFCs, please feel free to comment :wink: !

---

Looking at current code:

* `h2` parse `:authority` at https://github.com/hyperium/h2/blob/3383ef71e2dabe27fe979e8b2c82b61b51eca9f5/src/server.rs#L1467-L1476
* `h2` rely on [http](https://github.com/hyperium/http) crate to parse content
* if [`uri::Authority::from_maybe_shared`](https://docs.rs/http/0.2.6/http/uri/struct.Authority.html#method.from_maybe_shared) fail, h2 forward the error and make the whole request "malformed"
* In k8s case, parse fail because it contains `/` character.
* If `parts` contains valid data it will be used to enchanced `b`
  https://github.com/hyperium/h2/blob/3383ef71e2dabe27fe979e8b2c82b61b51eca9f5/src/server.rs#L1434
  https://github.com/hyperium/h2/blob/3383ef71e2dabe27fe979e8b2c82b61b51eca9f5/src/server.rs#L1520

---

I now think we are here in a corner case.

* There are arguments telling both `h2` and `k8s` are right.
* Fix can be done at multiple level
  * http crate (with an "opaque URI" parser)
  * h2 crate
  * kubernetes and all its extensions projects (way more harder ...)
* (We are really far away here from `*.sock` issue :wink:, which BTW does not seems described in RFCs)

@seanmonstar / [hyperium](https://github.com/hyperium) maintainers: what is your point on view on this ?

Should we consider `:authority` as an opaque string ?
Should we just ignore it when `Authority::from_maybe_shared` fail ?
Any other ideas or comments ?

bachp:
@arthurlm I like your proposal to only ignore the malformed `authority` on `CONNECT`. Maybe just open a new MR with the proposed change to bring it more visibility from the maintainers.

bachp (approved):

mingliangli-lml (approved):
I encountered the same problem, this solution can be applied to the scenario of UDS address.

Forsworns:
@arthurlm 18 days before, k8s approves a [PR](https://github.com/kubernetes/kubernetes/pull/112597) related to this. Seems now tonic and h2 works well with it.

cristiano-prato:
Hi, this issue is still present, is there any plan to fix?

Noah-Kennedy:
@seanmonstar I'm not sure how you would prefer this be tackled, so I'll leave the review to you here.

chensk:
I encountered the same problem. I'm writing a rust grpc server that implements bazel remote cache protocol. Bazel reports "PROTOCOL ERROR" when unix domain socket used. Port is fine. Are we going to fix that?

VasiliyS:
I've come across the same issue. I've implemented a `SPIFFE` Workload API gRPC server that is accepting connections on a `UDS`. Existing 3rd party clients, e.g. Pyhton send the full path of the socket as `authority`, which results in a protocol error.

Would be fantastic to get a fix here.

gauravgahlot:
@seanmonstar @bachp @mingliangli-lml Any plans for the fix to be merged?

PR #488, opened 2020-09-23
Client limits on concurrent pushed streams

eggyal:
Currently, clients enforce their maximum number of receive streams
in Recv::open, refusing any streams that would push the connection
over their specified limit.  This is called by both:

- `Streams::recv_headers` (on receiving a `HEADER` frame); and
- `Streams::recv_push_promise` (on receiving a `PUSH_PROMISE` frame).

The former goes on to call `Counts::inc_num_recv_streams`, which
will panic if the stream pushes the connection over the limit.
Since the counts remain locked between checking and incrementing,
this is okay.

However, in the latter case, the stream is correctly opened on
receipt of the `PUSH_PROMISE` and placed into "reserved (remote)"
state, which does not count toward the concurrency limit.  Per
Section 5.1.2:

> Streams that are in the "open" state or in either of the
> "half-closed" states count toward the maximum number of streams
> that an endpoint is permitted to open. Streams in any of these
> three states count toward the limit advertised in the
> SETTINGS_MAX_CONCURRENT_STREAMS setting. Streams in either of
> the "reserved" states do not count toward the stream limit.

Then, when `Streams::recv_headears` is invoked on the subsequent
receipt of a `HEADER` frame, the stream is already open and the
concurrency check is not performed.  Consequently, its subsequent
call to `Counts::inc_num_recv_streams` will panic if the limit
will now be exceeded.  This could happen because:

- the limit was reached when the `PUSH_PROMISE` was received
  (causing the stream to be refused) but the server sent the
  `HEADER` before receiving/processing the `RST_STREAM`; or

- the limit was reached after the `PUSH_PROMISE` was received,
  e.g. by `HEADER` frames from other promised streams.

In either case, we should refuse the stream instead of panicking.

### DIFF
--- a/src/proto/streams/counts.rs
+++ b/src/proto/streams/counts.rs
@@ -25,6 +25,12 @@ pub(super) struct Counts {
 
     /// Current number of pending locally reset streams
     num_reset_streams: usize,
+
+    /// Maximum number of remote reserved streams
+    max_reserved_streams: usize,
+
+    /// Current number of reserved streams
+    num_reserved_streams: usize,
 }
 
 impl Counts {
@@ -38,6 +44,8 @@ impl Counts {
             num_recv_streams: 0,
             max_reset_streams: config.local_reset_max,
             num_reset_streams: 0,
+            max_reserved_streams: config.remote_reserved_max,
+            num_reserved_streams: 0,
         }
     }
 
@@ -109,6 +117,22 @@ impl Counts {
         self.num_reset_streams += 1;
     }
 
+    /// Returns true if the number of reserved streams can be incremented.
+    pub fn can_inc_num_reserved_streams(&self) -> bool {
+        self.max_reserved_streams > self.num_reserved_streams
+    }
+
+    /// Increments the number of reserved streams.
+    ///
+    /// # Panics
+    ///
+    /// Panics on failure as this should have been validated before hand.
+    pub fn inc_num_reserved_streams(&mut self) {
+        assert!(self.can_inc_num_reserved_streams());
+
+        self.num_reserved_streams += 1;
+    }
+
     pub fn apply_remote_settings(&mut self, settings: &frame::Settings) {
         if let Some(val) = settings.max_concurrent_streams() {
             self.max_send_streams = val as usize;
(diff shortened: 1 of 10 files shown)


eggyal:
We do however currently have the slightly perverse situation where no amount of `PUSH_PROMISE` frames received when just 1 below the limit will trigger any of their streams to be reset, but any number received when at the limit each will.

It would be better not to perform this check on receipt of a `PUSH_PROMISE` but instead only on receipt of the subsequent `HEADER` (when transitioning to "half-closed" state, which counts toward the limit).  After all, just because a server has sent a `PUSH_PROMISE` does not mean that it will immediately send the associated `HEADER` frame: it may (indeed **must**) wait for its number of concurrent streams to drop below the client's limit, so for us to immediately reset the promised streams before the `HEADER` frame is received is premature.

Indeed, looking at [Figure 2] from the spec, it appears that receipt of a `PUSH_PROMISE` frame should not transition the stream through the "open" state at all (which is where we perform this check).

Instead, the related point in [this comment](https://github.com/hyperium/h2/blob/40566689ee65d6a41ab084fc206d394f729335f4/src/proto/streams/streams.rs#L491-L493):

> ```rust
>         // TODO: Streams in the reserved states do not count towards the concurrency
>         // limit. However, it seems like there should be a cap otherwise this
>         // could grow in memory indefinitely.
> ```

should be resolved entirely separately per [Section 10.5 Denial-of-Service Considerations](https://httpwg.org/specs/rfc7540.html#dos) of the spec:

> The number of [PUSH_PROMISE] frames is not constrained in the same fashion. A client that accepts server push SHOULD limit the number of streams it allows to be in the "reserved (remote)" state. An excessive number of server push streams can be treated as a stream error ([Section 5.4.2]) of type [ENHANCE_YOUR_CALM].
> 
> [PUSH_PROMISE]: https://httpwg.org/specs/rfc7540.html#PUSH_PROMISE
> [ENHANCE_YOUR_CALM]: https://httpwg.org/specs/rfc7540.html#ENHANCE_YOUR_CALM
> [Section 5.4.2]: https://httpwg.org/specs/rfc7540.html#StreamErrorHandler

[Figure 2]: https://httpwg.org/specs/rfc7540.html#StreamStatesFigure

I will update this PR.

src/proto/streams/counts.rs
```diff
@@ -122,10 +151,20 @@ impl Counts {
...
 
+        // If the reserved state has changed, adjust counter accordingly
+        if stream.state.is_reserved_remote() != was_reserved {
+            if was_reserved {
+                self.dec_num_reserved_streams();
+            } else {
+                self.inc_num_reserved_streams();
+            }
+        }
+
```
eggyal, line 167: I don't feel entirely comfortable about this: it's on quite a hot path (every single state transition) for relatively rare event (PUSH_PROMISE frame received and subsequent HEADER frame received)—but the alternatives seem a bit hacky too.  I'm going to take a closer look at how state transitions are verified and performed, as it feels like it may benefit from some refactoring.

PR #491, opened 2020-10-15
update to tokio 0.3

jxs:
akamai example needs https://github.com/tokio-rs/tls/pull/29

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,8 +43,8 @@ members = [
 futures-core = { version = "0.3", default-features = false }
 futures-sink = { version = "0.3", default-features = false }
 futures-util = { version = "0.3", default-features = false }
-tokio-util = { version = "0.3.1", features = ["codec"] }
-tokio = { version = "0.2", features = ["io-util"] }
+tokio-util = { version = "0.4.0", features = ["codec"] }
+tokio = { version = "0.3", features = ["io-util"] }
 bytes = "0.5.2"
 http = "0.2"
 tracing = { version = "0.1.13", default-features = false, features = ["std", "log"] }
@@ -66,9 +66,9 @@ serde = "1.0.0"
 serde_json = "1.0.0"
 
 # Examples
-tokio = { version = "0.2", features = ["dns", "macros", "rt-core", "sync", "tcp"] }
+tokio = { version = "0.3", features = ["rt-multi-thread", "macros", "sync", "net"] }
 env_logger = { version = "0.5.3", default-features = false }
-rustls = "0.16"
-tokio-rustls = "0.12.0"
+rustls = "0.18"
+tokio-rustls = "0.20.0"
 webpki = "0.21"
 webpki-roots = "0.17"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ http = "0.2"
 tracing = { version = "0.1.13", default-features = false, features = ["std", "log"] }
 tracing-futures = { version = "0.2", default-features = false, features = ["std-future"]}
 fnv = "1.0.5"
-slab = "0.4.0"
+slab = "0.4.2"
 indexmap = "1.0"
 
 [dev-dependencies]
--- a/src/codec/framed_write.rs
+++ b/src/codec/framed_write.rs
@@ -192,16 +192,23 @@ where
                         tracing::trace!(queued_data_frame = true);
 
                         if self.buf.has_remaining() {
-                            let n = ready!(Pin::new(&mut self.inner).poll_write(cx, self.buf.bytes()))?;
+                            let n =
+                                ready!(Pin::new(&mut self.inner).poll_write(cx, self.buf.bytes()))?;
                             self.buf.advance(n);
-                        } else {
-                            let n = ready!(Pin::new(&mut self.inner).poll_write(cx, frame.payload_mut().bytes()))?;
-                            frame.payload_mut().advance(n);
+                        }
+
+                        let buf = frame.payload_mut();
+
+                        if !self.buf.has_remaining() && buf.has_remaining() {
+                            let n = ready!(Pin::new(&mut self.inner).poll_write(cx, buf.bytes()))?;
+                            buf.advance(n);
                         }
                     }
                     _ => {
                         tracing::trace!(queued_data_frame = false);
-                        let n = ready!(Pin::new(&mut self.inner).poll_write(cx, &mut self.buf.bytes()))?;
+                        let n = ready!(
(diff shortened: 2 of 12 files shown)


src/codec/framed_write.rs
```diff
@@ -193,12 +193,12 @@ where
...
                         tracing::trace!(queued_data_frame = true);
-                        let mut buf = (&mut self.buf).chain(frame.payload_mut());
-                        ready!(Pin::new(&mut self.inner).poll_write_buf(cx, &mut buf))?;
+                        let mut buf = (&mut self.buf).chain(frame.payload_mut()).to_bytes();
+                        ready!(Pin::new(&mut self.inner).poll_write(cx, &mut buf))?;
                     }
                     _ => {
                         tracing::trace!(queued_data_frame = false);
-                        ready!(Pin::new(&mut self.inner).poll_write_buf(cx, &mut self.buf))?;
+                        ready!(Pin::new(&mut self.inner).poll_write(cx, &mut self.buf.to_bytes()))?;
```
LucioFranco, line 201: Can we use https://docs.rs/bytes/0.5.6/bytes/buf/trait.Buf.html#tymethod.bytes here instead of `to_bytes`?

src/codec/framed_write.rs
```diff
@@ -193,12 +193,12 @@ where
                 match self.next {
                     Some(Next::Data(ref mut frame)) => {
                         tracing::trace!(queued_data_frame = true);
-                        let mut buf = (&mut self.buf).chain(frame.payload_mut());
-                        ready!(Pin::new(&mut self.inner).poll_write_buf(cx, &mut buf))?;
+                        let mut buf = (&mut self.buf).chain(frame.payload_mut()).to_bytes();
```
LucioFranco, line 196: After discussing with @seanmonstar I think in this case actually want to copy the frame payload onto the `buf` type.

LucioFranco: Probably via https://docs.rs/bytes/0.5.6/bytes/struct.BytesMut.html#method.extend_from_slice

src/codec/framed_write.rs
LucioFranco: We also need to advance the buffer position here since that is what `poll_write_buf` did.

tests/h2-support/src/mock.rs
```diff
@@ -344,29 +345,30 @@ impl AsyncRead for Mock {
...
 
             me.rx_task = Some(cx.waker().clone());
             return Poll::Pending;
         }
 
-        let n = cmp::min(buf.len(), me.rx.len());
-        buf[..n].copy_from_slice(&me.rx[..n]);
+        let n = cmp::min(buf.filled().len(), me.rx.len());
+        let initialized_mut = buf.initialized_mut();
+        initialized_mut[..n].copy_from_slice(&me.rx[..n]);
```
quininer, line 368: and `buf.advance(n)` ?

tests/h2-support/src/mock.rs
```diff
@@ -344,29 +345,30 @@ impl AsyncRead for Mock {
...
+                return Poll::Ready(Ok(()));
             }
 
             me.rx_task = Some(cx.waker().clone());
             return Poll::Pending;
         }
 
-        let n = cmp::min(buf.len(), me.rx.len());
-        buf[..n].copy_from_slice(&me.rx[..n]);
+        let n = cmp::min(buf.filled().len(), me.rx.len());
```
quininer, line 366:
```suggestion
        let n = cmp::min(buf.remaining(), me.rx.len());
```

tests/h2-support/src/mock.rs
```diff
@@ -344,29 +345,30 @@ impl AsyncRead for Mock {
...
+        let initialized_mut = buf.initialized_mut();
```
quininer, line 367:
```suggestion
        let initialized_mut = buf.initialize_unfilled();
```

tests/h2-support/src/mock.rs
```diff
@@ -344,29 +345,30 @@ impl AsyncRead for Mock {
     fn poll_read(
         self: Pin<&mut Self>,
         cx: &mut Context<'_>,
-        buf: &mut [u8],
-    ) -> Poll<io::Result<usize>> {
+        buf: &mut ReadBuf,
+    ) -> Poll<io::Result<()>> {
         assert!(
-            buf.len() > 0,
+            buf.filled().len() > 0,
```
quininer, line 351: Should be `buf.remaining()` ?

Urhengulas:
Hello @jxs @quininer 😄 
Is there some blocker? Can I support in any way?

Cargo.toml
```diff
@@ -66,9 +66,9 @@ serde = "1.0.0"
 serde_json = "1.0.0"
 
 # Examples
-tokio = { version = "0.2", features = ["dns", "macros", "rt-core", "sync", "tcp"] }
+tokio = { version = "0.3", features = ["rt-multi-thread", "macros", "sync", "net"] }
```
Urhengulas, line 69: Since these are the `dev-dependencies` it is probably to just enable the `full` feature, to get all the features for tests and examples.

quininer:
I think this PR can already be merged, but only needs to be fix or temporarily ignore `update_max_frame_len_at_rest` test.

Urhengulas:
> I think this PR can already be merged, but only needs to be fix or temporarily ignore `update_max_frame_len_at_rest` test.

@quininer That sounds awesome!

What is the problem with this test?

Cargo.toml
jplatte: That would increase CI times.

Urhengulas: That is true. So it's a question of convenience vs. CI-time 😉

Cargo.toml
```diff
@@ -43,14 +43,14 @@ members = [
 futures-core = { version = "0.3", default-features = false }
 futures-sink = { version = "0.3", default-features = false }
 futures-util = { version = "0.3", default-features = false }
-tokio-util = { version = "0.3.1", features = ["codec"] }
-tokio = { version = "0.2", features = ["io-util"] }
+tokio-util = { version = "0.4.0", features = ["codec"] }
```
Urhengulas, line 46: Is there a reason to specify the `patch` version for `tokio-util`? _(It's not specified for most other dependencies...)_

seanmonstar:
I found the issue with the h2spec script (debug output of `SocketAddr`s changed to no longer include `V4(blah)`, just `blah`), it's on master, if you want to pull that in.

jxs:
thanks Sean, pulled from master tests seem to be passing now

Urhengulas (approved):

seanmonstar (approved):

PR #497, opened 2020-10-26
Upgrade to bytes 0.6

paolobarbolini:
Depends on:

* [x] https://github.com/tokio-rs/tokio/pull/3052
* [ ] https://github.com/hyperium/http/pull/440

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ futures-util = { version = "0.3", default-features = false }
 tokio-util = { version = "0.5", features = ["codec"] }
 tokio = { version = "0.3.2", features = ["io-util"] }
 bytes = "0.6"
-http = { git = "https://github.com/paolobarbolini/http.git", branch = "bytes06" }
+http = "0.2"
 tracing = { version = "0.1.13", default-features = false, features = ["std", "log"] }
 tracing-futures = { version = "0.2", default-features = false, features = ["std-future"]}
 fnv = "1.0.5"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,10 +45,10 @@ members = [
 futures-core = { version = "0.3", default-features = false }
 futures-sink = { version = "0.3", default-features = false }
 futures-util = { version = "0.3", default-features = false }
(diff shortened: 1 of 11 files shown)


hawkw (approved):

Cargo.toml
```diff
@@ -45,10 +45,10 @@ members = [
...
 futures-sink = { version = "0.3", default-features = false }
 futures-util = { version = "0.3", default-features = false }
-tokio-util = { version = "0.4.0", features = ["codec"] }
-tokio = { version = "0.3", features = ["io-util"] }
-bytes = "0.5.2"
-http = "0.2"
+tokio-util = { version = "0.5", features = ["codec"] }
+tokio = { version = "0.3.2", features = ["io-util"] }
+bytes = "0.6"
+http = { git = "https://github.com/paolobarbolini/http.git", branch = "bytes06" }
```
hawkw, line 51: rather than adding a Git dependency on `http`, let's just bump it separately when the next version is published. `bytes` is a private dependency of `http`, so it's not an issue that it still uses 0.5.
```suggestion
http = "0.2"
```

tests/h2-support/Cargo.toml
```diff
@@ -7,10 +7,10 @@ edition = "2018"
...
 h2 = { path = "../..", features = ["stream", "unstable"] }
 
-bytes = "0.5"
+bytes = "0.6"
 tracing = "0.1"
 tracing-subscriber = { version = "0.2", default-features = false, features = ["fmt", "chrono", "ansi"] }
 futures = { version = "0.3", default-features = false }
-http = "0.2"
-tokio = { version = "0.3", features = ["time"] }
+http = { git = "https://github.com/paolobarbolini/http.git", branch = "bytes06" }
```
hawkw, line 14:
```suggestion
http = "0.2"
```

hawkw (changes requested):
let's avoid the git dep on `http` for now (per a Discord conversation with @seanmonstar). then, I think this should be good to merge!

hawkw (approved):
looks good to me! if ci passes, i'll go ahead and merge

PR #503, opened 2020-12-22
Make some functions less-generic to reduce binary bloat

Marwes:
Didn't see #484 until after the fact so I can remove that commit if desired, it is arguably a cleaner separation though since it extracts a separate and private struct instead of passing multiple parameters.

### DIFF
--- a/src/proto/connection.rs
+++ b/src/proto/connection.rs
@@ -123,7 +123,7 @@ where
     /// Returns the maximum number of concurrent streams that may be initiated
     /// by this peer.
     pub(crate) fn max_send_streams(&self) -> usize {
-        self.streams.max_send_streams()
+        self.inner.streams.max_send_streams()
     }
 
     /// Returns `Ready` when the connection is ready to receive a frame.
--- a/src/proto/connection.rs
+++ b/src/proto/connection.rs
@@ -65,6 +65,10 @@ struct DynConnection<'a, B: Buf = Bytes> {
     go_away: &'a mut GoAway,
 
     streams: DynStreams<'a, B>,
+
+    error: &'a mut Option<Reason>,
+
+    ping_pong: &'a mut PingPong,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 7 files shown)


seanmonstar (commented):
This looks very cool, thanks! Do you have any data on the difference this makes?

We're preparing a major (0.x minor?) release for tomorrow, so I'd prefer to merge this after, just to give a timeline.

Marwes:
Testing with and without this on hyper's http_proxy example which triggers two instantiatiations of decode_frame, etc. Gives a ~10% reduction in emitted LLVM IR

553979 / 610618 = 0.90724315366

Marwes:
Any chance of this getting merged?

nox:
@Marwes The two last commits can be removed, as #514 fixed the deprecation problem.

src/proto/connection.rs
```diff
@@ -51,6 +59,18 @@ where
     _phantom: PhantomData<P>,
 }
 
+struct DynConnection<'a, B: Buf = Bytes> {
```
seanmonstar, line 62: Dyn in Rust usually means dynamic dispatch. It looks like this is just to borrow multiple fields at once, right? Maybe instead something with the name `Ref`, like `ConnectionInnerRef`?

src/proto/connection.rs
```diff
@@ -17,6 +17,17 @@ use tokio::io::{AsyncRead, AsyncWrite};
...
+    P: Peer,
+{
+    /// Read / write frame values
+    codec: Codec<T, Prioritized<B>>,
+
+    inner: ConnectionInner<P, B>,
+}
+
+#[derive(Debug)]
+struct ConnectionInner<P, B: Buf = Bytes>
```
seanmonstar, line 30: I think most of the changes are pretty clear. In this one (and perhaps in `streams.rs`), this separation is larger and not immediately obvious. I think a small comment on the struct about the motivation would be helpful (such as which generics are we trying to keep separate).

seanmonstar (commented):
Sorry for the slow review. I've finally gotten through most of it, and this is great stuff. I'm bummed the compiler cannot do this automatically, but I get it. A few thoughts inline.

src/proto/connection.rs
Marwes: It also replaces the `P: Peer` type parameter with an enum field. There was already `peer::Dyn` so I modeled the name on that (thereby `fn as_dyn`, instead of `fn as_ref` which is confusable with `AsRef::as_ref`).

seanmonstar: Oh right, I see now. Alright, seems fine.

seanmonstar (approved):

PR #504, opened 2020-12-22
Update to Tokio and Bytes 1.0

seanmonstar:
WIP because Tokio and Bytes aren't released yet, this uses git dependencies for now which we'll need to fix before merging. But this gets the code fixes ready.


seanmonstar:
Humm, example failing because it needs a new tokio-rustls, which isn't ready yet.

carllerche (approved):

Issue #485, opened 2020-09-13
panic: CONTINUATION frame write loop; header value too big to encode


jeromegn:
Started getting the following panic this morning:

```
CONTINUATION frame write loop; header value too big to encode

Thread 0 Crashed:
0   h2                              0x5653e120b6ff      h2::codec::framed_write::FramedWrite<T>::flush (macros.rs:13)
1   h2                              0x5653e12baf84      [inlined] h2::codec::framed_write::FramedWrite<T>::poll_ready (framed_write.rs:91)
2   h2                              0x5653e12baf84      [inlined] h2::codec::Codec<T>::poll_ready (mod.rs:130)
3   h2                              0x5653e12baf84      h2::proto::streams::prioritize::Prioritize::poll_complete (prioritize.rs:504)
4   h2                              0x5653e14242f9      [inlined] h2::proto::streams::send::Send::poll_complete (send.rs:294)
5   h2                              0x5653e14242f9      h2::proto::streams::streams::Streams<T>::poll_complete (streams.rs:623)
6   h2                              0x5653e1349f8e      h2::proto::connection::Connection<T>::poll (connection.rs:218)
7   h2                              0x5653e13df365      [inlined] h2::client::Connection<T>::poll (client.rs:1238)
8   futures_core                    0x5653e13df365      [inlined] futures_core::future::TryFuture::try_poll (future.rs:83)
9   futures_util                    0x5653e13df365      [inlined] futures_util::future::try_future::into_future::IntoFuture<T>::poll (into_future.rs:31)
10  futures_util                    0x5653e13df365      futures_util::future::future::map::Map<T>::poll (map.rs:67)
11  futures_util                    0x5653e1201f92      [inlined] futures_util::future::future::Map<T>::poll (lib.rs:107)
12  futures_util                    0x5653e1201f92      [inlined] futures_util::future::try_future::MapErr<T>::poll (lib.rs:107)
13  futures_util                    0x5653e1201f92      [inlined] futures_util::future::future::FutureExt::poll_unpin (mod.rs:554)
14  futures_util                    0x5653e1201f92      futures_util::future::select::Select<T>::poll (select.rs:62)
15  hyper                           0x5653e13becc5      [inlined] hyper::proto::h2::client::conn_task::{{closure}} (client.rs:159)
16  core                            0x5653e13becc5      core::future::from_generator::GenFuture<T>::poll (mod.rs:73)
[...]

PR #506, opened 2020-12-23
Return an error instead of panicking when stuck in a CONTINUATION loop

eaufavor:
A short term mitigation for https://github.com/hyperium/h2/issues/485.

It is not rare a large header can trigger such a CONTINUATION loop.
While the stream cannot recover from this issue, returning an error
instead of panicking makes the impact radius under better control.

### DIFF
--- a/src/codec/framed_write.rs
+++ b/src/codec/framed_write.rs
@@ -247,6 +247,9 @@ where
                             // If *only* the CONTINUATION frame header was
                             // written, and *no* header fields, we're stuck
                             // in a loop...
+                            tracing::warn!(
+                                "CONTINUATION frame write loop; header value too big to encode"
+                            );
                             return Poll::Ready(Err(std::io::Error::new(
                                 std::io::ErrorKind::InvalidInput,
                                 UserError::HeaderTooBig,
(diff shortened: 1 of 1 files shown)


src/codec/framed_write.rs
```diff
@@ -247,7 +247,10 @@ where
                             // If *only* the CONTINUATION frame header was
                             // written, and *no* header fields, we're stuck
                             // in a loop...
-                            panic!("CONTINUATION frame write loop; header value too big to encode");
```
seanmonstar, line 250: Wanna leave a `debug!` (should it be `warn!`?) with a similar message, so users can at least see *why* this header was too big? Maybe it's not important...

eaufavor: Yeah, added `warn` given this issue is technically is still an unresolved bug.

seanmonstar (approved):

eaufavor:
30ca83279056390bab51358844ab5a975b49d640 the refactor breaks it. I will update soon

seanmonstar:
Yea I just noticed I did the merge wrong, sorry!

seanmonstar:
I fixed up the conflict and merged in https://github.com/hyperium/h2/commit/11229702a03626ca1463f9216d7511616711acc7. Thanks!

PR #509, opened 2021-02-03
de-generify FramedRead::decode_frame

kornelski:
This is exactly #484, but with the couple of review comments addressed.


nox:
#514 landed so if this PR is rebased the CI should become green again.

seanmonstar (approved):

Issue #512, opened 2021-02-03
expose information about negotiated connection settings


hawkw:
`h2` currently permits configuring most of the HTTP/2 protocol settings that are negotiated with the remote peer; for example, the client builder [can configure settings like maximum frame size, maximum concurrent streams, etc](https://docs.rs/h2/0.3.0/h2/client/struct.Builder.html). 

However, once a connection is established, there's no way to _access_ information about the settings that were negotiated with a remote peer, besides [information about stream flow control](https://docs.rs/h2/0.3.0/h2/struct.FlowControl.html#implementations). For example, in my particular use case, we would like to be able to record the max concurrent streams limit set by the _server_ after initiating a client connection. It would be great to add some kind of API for querying connection-level settings that were received from a remote peer.

hawkw:
I'd be happy to work on this, but I'd love to get input on API design --- cc @seanmonstar.

seanmonstar:
Seems reasonable. My first instinct is to add getters to `Connection`.

hawkw:
> Seems reasonable. My first instinct is to add getters to `Connection`.

Hmm, that was my initial thought as well. I suppose the downside is that the `Connection` type is usually spawned immediately, so users need to access that information immediately. Probably not a huge issue.

Yneth:
current implementation still does not return the remote max_send_streams.
I am not a maintainer but see two ways to solve it:
- provide state machine API for Connection when the negotiated settings are received, example:
```
let (h2, connection) = client::Builder::new().initial_max_send_streams(1000000).handshake(tcp).await?;
let settings_negotiated = connection.await.unwrap();
tokio::spawn(async move {
   settings_negotiated.await.unwrap();
});
```
- or somehow include such functionality in `h2.ready()`, example:
```
let (h2, connection) = client::Builder::new().initial_max_send_streams(1000000).handshake(tcp).await?;
tokio::spawn(async move {
   connection.await.unwrap();
});
let mut h2 = h2.ready().await?;
// in current implementation, max_send_streams may not be populated
// adding tokio::time::sleep() helps, but it is ugly
h2.max_send_streams();
```

LucioFranco:
@Yneth maybe it makes sense to use ready as a signal that the data is ready then use global getters that return an `Option<&ConfigItem>`. Could maybe use something like a oncecell?

PR #513, opened 2021-02-04
add `Connection::max_concurrent_send_streams`

hawkw:
This PR adds accessors to `client::Connection` and `server::Connection`
that return the send stream concurrency limit on that connection, as
negotiated by the remote peer. This is part of issue #512.

I think we probably ought to expose similar accessors for other
settings, but I thought it was better to add each one in a separate,
focused PR.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1231,6 +1231,13 @@ where
 
     /// Returns the maximum number of concurrent streams that may be initiated
     /// by this client.
+    ///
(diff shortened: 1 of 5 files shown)


seanmonstar (approved):

hawkw:
added a bit more to the docs, including a reference to the RFC.

Yneth:
it does not return remote settings

PR #516, opened 2021-02-05
add `Connection::max_concurrent_recv_streams`

hawkw:
This commit adds accessors to `client::Connection` and
`server::Connection` that return the current value of the
`SETTINGS_MAX_CONCURRENT_STREAMS` limit that has been sent by this peer
and acknowledged by the remote.

This is analogous to the `max_concurrent_send_streams` methods added in
PR #513. These accessors may be somewhat less useful than the ones for
the values negotiated by the remote, since users who care about this
limit are probably setting the builder parameter. However, it seems
worth having for completeness sake --- and it might be useful for
determining whether or not a configured concurrency limit has been acked
yet...

Part of #512

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1242,13 +1242,28 @@ where
     /// by this client.
     ///
     /// This limit is configured by the server peer by sending the
-    /// [`SETTINGS_MAX_CONCURRENT_STREAMS` parameter][settings] in a `SETTINGS`
-    /// frame. This method returns the currently acknowledged value recieved
-    /// from the remote.
+    /// [`SETTINGS_MAX_CONCURRENT_STREAMS` parameter][1] in a `SETTINGS` frame.
+    /// This method returns the currently acknowledged value recieved from the
+    /// remote.
     ///
     /// [settings]: https://tools.ietf.org/html/rfc7540#section-5.1.2
     pub fn max_concurrent_send_streams(&self) -> usize {
         self.inner.max_send_streams()
+    }
+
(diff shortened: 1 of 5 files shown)


src/client.rs
```diff
@@ -1233,14 +1233,29 @@ where
     /// by this client.
     ///
     /// This limit is configured by the server peer by sending the
-    /// [`SETTINGS_MAX_CONCURRENT_STREAMS` parameter][settings] in a `SETTINGS`
-    /// frame. This method returns the currently acknowledged value recieved
-    /// from the remote.
+    /// [`SETTINGS_MAX_CONCURRENT_STREAMS` parameter][1] in a `SETTINGS` frame.
```
seanmonstar, line 1236: I think this `[1]` change is half-way. The `[settings]: ....` below needs to be updated as well.

seanmonstar (approved):
LGTM, just 1 docs fix needed.

src/client.rs
hawkw: oh whoops --- good catch!

src/client.rs
```diff
@@ -1233,14 +1233,29 @@ where
...
+    /// This method returns the currently acknowledged value recieved from the
+    /// remote.
     ///
     /// [settings]: https://tools.ietf.org/html/rfc7540#section-5.1.2
```
hawkw, line 1240:
```suggestion
    /// [1]: https://tools.ietf.org/html/rfc7540#section-5.1.2
```

PR #519, opened 2021-02-12
Explicitly enable the `std` feature of indexmap

Kestrer:
This crate depends on it anyway, and by explicitly turning it on we avoid unreliable platform target detection that causes build failures on some platforms.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ http = "0.2"
 tracing = { version = "0.1.21", default-features = false, features = ["std"] }
 fnv = "1.0.5"
 slab = "0.4.2"
-indexmap = { version = "1.0", features = ["std"] }
+indexmap = { version = "1.5.2", features = ["std"] }
 
 [dev-dependencies]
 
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -50,7 +50,7 @@ http = "0.2"
 tracing = { version = "0.1.21", default-features = false, features = ["std"] }
 fnv = "1.0.5"
 slab = "0.4.2"
-indexmap = "1.0"
+indexmap = { version = "1.0", features = ["std"] }
```
taiki-e, line 53:
```suggestion
indexmap = { version = "1.5.2", features = ["std"] }
```

this feature was added in indexmap 1.5.2

Kestrer: Ah sorry, I missed that. I've fixed it now.

taiki-e (approved):

seanmonstar:
Thanks! I'll merge this since it doesn't seem harmful and should fix some people.

I believe indexmap's change to make `std` a new cargo feature, default off unless autocfg detects something, was a breaking change (hence there's the various issues about it breaking people). Perhaps not important that I grumble, but I see textbox and thus do so.

seanmonstar (approved):

PR #521, opened 2021-02-18
Ignore 1xx frames

### LINKED ISSUE #515: Ignore 1xx frames
Currently when hyper encounters such frames it resets the stream with an error. The plan for now is to just ignore these informational frames. With the next step being an implementation of API to handle them in h2 and hyper (tracked in #167).

kornelski:
Fixes #515

This ignores (drops) header frames when the status is 1xx. 

Currently it allows multiple 1xx responses in a row. I couldn't find whether this is forbidden, but I've left it allowed, as I can imagine 103 early hints being followed by 100 continue.

There's no API to retrieve the data from the 1xx frames. Such API could look similar to push promises, but the existing push promise implementation is non-trivial, so I'll leave it for later.

### DIFF
--- a/src/frame/headers.rs
+++ b/src/frame/headers.rs
@@ -254,6 +254,11 @@ impl Headers {
         &mut self.header_block.pseudo
     }
 
+    /// Whether it has status 1xx
+    pub(crate) fn is_informational(&self) -> bool {
+        self.header_block.pseudo.is_informational()
+    }
+
     pub fn fields(&self) -> &HeaderMap {
         &self.header_block.fields
     }
@@ -598,6 +603,12 @@ impl Pseudo {
 
     pub fn set_authority(&mut self, authority: BytesStr) {
         self.authority = Some(authority);
+    }
+
+    /// Whether it has status 1xx
+    pub(crate) fn is_informational(&self) -> bool {
+        self.status
+            .map_or(false, |status| status.is_informational())
     }
 }
 
--- a/src/proto/streams/state.rs
+++ b/src/proto/streams/state.rs
@@ -2,7 +2,7 @@ use std::io;
 
 use crate::codec::UserError::*;
 use crate::codec::{RecvError, UserError};
-use crate::frame::Reason;
+use crate::frame::{self, Reason};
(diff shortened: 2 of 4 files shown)


seanmonstar:
> There's no API to retrieve the data from the 1xx frames. 

What do you mean with this? Just that currently this simply ignores them, without a public API to inspect that a 1xx response was received?

kornelski:
Yes. The client has no way of knowing what was in the 1xx frames. This change is necessary to make h2 work with servers that send 103 Early Hints, but not sufficient for clients to make use of these hints.

inikulin:
@seanmonstar sorry for poking you, but is any additional work required on the PR?

src/frame/headers.rs
```diff
@@ -599,6 +604,13 @@ impl Pseudo {
     pub fn set_authority(&mut self, authority: BytesStr) {
         self.authority = Some(authority);
     }
+
+    /// Whether it has status 1xx
+    pub(crate) fn is_informational(&self) -> bool {
+        self.status.map_or(false, |status| {
+            status.as_u16() >= 100 && status.as_u16() < 200
```
seanmonstar, line 611: This could be `status.is_informational()`.

src/proto/streams/state.rs
```diff
@@ -132,8 +132,10 @@ impl State {
 
     /// Opens the receive-half of the stream when a HEADERS frame is received.
     ///
+    /// is_informational: whether received a 1xx status code
+    ///
     /// Returns true if this transitions the state to Open.
-    pub fn recv_open(&mut self, eos: bool) -> Result<bool, RecvError> {
+    pub fn recv_open(&mut self, eos: bool, is_informational: bool) -> Result<bool, RecvError> {
```
seanmonstar, line 138: I start to worry when methods get multiple boolean arguments, it can be confusing at the call site in my opinion. What if we changed this to pass `&frame::Headers`?

seanmonstar (commented):
Cool, I think this looks good to start!

seanmonstar (approved):

PR #529, opened 2021-04-10
Initial oss-fuzz integration.

DavidKorczynski:
Hi all,

I have been working on getting hyperium-related repositories into oss-fuzz https://github.com/google/oss-fuzz/pull/5330 In this context I added a libfuzzer-based fuzzer to this project. Let me know if you are interested in this!


seanmonstar:
There's a bit of fuzzing in the crate already, around the hpack implementation. Would it be worth just plugging that in?

DavidKorczynski:
Are you referring to the code here? https://github.com/hyperium/h2/tree/master/tests/h2-fuzz

I think this is a bit outdated - it doesn't seem to compile as the `AsyncRead` has been updated. However, I think it would be nice to update it and migrate.

seanmonstar:
Oh that would be a good thing to fix too!

I was thinking of this code: https://www.github.com/hyperium/h2/tree/master/src/hpack/test/fuzz.rs

DavidKorczynski:
Sounds good - I will be taking a look at this!

DavidKorczynski:
The PR now does the following: 
- adds libfuzzer-fuzzer set up
- adds three fuzzers 
- fixes the existing honggfuzz-fuzzer so it is up to date with tokio

DavidKorczynski:
Ready for another review @seanmonstar

fuzz/fuzz_targets/fuzz_client.rs
```diff
@@ -0,0 +1,32 @@
...
+
+async fn fuzz_entry(inp: HttpSpec) {
+    if let Ok(req) = Request::builder()
+        .uri(&inp.uri[..])
+        .header(&inp.header_name[..], &inp.header_value[..])
+        .body(())
+    {
+        let (io, mut _srv) = mock::new();
+        let (mut client, _h2) = client::Builder::new()
+            .initial_stream_id(::std::u32::MAX >> 1)
```
seanmonstar, line 20: I don't think this is really needed (I mean, unless you wanted to fuzz the handling of this option, but I'm not worried about that really...).

DavidKorczynski: Thanks - should be fixed now!

seanmonstar (approved):

seanmonstar:
Thanks so much, this is excellent!

PR #534, opened 2021-04-26
Don't override empty path for CONNECT requests

nox:


seanmonstar (approved):

seanmonstar:
Does this need a test? Is there a similar one for OPTIONS?

nox:
> Does this need a test? Is there a similar one for OPTIONS?

I've made the existing CONNECT test use the helper that used to trigger this error without my patch.

seanmonstar (approved):

PR #542, opened 2021-05-05
Store buffered data size as usize (fixes #269)

### LINKED ISSUE #269: unguarded addition can result in an overflow.
I managed to trigger an overflow on [this](https://github.com/carllerche/h2/blob/51f6a124546fe624c05a713566691df928f29964/src/proto/streams/prioritize.rs#L159) line.

This can be done by sending data in a tight loop.

nox:


seanmonstar (approved):

seanmonstar:
Does it actually fix #269? I'd think that it's still possible to overflow this if you buffer a few bufs with combined space over a `usize`. Especially on 32-bit.

nox:
Probably but this patch still helps, I'll change the commit message though.

PR #543, opened 2021-05-05
Export the default value for the max number of reset streams.

boxdot:
To be used in https://github.com/hyperium/hyper/pull/2535.


boxdot:
Converted to draft as long https://github.com/hyperium/hyper/pull/2535#discussion_r626986699 is not resolved.

boxdot:
Decided in the linked review to use the default value implicitly from here.

PR #544, opened 2021-05-06
Read body in the example server

nox:


kornelski:
Whoa, `poll_closed` is a gotcha. Would it be possible to detect this error from `h2` somehow?

nox:
I don't think so, the connection isn't gone or anything, it's just that no one is polling it anymore, ever.

nox:
The docs could be improved and a proper future on which to await could be introduced though. `Connection::graceful_shutdown` exists but it still requires the connection to be polled until `Poll::Ready` afterwards, which is unfortunate IMO.

examples/server.rs
```diff
@@ -27,18 +27,33 @@ async fn handle(socket: TcpStream) -> Result<(), Box<dyn Error + Send + Sync>> {
...
+    let (mut request, mut respond) = match connection.accept().await {
+        Some(result) => result?,
+        None => {
+            println!("Didn't get a request");
+            return Ok(());
+        }
+    };
+
+    tokio::spawn(async move {
+        let _ = futures_util::future::poll_fn(|cx| connection.poll_closed(cx)).await;
```
seanmonstar, line 39: I think this change will only allow a single request, right? I'd probably say we should change the example to just _always_ spawn a task to handle the request. Something like:

```rust
while let Some(result) = connection.accept().await {
    tokio::spawn(handle_request(result));
}
```

nox: That is true.

nox: I've made it use `tokio::spawn` instead.

Btw the window stuff seems weirdly handled:

```
<<<< recv 16384
<<<< recv 16384
<<<< recv 16384
<<<< recv 16383
<<<< recv 1
<<<< recv 16384
<<<< recv 16384
<<<< recv 1696
```

seanmonstar (approved):
Looks good to me! I dunno if you're still tweaking something, but merge when done :)

nox:
> I dunno if you're still tweaking something

You know how it is, you think you are done and then the style lint complains.

PR #547, opened 2021-06-24
fuzz: fix e2e MockIo reading too much for the buffer

seanmonstar:


seanmonstar:
Meh, breakage on nightly from: https://github.com/rust-lang/rust/issues/86369

PR #552, opened 2021-08-20
Remove the ad-hoc header size limit

nox:
This limit is hardcoded to be 16KB-100B no matter if the other
peer already advertised that it accepts way larger frame sizes,
this is a problem sometimes for Google Chat's service workers.


nox:
Turns out this won't help and I need to implement actual header arbitrary splitting.

PR #553, opened 2021-08-20
Fix potential hang if extensions contain same StreamRef

seanmonstar:
If a user stored a `StreamRef` to the same stream in the request or
response extensions, they would be dropped while the internal store lock
was held. That would lead to a deadlock, since dropping a stream ref
will try to take the store lock to clean up.

Clear extensions of Request and Response before locking store, prevent
this.

Fixes hyperium/hyper#2621


davidpdrsn (approved):
:shipit:

PR #555, opened 2021-08-23
Support very large headers

nox:
This completely refactors how headers are hpack-encoded.

Instead of trying to be clever, constructing frames on the go
while hpack-encoding, we just make a blob of all the
hpack-encoded headers first, and then we split that blob
in as many frames as necessary.

### DIFF
--- a/src/codec/error.rs
+++ b/src/codec/error.rs
@@ -35,9 +35,6 @@ pub enum UserError {
     /// The payload size is too big
     PayloadTooBig,
 
-    /// A header size is too big
-    HeaderTooBig,
-
     /// The application attempted to initiate too many streams to remote.
     Rejected,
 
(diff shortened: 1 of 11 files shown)


ipostelnik:
Where does it check that total header size doesn't exceed SETTINGS_MAX_HEADER_LIST_SIZE?

nox:
Pretty sure it doesn't check that when encoding, note that it didn't beforehand either.

seanmonstar (commented):
This looks nice and clean to me! Perhaps a test in `h2-tests/codec_write` about sending a header that would have been too big before?

LPardue:
There's a risk with enforcing SETTINGS_MAX_HEADER_LIST_SIZE when sending. The input may come from some other component or system that doesn't have such limits, and/or may not be able to change they header list to fit that size. An H2 library that enforces the rule can cause problems, so if it's ever implemented it probably needs to be optional.

nox:
> This looks nice and clean to me! Perhaps a test in `h2-tests/codec_write` about sending a header that would have been too big before?

I added a 32KiB header in the existing test.

seanmonstar (approved):

PR #556, opened 2021-08-30
Refactor errors (fixes #530)

### LINKED ISSUE #530: Redesign the h2::Error type
> Yea, I think we'd need to redesign the `h2::Error` type some, so that it can include if it's a stream error, or a connection error (`GOAWAY`). Then we'd be better equipped to answer that programmatically.

_Originally posted by @seanmonstar in https://github.com/hyperium/hyper/issues/2500#issuecomment-821585355_

AFAICT we want `DynConnection::error` and `ConnectionInner::error` to at the very least be an `Option<frame::GoAway>` instead of a plain `Option<Reason>`.

nox:
h2::Error now knows whether protocol errors happened because the user
sent them, because it was received from the remote peer, or because
the library itself emitted an error because it detected a protocol
violation.

It also keeps track of whether it came from a RST_STREAM or GO_AWAY
frame, and in the case of the latter, it includes the additional
debug data if any.

### DIFF
--- a/src/codec/framed_read.rs
+++ b/src/codec/framed_read.rs
@@ -1,8 +1,8 @@
-use crate::codec::RecvError;
 use crate::frame::{self, Frame, Kind, Reason};
 use crate::frame::{
     DEFAULT_MAX_FRAME_SIZE, DEFAULT_SETTINGS_HEADER_TABLE_SIZE, MAX_MAX_FRAME_SIZE,
 };
+use crate::proto::Error;
 
 use crate::hpack;
 
@@ -98,8 +98,7 @@ fn decode_frame(
     max_header_list_size: usize,
     partial_inout: &mut Option<Partial>,
     mut bytes: BytesMut,
-) -> Result<Option<Frame>, RecvError> {
-    use self::RecvError::*;
+) -> Result<Option<Frame>, Error> {
     let span = tracing::trace_span!("FramedRead::decode_frame", offset = bytes.len());
     let _e = span.enter();
 
@@ -110,7 +109,7 @@ fn decode_frame(
 
     if partial_inout.is_some() && head.kind() != Kind::Continuation {
         proto_err!(conn: "expected CONTINUATION, got {:?}", head.kind());
-        return Err(Connection(Reason::PROTOCOL_ERROR));
+        return Err(Error::library_go_away(Reason::PROTOCOL_ERROR).into());
     }
 
     let kind = head.kind();
@@ -131,14 +130,11 @@ fn decode_frame(
                     // A stream cannot depend on itself. An endpoint MUST
                     // treat this as a stream error (Section 5.4.2) of type
                     // `PROTOCOL_ERROR`.
-                    return Err(Stream {
-                        id: $head.stream_id(),
-                        reason: Reason::PROTOCOL_ERROR,
-                    });
+                    return Err(Error::library_reset($head.stream_id(), Reason::PROTOCOL_ERROR));
                 },
                 Err(e) => {
                     proto_err!(conn: "failed to load frame; err={:?}", e);
-                    return Err(Connection(Reason::PROTOCOL_ERROR));
+                    return Err(Error::library_go_away(Reason::PROTOCOL_ERROR));
                 }
             };
 
@@ -151,14 +147,11 @@ fn decode_frame(
                 Err(frame::Error::MalformedMessage) => {
                     let id = $head.stream_id();
                     proto_err!(stream: "malformed header block; stream={:?}", id);
-                    return Err(Stream {
-                        id,
-                        reason: Reason::PROTOCOL_ERROR,
-                    });
+                    return Err(Error::library_reset(id, Reason::PROTOCOL_ERROR));
                 },
                 Err(e) => {
                     proto_err!(conn: "failed HPACK decoding; err={:?}", e);
-                    return Err(Connection(Reason::PROTOCOL_ERROR));
+                    return Err(Error::library_go_away(Reason::PROTOCOL_ERROR));
                 }
             }
 
@@ -183,7 +176,7 @@ fn decode_frame(
 
             res.map_err(|e| {
                 proto_err!(conn: "failed to load SETTINGS frame; err={:?}", e);
-                Connection(Reason::PROTOCOL_ERROR)
+                Error::library_go_away(Reason::PROTOCOL_ERROR)
             })?
             .into()
         }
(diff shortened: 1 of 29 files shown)


src/error.rs
```diff
@@ -23,10 +23,11 @@ pub struct Error {
 #[derive(Debug)]
 enum Kind {
     /// An error caused by an action taken by the remote peer.
-    ///
-    /// This is either an error received by the peer or caused by an invalid
-    /// action taken by the peer (i.e. a protocol error).
-    Proto(Reason),
+    Remote(Reason),
+
+    /// An error caused by an invalid  action taken by the peer
```
LPardue, line 28:
```suggestion
    /// An error caused by an invalid action taken by the peer
```

src/codec/framed_read.rs
```diff
@@ -110,7 +110,7 @@ fn decode_frame(
 
     if partial_inout.is_some() && head.kind() != Kind::Continuation {
         proto_err!(conn: "expected CONTINUATION, got {:?}", head.kind());
```
ipostelnik, line 112: is it possible to get this information into the error? `PROTOCOL_ERROR` is very generic and we lose details of  what specifically went wrong.

LPardue: PROTOCOL_ERROR is used a lot throughout RFC 7540. It sounds like you'd like internal errors to contain more detail than the protocol otherwise defines, e.g. by requiring two error types.

seanmonstar: A `GOAWAY` frame can include additional opaque debug data, which we can use to give a hint as to what the peer did. I think it'd be a good idea in most cases.

LPardue: Any additional information included in a GOAWAY frame is a potential security problem because the peer is untrusted and can probe things. Furthermore, data in the GOAWAY could be toxic, and discarded by the endpoints that receive it (see last para in https://datatracker.ietf.org/doc/html/rfc7540#section-6.8). Furthermore, some kinds of errors are specific to streams, and RST_STREAM can't carry any debug data.

So one school of thought is that most cases of error can be determined fairly straightforwardly by someone looking at the frames that were sent. A consistent error occurrence is usually strong enough grounds to cause someone to start taking a closer look at things. That's not to say making information available locally doesn't have merit, just don't rely on the GOAWAY debug data.

seanmonstar: Yea, all true. It's a balance, just like when returning 4xx errors from a server, you might include a message hinting at what the client did wrong, or you may decide you shouldn't. :shrug:

LPardue: One might be tempted to make the behaviour configurable, but that throws even more complication in :smile:

nox:
Seems like there are no tests for resets, will check that out.

src/proto/go_away.rs
```diff
@@ -12,35 +12,13 @@ pub(super) struct GoAway {
...
+    going_away: Option<frame::GoAway>,
     /// Whether the user started the GOAWAY by calling `abrupt_shutdown`.
     is_user_initiated: bool,
     /// A GOAWAY frame that must be buffered in the Codec immediately.
     pending: Option<frame::GoAway>,
 }
 
-/// Keeps a memory of any GOAWAY frames we've sent before.
-///
-/// This looks very similar to a `frame::GoAway`, but is a separate type. Why?
```
seanmonstar, line 24: Didn't think this was worth keeping? :D

src/proto/streams/prioritize.rs
```diff
@@ -791,7 +791,10 @@ impl Prioritize {
                         }),
                         None => {
                             if let Some(reason) = stream.state.get_scheduled_reset() {
-                                stream.state.set_reset(reason);
+                                let stream_id = stream.id;
```
seanmonstar, line 794: Isn't this redundant? The stream already knows it's ID?

src/error.rs
```diff
@@ -122,12 +130,20 @@ impl From<UserError> for Error {
...
-
         match self.kind {
-            Proto(ref reason) => write!(fmt, "protocol error: {}", reason),
-            User(ref e) => write!(fmt, "user error: {}", e),
-            Io(ref e) => fmt::Display::fmt(e, fmt),
+            Kind::Reset(_, reason, initiator) => {
+                write!(fmt, "stream reset by {}: {}", initiator, reason)
+            }
+            Kind::GoAway(ref debug_data, reason, initiator) => {
+                write!(fmt, "go away from {}: {}", initiator, reason)?;
```
seanmonstar, line 138: I'm not sure my thought is useful, but I figured I'd list it:

I kinda like `"connection error detected: flow control error"` and `"connection error received (goaway): flow control error"`. I suppose it might be less clear, but at the same time it feels like it makes more sense from the perspective of a user who doesn't know what a "reset" or "go away" is... 🤷

seanmonstar (commented):
Nice work! I think this is a big improvement to help identify where an error actually came from. Some notes inline, some important, some just rambling at a cloud.

src/proto/streams/prioritize.rs
nox: Do you think it is important to remove that `stream_id` from there?

I think I could, if I inline `proto::Error` variants into `streams::state::Cause`, but honestly I don't think it is worth it.

nox:
The errors are fixed by https://github.com/hyperium/h2/pull/557.

src/proto/streams/prioritize.rs
seanmonstar: I don't think it's a big deal. I was just noting it, in case there was a better way that popped into your mind.

src/error.rs
```diff
@@ -122,12 +130,20 @@ impl From<UserError> for Error {
...
+                if !debug_data.is_empty() {
+                    write!(fmt, " ({:?})", debug_data)?;
```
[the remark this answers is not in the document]
seanmonstar: Hm. Alternate output might be a good use here... Do you know what other HTTP/2 libraries do with the debug-data? Like netty, or grpc, or other...?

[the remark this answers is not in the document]
nox: [Netty](https://github.com/netty/netty/blob/201f6b6bb2dcdff07df6e30552b96fedbcf21759/codec-http2/src/main/java/io/netty/handler/codec/http2/StreamBufferingEncoder.java#L84-L112) doesn't format the related exception at all.

[grpc-java](https://github.com/grpc/grpc-java/blob/0838b736748b398292a696d2444f9aa0f6778081/netty/src/main/java/io/grpc/netty/NettyClientHandler.java#L282) seems to log debug data directly though.

IMO we should just always include it in Display otherwise most people won't benefit from it.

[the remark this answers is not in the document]
nox: @seanmonstar Let's merge it? 🥺

seanmonstar (approved):

PR #562, opened 2021-09-30
Fix Error::reason regression

### LINKED ISSUE #561: Regression: server no longer propagates reset codes
When [upgrading Linkerd from 0.3.4 to 0.3.5](https://github.com/linkerd/linkerd2-proxy/pull/1292), we hit a test error indicating that h2 no longer propagates reset codes.

For example, we see the log lines indicating that the server handles an `ENHANCE_YOUR_CALM` reset code and then immediately refers to it as an `INTERNAL_ERROR`:
```
[     0.016469s]  WARN test_server{version=Http2 addr=127.0.0.1:38789 test=tests::transparency::http2_rst_stream_is_propagated}:conn{addr=127.0.0.1:44188}: hyper::proto::h2::server: http2 service errored: error from user's Service: protocol error: detected excessive load generating behavior
[     0.016503s] TRACE test_server{version=Http2 addr=127.0.0.1:38789 test=tests::transparency::http2_rst_stream_is_propagated}:conn{addr=127.0.0.1:44188}: h2::proto::streams::send: send_reset(..., reason=INTERNAL_ERROR, initiator=User, stream=StreamId(1), ..., is_reset=false; is_closed=false; pending_send.is_empty=true; state=State { inner: HalfClosedRemote(AwaitingHeaders) } 
```

cc @hawkw

seanmonstar:
The `Error::reason()` method regressed by not checking if the internal kind was a user-created error. The second commit fixes this. The first commit was adding a test about `send_reset`, which didn't actually end up being the problem, but we didn't have an explicit test, so included it anyways.

Fixes #561

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -50,7 +50,9 @@ impl Error {
     /// action taken by the peer (i.e. a protocol error).
     pub fn reason(&self) -> Option<Reason> {
         match self.kind {
-            Kind::Reset(_, reason, _) | Kind::GoAway(_, reason, _) => Some(reason),
+            Kind::Reset(_, reason, _) | Kind::GoAway(_, reason, _) | Kind::Reason(reason) => {
+                Some(reason)
+            }
             _ => None,
         }
     }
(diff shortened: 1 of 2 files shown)


olix0r (approved):
thanks for the quick fix @seanmonstar!

nox:
Oh sorry for that, I added `Reason` after I was entirely done with the patch and completely forgot to patch that method.

PR #565, opened 2021-10-19
Implement the extended CONNECT protocol from RFC 8441

nox:


nox:
I could make `Protocol` own a `HeaderValue` instead of a `BytesStr`, but I don't really see the point of doing that given all other pseudo-headers are represented with a `BytesStr`, and making this change would also require a `set_pseudo!` refactor.

nox:
I've decided to not implement new user error cases on the client side given there are none for normal connect requests either anyway.

src/ext.rs
```diff
@@ -0,0 +1,55 @@
+//! Extensions specific to the HTTP/2.0 protocol.
```
LPardue, line 1:
```suggestion
//! Extensions specific to the HTTP/2 protocol.
```

src/proto/settings.rs
```diff
@@ -117,21 +117,22 @@ impl Settings {
...
             }
-
-            streams.apply_remote_settings(settings)?;
         }
 
         self.remote = None;
 
         match &self.local {
             Local::ToSend(settings) => {
+                tracing::trace!("AM TRYING: {:?}", settings);
```
LPardue, line 135: is this intended to be left in?

src/server.rs
```diff
@@ -1410,6 +1431,11 @@ impl proto::Peer for Peer {
             malformed!("malformed headers: missing method");
         }
 
+        let has_protocol = pseudo.protocol.is_some();
+        if !is_connect && has_protocol {
+            malformed!("malformed headers: missing method");
```
LPardue, line 1436: I think you want a different message

seanmonstar (approved):
I think this looks great! Everything looks taken care of, all the situations have tests. Impressive!

seanmonstar:
I'll leave merging up to you if there's anything you want to finish.

PR #574, opened 2021-11-09
Expose more h2 error kinds

eaufavor:
When handling incoming h2 errors, typically we pay attention to 3 types of
them: IO, GOAWAY and RST_STREAM. This change adds the other 2 besides is_io()
so that the error can be programatically handled.


eaufavor:
#593 and #618 covered this change.

PR #576, opened 2021-11-17
Make use of NLL to clean up handshaking logic

nox:


src/server.rs
```diff
@@ -1226,62 +1227,57 @@ where
         let span = self.span.clone(); // XXX(eliza): T_T
```
hawkw, line 1227: it could be worth seeing if we can remove the `clone` call now --- IIRC, this was necessary because we borrowed all of `self` previously, which may no longer be the case here?

src/server.rs
```diff
@@ -1497,42 +1493,9 @@ where
     #[inline]
     fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
         match *self {
-            Handshaking::Flushing(_) => write!(f, "Handshaking::Flushing(_)"),
-            Handshaking::ReadingPreface(_) => write!(f, "Handshaking::ReadingPreface(_)"),
-            Handshaking::Empty => write!(f, "Handshaking::Empty"),
+            Handshaking::Flushing(_) => write!(f, "Flushing(_)"),
+            Handshaking::ReadingPreface(_) => write!(f, "ReadingPreface(_)"),
+            Handshaking::Done => write!(f, "Done"),
```
hawkw, line 1498: minor nit: these could probably be changed to 
```suggestion
            Handshaking::Flushing(_) => f.write_str("Flushing(_)"),
            Handshaking::ReadingPreface(_) =>  f.write_str("ReadingPreface(_)"),
            Handshaking::Done =>  f.write_str("Done"),
```
the `write!` macro is unnecessary, and if we just used `write_str`, we could avoid some macro expansions, which might have a tiny effect on build time or something. but, it also doesn't really matter either way --- just maybe worth doing while we're changing this code.

I dunno why I ever used `write!` here --- probably, I just didn't know better at the time :)

hawkw (approved):
yup, this definitely looks much nicer than the pre-NLL code! after reading over this code, there doesn't appear to be any change in behavior, so this looks great to me!

i left a couple small suggestions we might want to address while we're cleaning this up, but they're not really blockers.

seanmonstar (approved):

src/server.rs
hawkw: sorry, i was wrong about this, looks like this broke the build: https://github.com/hyperium/h2/runs/4884456490?check_suite_focus=true

PR #577, opened 2021-11-19
Implement server::Builder::max_send_buffer_size

nox:


nox:
I tried using `wait_for_capacity` to write a test but I couldn't understand how to use it in conjunction with `.drive` on the server side.

src/proto/streams/prioritize.rs
```diff
@@ -630,6 +654,10 @@ impl Prioritize {
             tracing::trace!(?frame, "dropping");
         }
 
+        // FIXME(nox): Do we need to wake something up when the current send
+        // buffer size was equal to max send buffer size and decreased?
```
seanmonstar, line 658: Yes, since a stream may have been parked waiting for "capacity" while the send buffer was "full". In other places in this file, that is done by calling `assign_connection_capacity()`. It would be wrong to call that unmodified, since more room in the buffer doesn't mean the flow control capacity should be changed. But perhaps a similar `assign_connection_buffer()` that does similar work would be the right call.

seanmonstar: Maybe calling it with 0 is enough?

nox: Well, this comment is kinda outdated. It is in `Prioritize::clear_queue`, which has only 2 callers:
* `Send::send_reset`, which then unconditionally calls `Prioritize::reclaim_all_capacity`;
* `Send::handle_error`, which also unconditionally calls `Prioritize::reclaim_all_capacity`.

Then, `Prioritize::reclaim_all_capacity` unconditionally calls `Prioritize::assign_connection_capacity`.

I can't find the path on which I forgot to add a call to `assign_connection_capacity`.

src/proto/streams/send.rs
```diff
@@ -335,7 +339,10 @@ impl Send {
         if available as usize <= buffered {
             0
         } else {
-            available - buffered as WindowSize
+            cmp::min(
+                available - buffered as WindowSize,
+                self.prioritize.available(),
+            )
```
nox, line 345: The bug was using `self.prioritize.available()` here.

The capacity assigned to any given stream was already capped by the currently available buffer size, and the connection-level capacity was decreased accordingly, so we end up applying a min with 0.

nox:
@seanmonstar Do you think h2 should have a default maximum?

nox:
Also, do you have any idea how I could write a test for this?

src/proto/streams/prioritize.rs
```diff
@@ -175,6 +189,8 @@ impl Prioritize {
             self.try_assign_capacity(stream);
         }
 
+        self.current_send_buffer_size += sz as usize;
```
nox, line 192: I used to do this prior to calling `try_assign_capacity`. AFAIU that was wrong, right?

seanmonstar:
> Do you think h2 should have a default maximum?

Hm, probably yes.

Also, it might be good (once we have it working) to compare how this change affects hyper's http2 `end_to_end` benchmarks, just so we know.

src/proto/streams/prioritize.rs
seanmonstar: That sounds right. The connection-level and stream-level should be updated together, but since updating the stream-level may make it exceed it's requested capacity and thus needs to check for more capacity, that must happen in-between.

A code comment about the ordering being important may be prudent.

src/proto/streams/prioritize.rs
```diff
@@ -736,6 +765,8 @@ impl Prioritize {
                             tracing::trace_span!("updating stream flow").in_scope(|| {
                                 stream.send_flow.send_data(len);
 
+                                self.current_send_buffer_size -= len as usize;
```
seanmonstar, line 768: I think this area needs to potentially wake up any tasks that have been waiting on "capacity", since now there is "more buffer space to use". The reason it needs to be done here, is because with flow control capacity, the connection doesn't actually get more until the peer has sent another WINDOW_UPDATE frame. So `recv_window_update` will notify a task there. But buffer size isn't affected by WINDOW_UPDATE frames.

nox: I'm not sure I understand. Why do we even call `self.flow.assign_capacity(len);` below that line, if we are actually supposed to wait for the peer to send another WINDOW_UPDATE frame?

nox: I am even more consumed that we call `self.flow.send_data(len);` immediately afterwards we called `self.flow.assign_capacity(len);`.

nox: Doesn't that mean that `self.flow.available` never changes from `Prioritize::pop_frame`? I don't understand the logic there.

nox: I called `assign_connection_capacity` from `poll_complete` after the `reclaim_frame` call.

tests/h2-tests/tests/flow_control.rs
```diff
@@ -1561,3 +1561,57 @@ async fn data_padding() {
...
+            .expect("handshake");
+
+        let (_req, mut reply) = srv.next().await.unwrap().unwrap();
+        tokio::spawn(async move {
+            let rsp = http::Response::new(());
+            let mut stream = reply.send_response(rsp, false).unwrap();
+
+            for _ in 0..3 {
+                stream.reserve_capacity(5);
+                stream = util::wait_for_capacity(stream, 5).await;
```
nox, line 1605: The test is hanging there, and by the time h2 returns `Poll::Pending`, `Actions::task` is `None` so there is no way to wake up the connection task to call `poll_complete` from there.

tests/h2-tests/tests/flow_control.rs
```diff
@@ -1561,3 +1561,86 @@ async fn data_padding() {
...
+                eprintln!("[t2] got 1 cap, loop {}", n);
+                stream2.send_data("abcde".into(), false).unwrap();
+            }
+
+            stream2.send_data("".into(), true).unwrap();
+        });
+
+        t2.await.expect("srv body spawn");
+        t1.await.expect("srv body spawn");
+        t0.await.expect("srv end");
```
nox, line 1642: AFAIK the test is still wrong. We await for `t2` before we await for `t0`, so `t0` is pending and nothing will wake it up when `t2` notifies the connection task that it has frames to send, so the send buffer size never decreases and everything gets stuck.

nox: Never mind I misread. That being said, the test really doesn't help much finding where the issue with my patch is :/

nox: https://github.com/hyperium/h2/blob/2c53d600989209e263c8157df6e6e2280a9a3355/tests/h2-support/src/util.rs#L56-L66

There, `poll_capacity` returns `Ready(Some(0))`, and `capacity` returns `0`, so `act >= self.target` is `false`, and `Poll::Pending` gets returned even though nothing has been set up to wake up the task.

What do you think is wrong here? That `poll_capacity` returns `Ready(Some(0))`, or that `wait_for_capacity` is badly conceived, or both?

PR #579, opened 2021-12-02
Add test that would make wait_for_capacity hang if it doesn't loop

nox:


tests/h2-tests/tests/flow_control.rs
```diff
@@ -1561,3 +1561,55 @@ async fn data_padding() {
...
+
+        stream.reserve_capacity(5);
+
+        // Initial window size was 5 so current capacity is 0 even if we just reserved.
+        assert_eq!(stream.capacity(), 0);
+
+        // FIXME(nox): This currently hangs because the capacity was reduced to 0 by the `send_data`
+        // call, but the stream's `send_capacity_inc` flag was never cleared and is still true,
+        // so `poll_for_capacity` immediately returns `Ready(Some(0))`.
+        let mut stream = h2.drive(util::wait_for_capacity(stream, 5)).await;
```
nox, line 1606: @seanmonstar I managed to write a test that hangs on master, for the same reason your test hangs in my max send  buffer size branch. Do you think h2's innards are wrong, or is it `wait_for_capacity` that is wrong?

nox: https://github.com/hyperium/hyper/blob/6169db250c932dd012d391389826cd34833077b4/src/proto/h2/mod.rs#L129-L145

Hyper itself loops on `Ready(Some(0))`, so I guess `wait_for_capacity` is broken?

seanmonstar: Yes, I think `wait_for_capacity` is wrong to return `Pending` without having registered a task. It should probably be a loop, since calling `poll_capacity` again will have noticed `send_capacity_inc` is now false and registered for the next capacity wakeup.

nox: I changed `wait_for_capacity` in that commit then.

seanmonstar (approved):

PR #580, opened 2021-12-06
Add max send buffer per stream option

seanmonstar:
An alternative to #577.

This makes the max send buffer size **per stream**, instead of for the entire connection. It also does a lot less to try to enforce it, since the assigning of capacity is a complicated beast. All this does is clamp how high something like `SendStream::capacity()` will return. It's more just an advisory to the user of the `SendStream`.

Since hyper currently will only produce more data once the available capacity is over zero, it should behave well with this change. I've tried out this patch in the hyper HTTP/2 benchmarks, and see that they all still pass (don't hang). However, this is missing some unit tests in `flow_control.rs`...

### DIFF
--- a/src/proto/streams/send.rs
+++ b/src/proto/streams/send.rs
@@ -340,10 +340,18 @@ impl Send {
         let available = stream.send_flow.available().as_size();
         let buffered = stream.buffered_send_data;
 
+        tracing::info!(
+            ?available,
+            ?buffered,
+            max_buffer_size = self.max_buffer_size
+        );
+
         if available as usize <= buffered {
             0
         } else {
-            available.min(self.max_buffer_size as WindowSize) - buffered as WindowSize
+            available
+                .min(self.max_buffer_size as WindowSize)
+                .saturating_sub(buffered as WindowSize)
         }
     }
 
--- a/src/proto/streams/send.rs
+++ b/src/proto/streams/send.rs
@@ -28,6 +28,9 @@ pub(super) struct Send {
     /// > the identified last stream.
     max_stream_id: StreamId,
 
+    /// The maximum amount of bytes a stream should buffer.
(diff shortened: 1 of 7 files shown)


src/proto/streams/send.rs
```diff
@@ -339,7 +343,7 @@ impl Send {
         if available as usize <= buffered {
             0
         } else {
-            available - buffered as WindowSize
+            available.min(self.max_buffer_size as WindowSize) - buffered as WindowSize
```
nox, line 346: I think this can underflow, right?

nox: Never mind, clearly I had not woken up correctly yet.

seanmonstar: No, you were right. I add a test, and it did panic with an overflow. I've fixed that.

src/proto/streams/send.rs
```diff
@@ -340,10 +340,18 @@ impl Send {
...
+            ?buffered,
+            max_buffer_size = self.max_buffer_size
+        );
+
         if available as usize <= buffered {
             0
         } else {
-            available.min(self.max_buffer_size as WindowSize) - buffered as WindowSize
+            available
+                .min(self.max_buffer_size as WindowSize)
```
nox, line 353: AFAIK this is still not good: `self.max_buffer_size` could be bigger than `u32::MAX`.

src/proto/streams/send.rs
```diff
@@ -340,10 +340,18 @@ impl Send {
         let available = stream.send_flow.available().as_size();
         let buffered = stream.buffered_send_data;
 
+        tracing::info!(
```
nox, line 343: `info` is bigger than `debug` right? I would use `debug` or lower here.

PR #582, opened 2021-12-08
Prepare v0.3.8

seanmonstar:


robjtede:
> Add `max_send_buffer_size` options to client and server builders, and a default of ~400MB

We've got "large payload" tests hanging in Actix Web; seems like the default is 400KiB not MiB. 

https://github.com/hyperium/h2/blob/1945c027f9a51fb79372e84a35109aeafa4fedca/src/proto/mod.rs#L36

seanmonstar:
Er, 400kb, yes.

But still, we were expecting that nothing should hang because of this...

seanmonstar:
Is this the main place where h2 capacity is checked/used in actix? https://github.com/actix/actix-web/blob/master/actix-http/src/h2/dispatcher.rs#L226

robjtede:
spraying some logs: it's hanging on `poll_capacity` when it crosses the boundary of 400KiB of reserved capacity requested

PR #583, opened 2021-12-09
Increase max send buf default to u32::MAX

seanmonstar:
A hang was reported after the new release (https://github.com/hyperium/h2/pull/582#issuecomment-989340854).

@robjtede for now, this should make the default effectively off, while we investigate the hang.


seanmonstar:
We found a proper fix in #585

PR #586, opened 2021-12-09
Update tracing-subscriber and use tracing-tree when testing

nox:
This makes reading the logs way easier on the eyes.

### DIFF
--- a/tests/h2-support/Cargo.toml
+++ b/tests/h2-support/Cargo.toml
@@ -7,9 +7,11 @@ edition = "2018"
 [dependencies]
 h2 = { path = "../..", features = ["stream", "unstable"] }
 
+atty = "0.2"
 bytes = "1"
 tracing = "0.1"
-tracing-subscriber = { version = "0.2", default-features = false, features = ["fmt", "chrono", "ansi"] }
+tracing-subscriber = { version = "0.3", default-features = false, features = ["fmt"] }
+tracing-tree = "0.2"
 futures = { version = "0.3", default-features = false }
 http = "0.2"
 tokio = { version = "1", features = ["time"] }
--- a/tests/h2-support/src/trace.rs
+++ b/tests/h2-support/src/trace.rs
@@ -1,41 +1,17 @@
-use std::{io, str};
 pub use tracing;
 pub use tracing_subscriber;
 
+use tracing_subscriber::layer::SubscriberExt;
+use tracing_subscriber::util::SubscriberInitExt;
+
 pub fn init() -> tracing::dispatcher::DefaultGuard {
-    tracing::subscriber::set_default(
-        tracing_subscriber::fmt()
-            .with_max_level(tracing::Level::TRACE)
-            .with_span_events(tracing_subscriber::fmt::format::FmtSpan::CLOSE)
-            .with_writer(PrintlnWriter { _p: () })
-            .finish(),
-    )
-}
+    let use_colors = atty::is(atty::Stream::Stdout);
+    let layer = tracing_tree::HierarchicalLayer::default()
+        .with_writer(tracing_subscriber::fmt::writer::TestWriter::default())
+        .with_indent_lines(true)
+        .with_ansi(use_colors)
+        .with_targets(true)
+        .with_indent_amount(2);
 
-struct PrintlnWriter {
-    _p: (),
-}
-
-impl tracing_subscriber::fmt::MakeWriter for PrintlnWriter {
-    type Writer = PrintlnWriter;
-    fn make_writer(&self) -> Self::Writer {
-        PrintlnWriter { _p: () }
-    }
-}
-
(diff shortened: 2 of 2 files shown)


nox:
This removes timestamps, but durations are still shown, so I don't know if this is good enough or not.

tests/h2-support/src/trace.rs
```diff
@@ -1,41 +1,19 @@
...
-    }
-
-    fn write_fmt(&mut self, fmt: std::fmt::Arguments<'_>) -> io::Result<()> {
-        println!("{}", fmt);
-        Ok(())
-    }
+pub fn init() -> tracing::dispatcher::DefaultGuard {
+    let use_colors = atty::is(atty::Stream::Stderr);
+    let layer = tracing_tree::HierarchicalLayer::default()
+        .with_writer(std::io::stderr)
```
hawkw, line 10: the previous code used `println!` rather than writing directly to an IO stream so that the log output was captured by libtest's output capturing. I think we might want to change this code to do the same. That way, we can use `cargo test -- --show-output`, and get the output categorized by test, rather than having it all dumped to the console in a big mess when tests run in parallel. Also, it will only print logs for failing tests by default, which should make a `cargo test` run less noisy out of the box.

`tracing-subscriber` has a [`TestWriter` type](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.TestWriter.html) that does this (I believe it was added after the `PrintlnWriter` here was written). 

Since that uses `println!` rather than `eprintln!`, I think we'll want to check if stdout is a tty, rather than stderr:
```suggestion
    let use_colors = atty::is(atty::Stream::Stdout);
    let layer = tracing_tree::HierarchicalLayer::default()
        .with_writer(tracing_subscriber::fmt::writer::TestWriter::default())
```

tests/h2-support/src/trace.rs
```diff
@@ -1,41 +1,19 @@
...
+        .with_indent_lines(true)
+        .with_ansi(use_colors)
+        .with_targets(true)
+        .with_indent_amount(2);
```
hawkw, line 14: take it or leave it: perhaps we want to add
```suggestion
        .with_indent_amount(2)
        .with_thread_ids(true)
        .with_thread_names(true);
```
in particular, adding thread names will give us a nice way to know which test the events were logged from, since the thread's name defaults to the name of the test.

tests/h2-support/Cargo.toml
```diff
@@ -7,9 +7,11 @@ edition = "2018"
 [dependencies]
 h2 = { path = "../..", features = ["stream", "unstable"] }
 
+atty = "0.2"
 bytes = "1"
 tracing = "0.1"
-tracing-subscriber = { version = "0.2", default-features = false, features = ["fmt", "chrono", "ansi"] }
+tracing-subscriber = { version = "0.3", default-features = false, features = ["fmt", "ansi"] }
```
hawkw, line 13: do we actually need to enable the `ansi` feature here?

tests/h2-support/src/trace.rs
nox: I added it but I'm not sure this is useful: I only ever look at logs running a single test, not many of them.

tests/h2-support/Cargo.toml
nox: Nope.

tests/h2-support/src/trace.rs
nox: Capture d’écran 2022-01-20 à 10 45 20

It's a bit verbose as you can see.

hawkw: hmm, we could remove it --- i suppose that since libtest is capturing the logs now, they'll also be printed out per-test. the only reason to consider having thread names here is if we have tests that spawn multiple threads and they give names to those threads...and we only have one test that spawns threads (https://github.com/hyperium/h2/search?q=thread%3A%3Aspawn).

i'm fine with removing the thread-name configuration, then!

tests/h2-support/src/trace.rs
```diff
@@ -1,41 +1,18 @@
...
-    }
+pub fn init() -> tracing::dispatcher::DefaultGuard {
+    let use_colors = atty::is(atty::Stream::Stdout);
+    let layer = tracing_tree::HierarchicalLayer::default()
+        .with_writer(tracing_subscriber::fmt::writer::TestWriter::default())
+        .with_indent_lines(true)
+        .with_ansi(use_colors)
+        .with_targets(true)
+        .with_indent_amount(2)
+        .with_thread_names(true);
```
hawkw, line 15: since --- as we discussed --- the thread names are actually not that useful, i'm fine with removing them to make the logs a little less verbose. up to you!
```suggestion
        .with_indent_amount(2);
```

hawkw (approved):
okay, this looks good to me!

PR #600, opened 2022-02-02
Specify a closed stream with the `CANCEL` code as valid

DDtKey:
According to HTTP/2, the `CANCEL` code only indicates that the stream is no longer needed.
`CANCEL (0x8):  Used by the endpoint to indicate that the stream is no longer needed.`

But when the client receives the response body data and the `CANCEL` code from the server, it returns an error.
Error example with `hyper`: 
```rs
hyper::Error(Body, Error { kind: Reset(StreamId(3), CANCEL, Remote) })
```

This happens due to incorrect checking of an open stream:: [link](https://github.com/hyperium/h2/blob/508bcb1d27e1e90f58c3254cac5d5b4fa22c17bc/src/proto/streams/state.rs#L431)


seanmonstar:
Hey, thanks for the PR! Hm, I'd love to understand this better. I _think_ the current behavior is correct, that if the server sends a `CANCEL` reset, we should return an "error" that includes that the stream is canceled. It's not saying that it's a huge internal error, but it is different from the normal `Ok` case, no?

DDtKey:
Not really 🤔
This is a legal situation where the data has been fully received and server has also returned the CANCEL code.  
But now, the client receives the entire data set and returns an error during the polling the stream, although in fact it was polled to the end. 

Although the name of code is contradictory, this code only says that the stream is no longer needed:
`CANCEL (0x8):  Used by the endpoint to indicate that the stream is no longer needed.`

glebpom:
Hi. I want to add a few details since we worked with @DDtKey on this issue. We met a flaky integration test, where we sent a large request from `reqwest` to a `poem` web server (based on hyper). In some circumstances, we got the `canceled` error instead of the expected `413` error. The problem existed on CI, but not locally. However, when we changed the reactor to `current-thread` or limited the number of threads to `1`, it started to reliable reproduce locally as well. Wireshark has shown that the response was entirely generated, and the stream was `CANCEL`ed after that. We used the `HTTPS` endpoint, so we decrypt the traffic with pre-master keys.
image
The data was fully sent, and the `content-length` was also correct. However, the client returned the error [here](https://github.com/hyperium/hyper/blob/913be88f7155ea4695c7f8098d968bf1f340f8be/src/body/to_bytes.rs#L61). The first chunk contained only a part of the data. 
Applying the patch above reliably fixes the problem.

DDtKey:
To be clear, `reqwest` successfully received headers and formed the `rewest::Response`.  
But when we tried to get the response body (`bytes()` -> `hyper::to_bytes()` -> .. ), we got this behavior. The data was received, but the client returned an error even though the stream was actually polled to the end.

seanmonstar:
I see, I think I understand the issue now. The problem isn't so much about `CANCEL` specifically, it's that if a stream is reset, even though some data was received already, the reset code trumps getting anymore of the data out. I think what we want to do is allow a user to poll more data out of the buffer, and only when there is nothing left, _then_ check if there is an error code. That would handle any reset code. What do you think?

glebpom:
Looks good to me. However, I'm still thinking that, in accordance with spec, `CANCEL` should be treated as a closed channel, rather than an error. So, in my opinion, both approaches may be implemented.

olix0r:
@glebpom While I understand that applications may want to handle these cases similarly--and I agree it's important to fix the current behavior--I think it would be bad if we _lost_ the ability to differentiate a `CANCEL` from an `END_STREAM`.

For example, imagine if a proxy is transporting a HTTP/2 stream. We'd want the proxy to propagate the `CANCEL` reset and not simply treat it as an `END_STREAM`, right? If the library treats these cases identically, it would be impossible for such a proxy to preserve the original server behavior.

glebpom:
@olix0r Ok, that makes sense. Seems like on the `h2` layer the `CANCEL` error should propagate. 👍

DDtKey:
It looks good to me too.
But how do you propose to implement it? 🤔

DDtKey:
@seanmonstar any comments or news on top of this? 🙂

glebpom:
A long time has passed since we discussed the bug. We observed the same problem a few more times with different client implementations on the unpatched h2 versions. 
So, what would be the plan to fix it on a hyper level properly? Is there an issue in the hyper repo we can link to? Or can we go with the patch in h2?
Thanks.

glebpom:
Hi, @seanmonstar @olix0r. Any updates on this? Thanks

DDtKey:
After testing the case against latest `h2` - this patch doesn't seem to work after https://github.com/hyperium/h2/pull/634

Now the `h2` uses `Reset(NO_ERROR)` - what's actually seems correct, but probably it should be fixed on `hyper` side, because this causes error on client side:
`hyper::Error(Body, Error { kind: Reset(StreamId(3), NO_ERROR, Remote) }) `

cc @seanmonstar 

I'm closing this PR because the way to fix should be different now.

PR #601, opened 2022-02-04
Support SendResponse::send_continue

rawler:
Support for `Expect: 100-continue` is mandated as MUST by
https://datatracker.ietf.org/doc/html/rfc7231#section-5.1.1. Yet servers
built on `h2` cannot currently support this requirement.

One example of such usage, is [hyper #2743](https://github.com/hyperium/hyper/issues/2743).

This approach adds a `send_continue` method to `SendResponse` that a server
application can use to implement support itself. This PR does _not_ solve
the feature itself, it merely provides sufficient support for a server
application to implement the functionality as desired.


rawler:
One possible alternate solution, could be to implement more of the functionality directly, by hooking up the first attempt to read the request-body, as a signal to send `100-continue`, if the header was there on the Request. Ideally, that would be implemented in `Request::into_body`, or add the behavior to the first call of `RecvStream::poll_data`. However, it seemed more invasive than I'd like it to be, and I didn't manage to wire it up in an acceptable way anyways.

src/proto/streams/state.rs
```diff
@@ -111,7 +111,7 @@ impl State {
                     Open { local, remote }
                 }
             }
-            HalfClosedRemote(AwaitingHeaders) | ReservedLocal => {
+            HalfClosedRemote(AwaitingHeaders | Streaming) | ReservedLocal => {
```
rawler, line 114: I spent some time thinking about this in particular. There is no documentation on the struct hinting at how to reason about the state, but if I understood it correctly `HalfClosedRemote` means that we've gotten an EOS from the remote side. `HalfClosedRemote(AwaitingHeaders)` this side currently haven't written headers yet (the enum-name `Peer` is a bit misleading here?). This used to encode the rule that headers should only be sent _once_. With 100-continue in the picture, however, that doesn't really hold true. The easiest way I could see to model this, is to simply relax this rule as proposed here. Perhaps another approach could be feasible? I.E. something not "Opening the send-half", but instead explicitly "finish sending headers"?

seanmonstar: The code comments at the top of file include the flow chart from the spec, but the the spec includes more description of each possible state: https://httpwg.org/specs/rfc7540.html#StreamStates.

It looks like your understanding is correct. I've also found the usage of `Peer` in this crate to be the opposite of what I usually think, but it is consistently used in this file and elsewhere to mean "me" (I usually think of peer as the remote, but I didn't come up with those names :shrug:).

src/server.rs
```diff
@@ -1049,6 +1049,21 @@ impl<B: Buf> SendResponse<B> {
...
+    /// Send a 100-continue response to a client request.
+    ///
+    /// The [`SendResponse`] instance is already associated with a received
+    /// request.  This function may only be called once per instance and only if
+    /// [`send_reset`] or [`send_response`] has not been previously called.
+    ///
+    /// [`SendResponse`]: #
+    /// [`send_reset`]: #method.send_reset
+    /// [`send_response`]: #method.send_response
+    pub fn send_continue(&mut self) -> Result<(), crate::Error> {
```
seanmonstar, line 1061: It might be out-of-scope to suggest this, but should we perhaps consider a slightly broader method, that allows sending other informational responses, like `103 Early Hints` or future codes? In that case, we'd want to be able to send headers too...

src/proto/streams/state.rs
seanmonstar: However, I think we don't want to change the state here, since sending a `HEADERS` frame with a 1xx code shouldn't mean it is now waiting to send `DATA` frames. It still needs to send the "final" code in another `HEADERS` frame.

src/server.rs
rawler: I thought about it when writing the PR, but avoided it since I'm not sure what it's signature should be.  `send_info(Response<()>)` feels a bit too wide, given that practically only 1xx responses are allowed here. OTOH, `send_response(Response<()>)` is _also_ wider than it technically supports. (Extensions are cleared, the body unused, and the only `Version` supported really, is HTTP2), so maybe such API is fine.

rawler:
Thank you :pray: for a good and clear review. I've changed the according to the feedback.

rawler:
Ping

rawler:
Ping

rawler:
Ping @seanmonstar

rawler:
Maybe @LucioFranco have some time for a review? :pray:

zh-jq:
Any progress? It's an important feature

zh-jq-b:
@seanmonstar This is ready for a new review, do you have anytime to do it?

mspublic:
Any update on this @seanmonstar @LucioFranco ?

GlenDC:
As discussed on discord, I'm willing to help out in any capacity that you can use me to get this PR shipped. Be it more code contribution, review, testing, docs, ... do let me know.

src/server.rs
```diff
@@ -1049,6 +1049,22 @@ impl<B: Buf> SendResponse<B> {
...
+    /// Send a non-final 1xx response to a client request.
+    ///
+    /// The [`SendResponse`] instance is already associated with a received
+    /// request.  This function may only be called if [`send_reset`] or
+    /// [`send_response`] has not been previously called.
+    ///
+    /// [`SendResponse`]: #
+    /// [`send_reset`]: #method.send_reset
+    /// [`send_response`]: #method.send_response
+    pub fn send_info(&mut self, response: Response<()>) -> Result<(), crate::Error> {
```
seanmonstar, line 1061: Perhaps this just needs to document with a `# Panics` section that passing a response with a non-1xx status code is a misuse of the method.

rawler: Agreed, and fixed.

mspublic:
Any status on this? @seanmonstar

arthurdandrea:
As far as I've understood this has been superseded by #865, that was merged and released in [v0.4.13](https://github.com/hyperium/h2/releases/tag/v0.4.13)

PR #602, opened 2022-02-06
HTTP/2 is the official name, so update terms

LPardue:
The IETF only define HTTP/2 with no minor version. And at this stage I think its safe to assume that will never change.

Remaining references to HTTP/2.0 after this PR would be: one in the changelog, one in the GitHub repo About section. 
I avoided touching the changelog since it felt like rewriting history, but let me know if you feel different


hawkw (approved):
this seems good to me!

PR #604, opened 2022-02-11
Upgrade dev-dependencies

djc:


Cargo.toml
```diff
@@ -55,22 +55,21 @@ indexmap = { version = "1.5.2", features = ["std"] }
...
 
 # Examples
 tokio = { version = "1", features = ["rt-multi-thread", "macros", "sync", "net"] }
-env_logger = { version = "0.5.3", default-features = false }
-rustls = "0.19"
-tokio-rustls = "0.22"
-webpki = "0.21"
-webpki-roots = "0.21"
+env_logger = { version = "0.9", default-features = false }
+rustls = "0.20.2"
```
olix0r, line 70: It would probably best to avoid the explicit rustls dependency in favor of `tokio_rustls::rustls`, right? That way there's no chance for them to get out of sync

olix0r (approved):

Cargo.toml
djc: Good point, done.

PR #605, opened 2022-02-24
CI: use minimal-versions on MSRV check

seanmonstar:


seanmonstar:
Oh right, woops, this can never work since MSRV is never nightly.

Issue #525, opened 2021-03-18
malformed authority error with HTTP2 and gRPC


arthurlm:
I am running h2 as a service in k8s to process gRPC requests.

Some clients of my service (envoy HTTP proxy) got error responses du to invalid authority field:
```
[2021-03-18T10:54:07Z DEBUG h2::server] malformed headers: malformed authority (b"extension/k8s-namespace-arthurlm-test/auth-server"): invalid uri character 
```

I do not have any ability to change envoy behavior: it need to use HTTP2 + TLS + cannot change request headers and payload.
I agree this authority is a little bit unexpected...

Should we make this kind of Authority allowed ?
Is there any way to allow h2 processing the request even if `Authority::parse` failed ?

nox:
Did you file an issue against the envoy HTTP proxy so they don't emit malformed authorities

nox:
#442

arthurlm:
Unfortunately no 😞.

I have fork h2 and just  show a log warn if authority field parsing fail.
See diff: https://github.com/hyperium/h2/compare/master...arthurlm:master

I would be happy to merge this change in h2 but I do not know if this policy is the expected one by h2.

nox:
I am not sure what you mean by "unfortunately no". Isn't Envoy [a FOSS project](https://github.com/envoyproxy/envoy)? Personally I don't think a patch to allow completely malformed authorities should be accepted.

nox:
Let's move this discussion to #442 please, as it is the same issue.

Issue #442, opened 2020-01-18
Support for h2 over UDS


gila:
When using tonic/h2 over UDS, I think it should be possible to allow the '/' in the authority header. This looks related to #345 where Jan is also mentioning potential issues might arise once h2 would be used over UDS.  

From RFC-3986:

```
For example, the "file" URI  scheme is defined so that no authority, 
an empty host, and "localhost"  all mean the end-user's machine, 
whereas the "http"  scheme considers a missing authority or
empty host invalid.
```

A longer description of the issue can be found at https://github.com/hyperium/tonic/issues/243

Perhaps it would be possible to support file:// (and then also perhaps the unix:// scheme) and depending on that, determine if the authority header is valid for the current scheme?

saschagrunert:
Hey, any update on this one? I think the addition would make sense, wouldn't it?

nox:
AFAICT grpc-go stopped sending a malformed authority so I'm closing this.

teozkr:
For what it's worth, I'm still getting this with grpc-go v1.38.0, which was released on 2021-05-19 (a few weeks after your comment):

> 2021-11-23T22:02:25.675177Z DEBUG Connection{peer=Server}: h2::server: malformed headers: malformed authority (b"/foo/bar.sock"): invalid uri character

I'll try to reproduce with grpc-go v1.42.0 (the current latest release).

teozkr:
Same issue with grpc-go v1.42.0.

PR #612, opened 2022-04-01
Deny requerst if :authority field is invalid only with CONNECT method

arthurlm:
Looking at RFCs:

1. [RFC 7540: HTTP V2@section 8.1.2.3. Request Pseudo-Header Fields](https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.3).

> The ":authority" pseudo-header field includes the authority
      portion of the target URI ([[RFC3986], Section 3.2](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2)).  The authority
      MUST NOT include the deprecated "userinfo" subcomponent for "http"
      or "https" schemed URIs.

> All HTTP/2 requests MUST include exactly one valid value for the
   ":method", ":scheme", and ":path" pseudo-header fields, unless it is
   a CONNECT request ([Section 8.3](https://datatracker.ietf.org/doc/html/rfc7540#section-8.3)).  An HTTP request that omits
   mandatory pseudo-header fields is malformed ([Section 8.1.2.6](https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.6)).

2. [RFC 3986: Uniform Resource Identifier@section 3.2. Authority](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2)

> Many URI schemes include a hierarchical element for a naming
   authority so that governance of the name space defined by the
   remainder of the URI is delegated to that authority (which may, in
   turn, delegate it further).

> The authority component is preceded by a double slash ("//") and is
   terminated by the next slash ("/"), question mark ("?"), or number
   sign ("#") character, or by the end of the URI.

> Non-validating parsers (those that merely separate a URI reference into
   its major components) will often ignore the subcomponent structure of
   authority, treating it as an opaque string from the double-slash to
   the first terminating delimiter, until such time as the URI is
   dereferenced.

3. [RFC 7540: HTTP V2@section-8.3. The CONNECT Method](https://datatracker.ietf.org/doc/html/rfc7540#section-8.3)

> In HTTP/2, the CONNECT method is used to establish a tunnel over a
   single HTTP/2 stream to a remote host for similar purposes.  The HTTP
   header field mapping works as defined in [Section 8.1.2.3](https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.3) ("Request
   Pseudo-Header Fields"), with a few differences.  Specifically:

> o  The ":authority" pseudo-header field contains the host and port to
      connect to (equivalent to the authority-form of the request-target
      of CONNECT requests (see [[RFC7230], Section 5.3](https://datatracker.ietf.org/doc/html/rfc7230#section-5.3))).

>  A CONNECT request that does not conform to these restrictions is
   malformed ([Section 8.1.2.6](https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.6)).

So, from my understanding:

* Following (1):
  * `:authority` field is not mandatory in HTTP2.
  * It should be a valid URI without scheme and userinfo.

* Following (2):
  * URI can include sub path component (even if `/` character is not expected for regular authority).
  * There are HTTP2 clients in the wild that might send "invalid" `:authority`. It should be treated as opaque string.

* Following (3):
  * `:authority` field should be a valid host and port in the case of `CONNECT` method.

**NOTE 1**: I have not read the whole RFCs but just read in details the mentioned sections. If anyone have better understanding of this RFCs, please feel free to comment / edit this PR :wink: !

**NOTE 2**: Please have a look at my first [comment](https://github.com/hyperium/h2/pull/487#issuecomment-1080658944) about k8s usage with `tonic` for more details.

---

This change could fix a lot of already referenced issues / PRs using `h2` and gRPC / `tonic`.

* https://github.com/hyperium/h2/issues/525
* https://github.com/hyperium/h2/issues/442
* https://github.com/hyperium/h2/pull/487
* https://github.com/projectcontour/contour/issues/4278

It will also:

* Improve support for gRPC call using go clients
* Allow people to develop nice k8s extension with `tonic` (which is not possible for now without forking `h2`)
* Help `h2` to better fit to HTTP2 RFC specification :grin:


bachp (approved):
This makes it possible to use h2 with k8s.

bachp:
@seanmonstar do you see any issue with this change?

Forsworns:
Any update on this one? Just met the same problem when communicate with k8s via tonic :(

tiagolobocastro:
Any update on this? Would be good to stop using a fork :)

mythi:
> Any update on this one? Just met the same problem when communicate with k8s via tonic :(

k8s 1.26 kubelet sets `localhost` authority for device-plugin and CSI.

arthurlm:
Even if recent version of k8s set `localhost` for `:authority` field:
- not everyone has access to latest version of k8s (and in big company it can takes years)
- in cloud as service provider (EKS for example) latest version of k8s is still 1.24
- `h2` still does not respect HTTP2 RFC without this change

I do not really understand why maintainers do not want to review / merge / close this PR.
I guess it is probably dues to some lack of time or PR visibility / interest.

I still hope this will be merged or discussed one day 😄.
So this is why I have not closed this for now...

mythi:
@arthurlm FWIW I agree with you. Just wanted to share what I know k8s kubelets can do today.

PR #614, opened 2022-05-10
Remove `B: Buf` bound on `SendStream`'s parameter

djkoloski:
This bound is only necessary when calling methods on `SendStream` because we don't use any associated types of `B`'s `Buf` impl in the type. Relaxing this bound allows us to construct `SendStream`s with non-`Buf` `B` (though we can't do anything with them).


djkoloski:
https://github.com/hyperium/h2/pull/616 should fix the nightly CI.

djkoloski:
Bumping this after #616 was merged.

LucioFranco (approved):
This seems fine to me, I prefer to not have the bounds on the struct anyways. Also, I don't think this is breaking so I think we are fine to merge?

djkoloski:
Quick bump on this, if there are no objections I think this would be a good improvement. 🙂

LucioFranco:
Oh sorry, we can get this merged :D

PR #615, opened 2022-05-12
Add a message for EOF-related broken pipe errors

nox:
It's quite confusing from production logs when all I get is "broken pipe" and I don't know which path was taken for that error to be logged.

### DIFF
--- a/src/proto/streams/state.rs
+++ b/src/proto/streams/state.rs
@@ -303,7 +303,13 @@ impl State {
             Closed(..) => {}
             ref state => {
                 tracing::trace!("recv_eof; state={:?}", state);
-                self.inner = Closed(Cause::Error(io::ErrorKind::BrokenPipe.into()));
+                self.inner = Closed(Cause::Error(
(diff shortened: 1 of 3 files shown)


src/proto/streams/state.rs
```diff
@@ -303,7 +303,13 @@ impl State {
             Closed(..) => {}
             ref state => {
                 tracing::trace!("recv_eof; state={:?}", state);
-                self.inner = Closed(Cause::Error(io::ErrorKind::BrokenPipe.into()));
+                self.inner = Closed(Cause::Error(
+                    io::Error::new(
+                        io::ErrorKind::BrokenPipe,
+                        "stream closed because of broken pipe",
```
LucioFranco, line 309: Shouldn't it be `because of a broken pipe`?

LucioFranco (approved):
LGTM

nox:
@seanmonstar 🥺

What would be the best error message here?

seanmonstar (approved):

PR #616, opened 2022-05-17
Move unexported macro doctest into unit test

djkoloski:
Nightly has begun running doctests for unexported macros as of
https://github.com/rust-lang/rust/pull/96630, which caused a doctest for
test_unpack_octets_4 which was previously ignored to be run. This broke
the CI because macros that are not exported with `#[macro_export]`
cannot be used from external crates (and thus cannot be doctested). This
change ignores the doctest and copies the relevant code into a unit
test.

### DIFF
--- a/src/frame/mod.rs
+++ b/src/frame/mod.rs
@@ -11,7 +11,8 @@ use std::fmt;
 ///
 /// # Examples
 ///
-/// ```rust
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):
Thanks :)

PR #618, opened 2022-05-18
Add `Status::is_reset` function

### LINKED ISSUE #617: No way to determine if the server reset the stream
When a server (GCP Pub/Sub in my case) sends a reset frame, h2 reports it as a status with `Kind::Reset`. However, there is no direct way of testing if the status was a reset, analogous to `is_go_away` or `is_io`.

If an `is_reset` function would be an acceptable solution, I can provide a PR for this.

bruceg:
We are trying to determine if the remote issued a stream reset and found that this test is missing from the `Status` type.

Closes #617

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -59,10 +59,7 @@ impl Error {
 
     /// Returns true if the error is an io::Error
     pub fn is_io(&self) -> bool {
-        match self.kind {
-            Kind::Io(_) => true,
-            _ => false,
-        }
+        matches!(self.kind, Kind::Io(..))
     }
 
     /// Returns the error if the error is an io::Error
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -92,6 +89,11 @@ impl Error {
         matches!(self.kind, Kind::GoAway(..))
     }
 
+    /// Returns true if the error is from a `RST_STREAM`.
+    pub fn is_reset(&self) -> bool {
+        matches!(self.kind, Kind::GoAway(..))
```
seanmonstar, line 94:
```suggestion
        matches!(self.kind, Kind::Reset(..))
```

Issue #621, opened 2022-06-08
Stream stacking occurs when H2 processes HTTP2 RST_STREAM frames.


silence-coding:
@seanmonstar As described in the issue in hyper, I think the size of the slab should be limited to max_concurrent_streams + max_concurrent_reset_streams . When the size of the slab exceeds the upper limit, the recv_frame method should discard the received Headers and PushPromise frames and send the protocol error to the client.
https://github.com/hyperium/hyper/issues/2877

silence-coding:
Requests can be provided to users after REST. However, num_reset_streams must be reduced after the request is processed to ensure that max_reset_streams takes effect.
https://github.com/hyperium/h2/issues/30

lidong14:
@seanmonstar  please check this MR

kevinji:
@silence-coding this should be closed too now that #668 is merged.

PR #622, opened 2022-06-09
Fix: Limit the number in the slab

silence-coding:
FIX: https://github.com/hyperium/h2/issues/621


silence-coding:
@seanmonstar @carllerche

silence-coding:
@hawkw Thank you.

silence-coding:
@hawkw Hello, when are you free to check

hawkw (commented):
hmm, is it really correct to GOAWAY the whole connection in this case? sending a GOAWAY will tell the client it can *never* initiate new streams on this connection, but those slab indices will become available again (in [`clear_expired_reset_streams`](https://github.com/hyperium/h2/blob/f48f6b22df75bc73a238163baa11d9fb5f68a535/src/proto/streams/recv.rs#L857-L871])). so, this condition is transient. is there a reason we can't reset the new stream rather than GOAWAYing the whole connection? it is entirely possible i'm overlooking something here, i haven't looked at this code in a while...

seanmonstar:
It seems in `fn open()`, we check that the number of currently active received streams is under the max, and if at the limit, then the stream is rejected (likely reset). With this new change, the stream would be accepted, but if we had too many pending reset streams, the connection would go away. That seems... aggressive...

Then again, if a user has decided to have no more than some limit, it's probably to reduce resource usage... Resetting more streams would just add to the "too many"... The problem, I think, is if a remote has been keeping the local peer at the max, and then a stream is reset, it would pretty much immediately close it all down.

seanmonstar:
Shouldn't it be sufficient to just set the `max_reset_streams` to something? Anything over that limit should be forgotten immediately, no?

silence-coding:
> hmm, is it really correct to GOAWAY the whole connection in this case? sending a GOAWAY will tell the client it can _never_ initiate new streams on this connection, but those slab indices will become available again (in [`clear_expired_reset_streams`](https://github.com/hyperium/h2/blob/f48f6b22df75bc73a238163baa11d9fb5f68a535/src/proto/streams/recv.rs#L857-L871%5D)). so, this condition is transient. is there a reason we can't reset the new stream rather than GOAWAYing the whole connection? it is entirely possible i'm overlooking something here, i haven't looked at this code in a while...

I'm just starting to get to know H2 and I'm not familiar with it. I just think it's safer to disconnect in this scenario. Developers can optimize this problem by tweaking max_concurrent_streams and max_concurrent_reset_streams.

silence-coding:
> It seems in `fn open()`, we check that the number of currently active received streams is under the max, and if at the limit, then the stream is rejected (likely reset). With this new change, the stream would be accepted, but if we had too many pending reset streams, the connection would go away. That seems... aggressive...
> 
> Then again, if a user has decided to have no more than some limit, it's probably to reduce resource usage... Resetting more streams would just add to the "too many"... The problem, I think, is if a remote has been keeping the local peer at the max, and then a stream is reset, it would pretty much immediately close it all down.

@seanmonstar Sorry, I didn't understand you well. Is it a direct GOAWAY or REST_STREAM? Or something better.

silence-coding:
> Who can answer the problem of the PR or submit a new PR with a better solution to fix the problem? #621

@seanmonstar @hawkw Can someone answer that? If this problem is not solved, services built on the crate using h2 are at high ddos risk.

silence-coding:
How will the h2 community deal with this problem in the future? If you guys are busy, is there anyone else in the community familiar with this issue? https://github.com/hyperium/h2/issues/621

silence-coding:
Does anyone pay attention to this issue?

silence-coding:
@LucioFranco Hello, can this take a little of your precious time? Thank you.

src/proto/streams/streams.rs
```diff
@@ -312,7 +312,13 @@ where
 impl<B> DynStreams<'_, B> {
     pub fn recv_headers(&mut self, frame: frame::Headers) -> Result<(), Error> {
         let mut me = self.inner.lock().unwrap();
-
+        if me.store.num_wired_streams() > me.counts.max_streams() {
+            tracing::error!("HEADERS: number of streams exceeds the upper limit ({})", me.counts.max_streams());
+            return Err(Error::remote_go_away(
```
LucioFranco, line 317: Its not clear to me that this is the correct response as stated above. Would be good to get some reference from the spec or any other implementations for what they do here.

LucioFranco (changes requested):
Hi sorry for the delay, I think we've had a lot of open issues and it takes some time to get to all of them.

LucioFranco:
Would also be good to get CI to run against this change which may just need a new commit to trigger it.

silence-coding:
Sorry, I clicked re-request by mistake. @LucioFranco

src/proto/streams/streams.rs
silence-coding: I didn't find anything more useful
https://www.rfc-editor.org/rfc/rfc7540#section-6.5.2

PR #625, opened 2022-07-18
Remove PollExt trait

kckeiks:
Hello,
[Rust #63512](https://github.com/rust-lang/rust/pull/63512) was stabilized so this PR removes the `PollExt` trait.


seanmonstar:
Thanks for the PR! It looks like the MSRV job is failing (I've fixed that on master), care to rebase? When did the methods stabilize? We try to keep the MSRV at least 6 months old, and prefer to only upgrade when there's a decent improvement.

kckeiks:
@seanmonstar They were stabilized on January of 2021 (https://github.com/rust-lang/rust/pull/80968).

seanmonstar (approved):
Wonderful, thank you :)

PR #627, opened 2022-07-25
Fix poll_capacity returning Ready(Some(0))

vadim-eg:
`poll_capacity` should never return `Some(0)` - because the caller has
no way to be woken up when send capacity is available.

Check capacity and return `Pending` if it is zero. The wakeup code is
already correctly triggers notification when capacity changes to non-zero.

### DIFF
--- a/src/proto/streams/send.rs
+++ b/src/proto/streams/send.rs
@@ -321,14 +321,13 @@ impl Send {
             return Poll::Ready(None);
         }
 
-        if !stream.send_capacity_inc {
-            stream.wait_send(cx);
-            return Poll::Pending;
+        let capacity = self.capacity(stream);
+        if capacity > 0 {
+            return Poll::Ready(Some(Ok(capacity)));
         }
 
-        stream.send_capacity_inc = false;
-
-        Poll::Ready(Some(Ok(self.capacity(stream))))
+        stream.wait_send(cx);
+        return Poll::Pending;
     }
 
     /// Current available stream send capacity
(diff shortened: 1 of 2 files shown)


seanmonstar:
Thanks for the PR! I'm curious, do you know if we have a unit test that exercises?

vadim-eg:
> Thanks for the PR! I'm curious, do you know if we have a unit test that exercises?

It looks like `poll_capacity_after_send_data_and_reserve[_with_max_send_buffer_size]` do attempt to test the condition  - frankly I haven't figured out the exact conditions triggering it.

seanmonstar:
It looks like in CI, this causes a test to hang:

> test `reserved_capacity_assigned_in_multi_window_updates` has been running for over 60 seconds

vadim-eg:
> It looks like in CI, this causes a test to hang:
> 
> > test `reserved_capacity_assigned_in_multi_window_updates` has been running for over 60 seconds

@seanmonstar Yes, I'll try to see what is going on.

vadim-eg:
@seanmonstar I think I got an explanation on the original issue  - Poll(Ready(Some(0))). It appears to be a concurrency/timing issue - it can happen when wakeup is called from one thread while the outer is sending data on a stream - e.g. - 
  - thread A - sender called poll_capacity, got 16K and proceeded to send_data
  - thread B - connection got a window update and sent out some pending frames. capacity did not increase but  wakeup was scheduled and send_capacity_inc set. (may be this can  e avoided)
  - thread A - sent the data and capacity is now 0, send_capacity_inc set
  - thread A  - poll_capacity returns 0

vadim-eg:
I think I have an idea what is going on  I'll open an issue and take it there

PR #638, opened 2022-08-23
Add a setter for `header_table_size`

### LINKED ISSUE #637: Add a setter for `header_table_size`
I'd like to make a PR to uncomment the setter for `header_table_size`, found here:
https://github.com/hyperium/h2/blob/87969c1f296173a7838956165014b4828dc5d5db/src/frame/settings.rs#L124-L128

When playing around with the setter I found out that connections would sometimes fail the following check, Since the decoder's size is never updated from the default `DEFAULT_SETTINGS_HEADER_TABLE_SIZE`.
https://github.com/hyperium/h2/blob/88b078925416d9d220b69d66192f7be63457d6b8/src/hpack/decoder.rs#L256-L258

Would [this](https://github.com/4JX/h2/commit/75f243426f40ae1cd51a05a5c51c76f7c5517cb3) be the place to fix that?

4JX:
The function's description may need some refinement

Closes #637

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1072,6 +1072,39 @@ impl Builder {
         self
     }
 
+    /// Sets the header table size.
+    ///
+    /// This setting informs the peer of the maximum size of the header compression
+    /// table used to encode header blocks, in octets. The encoder may select any value
+    /// equal to or less than the header table size specified by the sender.
+    ///
+    /// The default value is 4,096.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use tokio::io::{AsyncRead, AsyncWrite};
+    /// # use h2::client::*;
+    /// # use bytes::Bytes;
+    /// #
+    /// # async fn doc<T: AsyncRead + AsyncWrite + Unpin>(my_io: T)
+    /// # -> Result<((SendRequest<Bytes>, Connection<T, Bytes>)), h2::Error>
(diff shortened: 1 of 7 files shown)


seanmonstar (commented):
Nice work! What do you think of adding a test to `tests/h2-tests/src/client.rs`?

4JX:
> Nice work! What do you think of adding a test to `tests/h2-tests/src/client.rs`?

Used the file that most closely matched what you said.

seanmonstar:
Thanks for including the test! I think this is probably good to go... Though, I'd want to ask you, since you're thinking about this topic, you might have certain expected behavior in mind. Should we be testing for that behavior?

4JX:
What I need is covered by the current test (and small fix). That is, the setting is included in the `SETTINGS` frame with the value set by the user.

Might be worth to also test that it will throw a protocol error in the `server requested size > header_table_size` case?

alexforster:
@seanmonstar Is there anything I can do to help this get merged?

4JX:
Rebased onto master and squashed the commits together

4JX:
Just making sure this is not waiting on CI? Doesn't seem like the error is related to code touched by this PR.

seanmonstar:
Sorry, I fixed CI in #722, restarting it here.

PR #639, opened 2022-09-29
add accessor for StreamId u32

ehaydenr:


hawkw (approved):
I think it seems pretty reasonable to add this. It is a public API commitment that a `StreamId` will always be convertible into a `u32`, but since the HTTP/2 standard defines stream IDs as being 31 bits, I think that's a pretty safe commitment for this library to make...

LucioFranco:
Should it be a `From` impl or a `get() -> u32`?

LucioFranco:
Personally, I don't like the hidden from statements but we could just add a `get_u32` that is clear and verbose.

LucioFranco (approved):

PR #641, opened 2022-10-14
add wait for remote settings being set before sending requests

### LINKED ISSUE #512: expose information about negotiated connection settings

Yneth:
Fixes https://github.com/hyperium/h2/issues/512

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -534,6 +534,11 @@ where
     pub fn is_extended_connect_protocol_enabled(&self) -> bool {
         self.inner.is_extended_connect_protocol_enabled()
     }
+
+    /// Returns negotiated max send streams
+    pub fn max_send_streams(&self) -> usize {
+        self.inner.max_send_streams()
+    }
 }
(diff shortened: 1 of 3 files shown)


Yneth:
Is there anything I can do to fix pipeline and update PR to be reviewed and merged?

PR #643, opened 2022-10-24
Made mod hpack public as unstable feature. Allows using only the parsing part of h2 on an existing http2 message.

thesw4rm:


LucioFranco:
I mean would it just make sense to make it its own crate?

thesw4rm:
I can do that as well if hyperium is fine with such a big architectural change. Planning to throw in a few examples on how to parse a byte array into an http2 frame.

PR #649, opened 2022-11-27
Removed unnecessary `'static` & Unpin bounds from body generic

gtsiam:
## Motivation

When proxying large requests, I want to be able to pass the request body straight through my server without loading all of it it in memory at once. This is currently very difficult due to the (seemingly) arbitrary `'static` bound on Hyper's body generic. And so, I want to remove the `'static` bounds for hyper 1.0, but before I can do that I need to remove them from here. (Will this require an h2 release before I can make the hyper patch?)

## Changes

- Removed unnecessary bounds from body generic
- ~~Fixed all clippy warnings~~ (Moved to #652)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -341,7 +341,7 @@ pub(crate) struct Peer;
 
 impl<B> SendRequest<B>
 where
-    B: Buf + 'static,
+    B: Buf,
 {
     /// Returns `Ready` when the connection can initialize a new HTTP/2
     /// stream.
@@ -584,7 +584,7 @@ where
 
 impl<B> Future for ReadySendRequest<B>
 where
-    B: Buf + 'static,
+    B: Buf,
 {
     type Output = Result<SendRequest<B>, crate::Error>;
 
@@ -1100,7 +1100,7 @@ impl Builder {
     ) -> impl Future<Output = Result<(SendRequest<B>, Connection<T, B>), crate::Error>>
     where
         T: AsyncRead + AsyncWrite + Unpin,
-        B: Buf + 'static,
+        B: Buf,
     {
         Connection::handshake2(io, self.clone())
     }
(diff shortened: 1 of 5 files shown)


seanmonstar:
It looks like the change here is actually quite small (the clippy stuff seems unrelated and could be a separate PR). I can't remember why the static bounds are there in the first place...

.gitignore
```diff
@@ -8,3 +8,6 @@ h2spec
 # Files generated by honggfuzz
 hfuzz_target
 hfuzz_workspace
+
+# Editor-specific settings
+.vscode
```
LucioFranco, line 13: This should be moved to a system global gitignore instead of in each repo.

LucioFranco (changes requested):
Clippy changes like sean said should go in a separate PR but overall lgtm for the bound changes.

.gitignore
gtsiam: Huh. Honestly never though about doing that. I'll remove that one then.

gtsiam:
> the clippy stuff seems unrelated and could be a separate PR

I still have a pr sitting in warp for more than a year, so I was expecting this to take ages too. But I'll keep it in mind for future contributions to hyper. Probably more effort than it's worth splitting it now.

gtsiam:
On second thought I'll just make it a separate pr, no need for the extra back and forth

LucioFranco (approved):
If CI is happy then LGTM! This should be a non-breaking change as well I believe.

seanmonstar:
> I still have a pr sitting in warp for more than a year, so I was expecting this to take ages too.

gtsiam:
> > I still have a pr sitting in warp for more than a year, so I was expecting this to take ages too.
> 
> Sorry! I haven't been able to keep up with warp. My attention has been sucked up with hyper and its dependencies.

It's understandable, there's only so much time in a day. I don't use warp much anyway. It's just that warp & mime is what shaped my expectation that this would be sitting here forever and pushed me to dump everything in one pr (not making it very easy to review I'll admit).

That said, it might be worth looking into getting some maintainers for your undead projects or mark them as such at some point.

PR #651, opened 2022-12-01
fix issue#648: drop frame if stream is released

### LINKED ISSUE #648: Does the memory have residual memory?
Does the memory have residual memory?
https://github.com/hyperium/h2/blob/294000c0745c64009151a1ab39978cd6f02dfd68/src/proto/streams/recv.rs#L612

Currently, the count_buf method has been used to find that self.buffer has accumulated data.
```
    pub fn poll_data(
        &mut self,
        cx: &Context,
        stream: &mut Stream,
    ) -> Poll<Option<Result<Bytes, proto::Error>>> {
        // TODO: Return error when the stream is reset
        match stream.pending_recv.pop_front(&mut self.buffer) {
            Some(Event::Data(payload)) => {
                self.count_buf(); 
                Poll::Ready(Some(Ok(payload)))
            },
    // ignore
}
    fn count_buf(&mut self) {
        if self.time.elapsed().as_secs() < 2 {
            return;
        }
        self.time = Instant::now();
        let mut count = 0;
        let mut body_count = 0;
        for (k, s) in self.buffer.slab.iter() {
            match &s.value {
                Event::Data(payload) => {
                    count += 1;
                    body_count += 1;
                },
                _ => {
                    count += 1;
                }
            }
        }
        println!("------------pull {}, {}", count, body_count);
    }
```

When the client has only one request sent per second, the self.buffer count is still not released.
image

Scenarios：
client:
```
use std::time::{Duration, SystemTime};
use hyper::{Body, Client, Request};
use hyper::body::Bytes;
use hyper::client::HttpConnector;
use hyper::header::CONTENT_LENGTH;

#[tokio::main]
async fn main() {
    let client = hyper::client::Client::builder().http2_only(true).build_http();
    for i in 0..20 {
        let client = client.clone();
       tokio::spawn(async move {
           for i in 0..100 {
               send_req(client.clone()).await;
               tokio::time::sleep(Duration::from_millis(50)).await;
           }
           println!("xxx exit");
       });
        tokio::time::sleep(Duration::from_millis(20)).await;
    }
    for i in 0..100 {
        send_req(client.clone()).await;
        tokio::time::sleep(Duration::from_secs(1)).await;
    }
    tokio::time::sleep(Duration::from_secs(100)).await;
}

async fn send_req(client: Client<HttpConnector>) {
    let (mut sender, body) = Body::channel();
    let num = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).unwrap().as_millis() % 1000;
    tokio::spawn(async move {
        for i in 0..num {
            sender.send_data("ss".into()).await;
        }
        sender.send_data(Bytes::new()).await;
    });
    let request = Request::builder().uri("http://127.0.0.1:8081")
        .header(CONTENT_LENGTH, (2 * num).to_string()).body(body).unwrap();
    let response = client.request(request).await.unwrap();
    if !response.status().is_success() {
        println!("{:?}", response);
    }
}
```
server:
```
pub async fn handle(mut req: Request<Body>) -> Result<Response<Body>, Error> {
    // println!("xxx {:?}", req);
    if req.version() == Version::HTTP_11 {
        hyper::body::to_bytes(req.into_body()).await;
        return Ok(Response::new(Body::from("sss")));
    }
    let client = hyper::client::Client::new();
    *req.version_mut() = Version::HTTP_11;
    let result = client.request(req).await;
    if result.is_err() {
        println!("{:?}", result);
    }
    let mut response = Response::new(Body::from("sss"));
    response
        .headers_mut()
        .insert("content-type", "application/json".parse().unwrap());
    hyper::Result::Ok(response)
}
```

For the crate of bytes, even if there is only one empty Bytes reference, the corresponding shared memory is not released.
image

silence-coding:
fix #648

### DIFF
--- a/src/proto/streams/recv.rs
+++ b/src/proto/streams/recv.rs
@@ -600,6 +600,16 @@ impl Recv {
             }
         }
 
+        // Received a frame, but no one cared about it. fix issue#648
+        if !stream.is_recv {
+            tracing::trace!(
+                "recv_data; frame ignored on stream release {:?} for some time",
+                stream.id,
+            );
+            self.release_connection_capacity(sz, &mut None);
+            return Ok(());
+        }
+
         // Update stream level flow control
         stream.recv_flow.send_data(sz);
 
--- a/src/proto/streams/stream.rs
+++ b/src/proto/streams/stream.rs
@@ -99,6 +99,9 @@ pub(super) struct Stream {
     /// Frames pending for this stream to read
     pub pending_recv: buffer::Deque,
 
+    /// When the RecvStream drop occurs, no data should be received.
+    pub is_recv: bool,
+
     /// Task tracking receiving frames
     pub recv_task: Option<Waker>,
 
@@ -180,6 +183,7 @@ impl Stream {
             reset_at: None,
             next_reset_expire: None,
             pending_recv: buffer::Deque::new(),
+            is_recv: true,
             recv_task: None,
             pending_push_promises: store::Queue::new(),
             content_length: ContentLength::Omitted,
--- a/src/proto/streams/streams.rs
+++ b/src/proto/streams/streams.rs
@@ -1345,12 +1345,13 @@ impl OpaqueStreamRef {
             .release_capacity(capacity, &mut stream, &mut me.actions.task)
     }
 
+    /// Clear the receive queue and set the status to no longer receive data frames.
     pub(crate) fn clear_recv_buffer(&mut self) {
         let mut me = self.inner.lock().unwrap();
         let me = &mut *me;
 
         let mut stream = me.store.resolve(self.key);
-
+        stream.is_recv = false;
         me.actions.recv.clear_recv_buffer(&mut stream);
     }
 


silence-coding:
@seanmonstar  Please!

seanmonstar:
Code looks good, letting CI run. I guess there isn't really a unit test we can do, huh? Since it's mostly just about reducing memory usage...

silence-coding:
> 代码看起来不错，让CI运行。我想我们没有真正的单元测试可以做，是吧？因为它主要是为了减少内存使用...

Yes, mainly to reduce memory

silence-coding:
> Code looks good, letting CI run. I guess there isn't really a unit test we can do, huh? Since it's mostly just about reducing memory usage...

Yes, mainly to reduce memory @seanmonstar

seanmonstar (approved):

maxbear1988:
@seanmonstar cloud you release a new version?

PR #652, opened 2022-12-01
Fix all clippy warnings

gtsiam:
Fixes all clippy warnings.

Split from #649.


src/frame/headers.rs
```diff
@@ -309,17 +309,20 @@ impl fmt::Debug for Headers {
 
 // ===== util =====
 
-pub fn parse_u64(src: &[u8]) -> Result<u64, ()> {
+#[derive(Debug)]
+pub struct ParseU64Error;
+
+pub fn parse_u64(src: &[u8]) -> Result<u64, ParseU64Error> {
```
LucioFranco, line 315: im not totally sure about this change...can you point to the clippy lint?

src/frame/headers.rs
```diff
@@ -364,9 +367,9 @@ impl PushPromise {
         // A promised request "that indicates the presence of a request body
         // MUST reset the promised stream with a stream error"
         if let Some(content_length) = req.headers().get(header::CONTENT_LENGTH) {
-            let parsed_length = parse_u64(content_length.as_bytes());
-            if parsed_length != Ok(0) {
-                return Err(InvalidContentLength(parsed_length));
+            match parse_u64(content_length.as_bytes()) {
+                Ok(0) => (),
+                parsed => return Err(InvalidContentLength(parsed)),
```
LucioFranco, line 372: This one seems like its almost harder to read? I prefer the current code here.

src/frame/settings.rs
```diff
@@ -88,7 +88,7 @@ impl Settings {
 
     pub fn set_max_frame_size(&mut self, size: Option<u32>) {
         if let Some(val) = size {
-            assert!(DEFAULT_MAX_FRAME_SIZE <= val && val <= MAX_MAX_FRAME_SIZE);
+            assert!((DEFAULT_MAX_FRAME_SIZE..=MAX_MAX_FRAME_SIZE).contains(&val));
```
LucioFranco, line 91: Same here I feel like the readability of this is worse here

src/frame/settings.rs
```diff
@@ -182,10 +182,10 @@ impl Settings {
                     }
                 }
                 Some(MaxFrameSize(val)) => {
-                    if val < DEFAULT_MAX_FRAME_SIZE || val > MAX_MAX_FRAME_SIZE {
-                        return Err(Error::InvalidSettingValue);
-                    } else {
+                    if (DEFAULT_MAX_FRAME_SIZE..=MAX_MAX_FRAME_SIZE).contains(&val) {
```
LucioFranco, line 185: Same here

src/hpack/decoder.rs
```diff
@@ -396,7 +396,7 @@ fn decode_int<B: Buf>(buf: &mut B, prefix_size: u8) -> Result<usize, DecoderErro
     const VARINT_MASK: u8 = 0b0111_1111;
     const VARINT_FLAG: u8 = 0b1000_0000;
 
-    if prefix_size < 1 || prefix_size > 8 {
+    if !(1..=8).contains(&prefix_size) {
```
LucioFranco, line 399: I don't like this lint tbh....we can just allow it at the root of the project imho

src/proto/streams/streams.rs
```diff
@@ -1270,6 +1270,7 @@ impl OpaqueStreamRef {
         me.actions.recv.poll_response(cx, &mut stream)
     }
     /// Called by a client to check for a pushed request.
+    #[allow(clippy::type_complexity)]
```
LucioFranco, line 1273: can prob just add this to the root... I don't really like this lint tbh

LucioFranco (changes requested):
LGTM but left some comments, we should back out some of the clippy changes and add them to a global allow.

gtsiam:
I do not understand why github thought I wanted to start a review, but anyway - the replies are visible if you just skip the mess at the end. Let me make all the changes.

gtsiam:
Also while I'm at it: Would you like me to add a style check to the ci so there's no need to do this kind of large pr again?

src/frame/headers.rs
gtsiam: I find using matches for these kinds of checks much easier cognitively since I don't have to reason about individual checks. So I'd read the new code as:

- parse a u64
- if it's Ok(0), all good
- otherwise, error out

Contrast with:

- parse a u64
- if it isn't an Ok(0), error out (So you have to think that Ok(0) is the good case)
- otherwise continue (?)

It's wasn't bad code, just that the extra not you have to do had me doing mental gymnastics. Admittedly I wrote this late at night so I can change it back if you prefer.

src/frame/settings.rs
gtsiam: It's this lint: https://rust-lang.github.io/rust-clippy/master/index.html#manual_range_contains

Honestly, I'm inclined to agree; writing those did require some mental gymnastics. And I got one wrong at first and didn't realize until I run hyper's test suite.

src/frame/settings.rs
gtsiam: And same response.

Though it's probably worth keeping the flipped if. I had to apply de morgan's rule to understand what that really meant at first which is never a good sign.

src/hpack/decoder.rs
gtsiam: Yeah lol. I'll add an allow for it in lib.rs and revert those

src/proto/streams/streams.rs
gtsiam: will do

src/frame/headers.rs
LucioFranco: I think the if statement feels pretty normal to check if this is wrong and return an error otherwise continue with the flow. My issue is with the branch returning the unit type just feels extra not needed. I would like this change backed out.

src/frame/headers.rs
LucioFranco: We can keep this for now but this is just an example of a clippy lint I don't totally agree with.

gtsiam: I'd much prefer seeing `Err(ParseU64Error)` to `Err(())` in logs. Private code shouldn't be a free pass for odd api decisions, since it can always leak to the user one way or another (logs in this case).

Then again, since you've agreed to keep this I'm not really arguing for anything so I'll just get on with it and resolve the conversation.

src/frame/headers.rs
gtsiam: I strongly disagree: Matches generally make control flow much easier to follow... But it really doesn't matter. I wasn't even planning to change this at first. I just changed it while trying to make sense of it and forgot about it. So I'll just back it out.

gtsiam:
> You mean run clippy in ci? I personally don't like it since it adds a lot of noise.

Yes. Though I really wouldn't call clippy warnings "noise". With the exception of the `manual_range_contains` lint, most of the changes in this pr should never have needed to be done in mass. Catching them earlier would be much easier for all involved.

But I'm not gonna push for it. You're the maintainer. If you don't want it, you don't want it and that's the end of that.

LucioFranco (approved):
LGTM thanks!

PR #654, opened 2022-12-14
server: Ignore `NotConnected` errors on shutdown

LucioFranco:
Related to https://github.com/hyperium/tonic/issues/1183

I started to see this error come up when debugging the above tonic issue, to me ignoring this error seems okay in this context, but I'd like to have some other eyes on it.

cc @seanmonstar @hawkw

### DIFF
--- a/src/proto/connection.rs
+++ b/src/proto/connection.rs
@@ -283,11 +283,16 @@ where
                     tracing::trace!("connection closing after flush");
                     // Flush/shutdown the codec
                     if let Err(e) = ready!(self.codec.shutdown(cx)) {
-                        // If the error kind is NotConnected then ignore that
-                        // since it means the connection is already shutdown.
+                        // If the error kind is NotConnected, ignore it, since
+                        // it just means the connection is already shutdown.
                         if e.kind() != io::ErrorKind::NotConnected {
                             return Poll::Ready(Err(e.into()));
-                        }
+                        } else {
+                            tracing::trace!(
+                                "ignoring NotConnected error \
+                                (the connection has already closed)"
+                           );
+                       }
                     }
 
                     // Transition the state to error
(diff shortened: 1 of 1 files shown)


src/proto/connection.rs
```diff
@@ -282,7 +282,13 @@ where
                 State::Closing(reason, initiator) => {
                     tracing::trace!("connection closing after flush");
                     // Flush/shutdown the codec
-                    ready!(self.codec.shutdown(cx))?;
+                    if let Err(e) = ready!(self.codec.shutdown(cx)) {
+                        // If the error kind is NotConnected then ignore that
+                        // since it means the connection is already shutdown.
```
hawkw, line 287: nit: maybe:
```suggestion
                        // If the error kind is NotConnected, ignore it, since
                        // it just means the connection is already shutdown.
```

src/proto/connection.rs
```diff
@@ -282,7 +282,13 @@ where
...
+                        if e.kind() != io::ErrorKind::NotConnected {
+                            return Poll::Ready(Err(e.into()));
+                        }
```
hawkw, line 290: what do you think about logging something at TRACE here, so if a user is looking at very verbose logs, they would be able to determine that an error was ignored here?
```suggestion
                        } else {
                            tracing::trace!(
                                "ignoring NotConnected error \
                                (the connection has already closed)"
                           );
                       }
```

hawkw (approved):
i think it seems fine to ignore the error in this case. the idea of completely ignoring an error makes me mildly uncomfortable, so _maybe_ we should be logging it? otherwise, lgtm.

seanmonstar:
I'm not so sure we should choose to ignore the error. We don't know the full context of the IO transport. It feels more correct to me to pass that error along to the user, and *they* can decide if to ignore it or not.

seanmonstar (changes requested):
My concern still [exists](https://github.com/hyperium/h2/pull/654#issuecomment-1352140926)

LucioFranco:
> I'm not so sure we should choose to ignore the error. We don't know the full context of the IO transport. It feels more correct to me to pass that error along to the user, and they can decide if to ignore it or not.

Sounds good to me

PR #655, opened 2022-12-27
Add Protocol extension to Request on Extended CONNECT

### LINKED ISSUE #347: Support WebSocket upgrade
[RFC 8441](https://tools.ietf.org/html/rfc8441) specifies the upgrade mechanism to WebSocket over HTTP/2. Since it includes several extensions to the HTTP/2 protocol, the server API might need to support the configuration of them.

cloneable:
This exposes the :protocol pseudo header as Request extension.

Fixes #347

### DIFF
--- a/tests/h2-tests/tests/server.rs
+++ b/tests/h2-tests/tests/server.rs
@@ -1214,7 +1214,12 @@ async fn extended_connect_protocol_enabled_during_handshake() {
 
         let mut srv = builder.handshake::<_, Bytes>(io).await.expect("handshake");
 
-        let (_req, mut stream) = srv.next().await.unwrap().unwrap();
+        let (req, mut stream) = srv.next().await.unwrap().unwrap();
+
+        assert_eq!(
+            req.extensions().get::<crate::ext::Protocol>(),
+            Some(&crate::ext::Protocol::from_static("the-bread-protocol"))
+        );
 
         let rsp = Response::new(());
         stream.send_response(rsp, false).unwrap();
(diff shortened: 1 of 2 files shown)


cloneable:
@nox, have you had a chance to look at this?

cloneable:
@seanmonstar, looks like @nox hasn't been active for months now. Can you recommend another reviewer?
Btw, how does the following even work in hyper without this PR?
https://github.com/hyperium/hyper/blob/5bf16402f7a146417a9f997387401f9dc5a7d63d/src/proto/h2/server.rs#L291-L293

seanmonstar:
> how does the following even work in hyper without this PR?

Hm, well, the code itself is just conditional, _if_ the extension exists, replace it with another. So it never _fails_. But I imagine we don't have a test anywhere that it ever _succeeds_, either. Would be worth making one, I think. I'll review this shortly.

tests/h2-tests/tests/server.rs
```diff
@@ -1214,7 +1214,12 @@ async fn extended_connect_protocol_enabled_during_handshake() {
 
         let mut srv = builder.handshake::<_, Bytes>(io).await.expect("handshake");
 
-        let (_req, mut stream) = srv.next().await.unwrap().unwrap();
+        let (req, mut stream) = srv.next().await.unwrap().unwrap();
+
+        assert_eq!(
```
seanmonstar, line 1219: Just to confirm, this assertion fails without the above change, right?

cloneable: Yes, sorry, this would fail.

seanmonstar (approved):

PR #657, opened 2023-01-06
fix: panic in pop_frame()

aftersnow:
We met the panic in our production environment, so handle this panic
    condition before panic. stack backtrace:
       0: rust_begin_unwind
                 at
    /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:517:5
       1: core::panicking::panic_fmt
                 at
    /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:101:14
       2: core::panicking::panic
                 at
    /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:50:5
       3: h2::proto::streams::flow_control::FlowControl::send_data
                 at
    /build/vendor/h2/src/proto/streams/flow_control.rs:176:9
       4: h2::proto::streams::prioritize::Prioritize::pop_frame::{{closure}}
                 at
    /build/vendor/h2/src/proto/streams/prioritize.rs:737:33
       5: tracing::span::Span::in_scope
                 at /build/vendor/tracing/src/span.rs:982:9
       6: h2::proto::streams::prioritize::Prioritize::pop_frame
                 at
    /build/vendor/h2/src/proto/streams/prioritize.rs:736:29
       7: h2::proto::streams::prioritize::Prioritize::poll_complete
                 at
    /build/vendor/h2/src/proto/streams/prioritize.rs:497:19
       8: h2::proto::streams::send::Send::poll_complete
                 at /build/vendor/h2/src/proto/streams/send.rs:297:9
       9: h2::proto::streams::streams::Inner::poll_complete
                 at /build/vendor/h2/src/proto/streams/streams.rs:850:16
      10: h2::proto::streams::streams::Streams<B,P>::poll_complete
                 at /build/vendor/h2/src/proto/streams/streams.rs:180:9
      11: h2::proto::connection::Connection<T,P,B>::poll
                 at /build/vendor/h2/src/proto/connection.rs:253:36
      12: <h2::client::Connection<T,B> as
    core::future::future::Future>::poll
      13: ...

### DIFF
--- a/src/proto/streams/flow_control.rs
+++ b/src/proto/streams/flow_control.rs
@@ -173,12 +173,15 @@ impl FlowControl {
             self.available
         );
 
-        // Ensure that the argument is correct
-        assert!(self.window_size >= sz as usize);
+        // If send size is zero it's meaningless to update flow control window
+        if sz > 0 {
+            // Ensure that the argument is correct
+            assert!(self.window_size >= sz as usize);
 
-        // Update values
-        self.window_size -= sz;
-        self.available -= sz;
+            // Update values
+            self.window_size -= sz;
+            self.available -= sz;
+        }
     }
 }
 
--- a/src/proto/streams/prioritize.rs
+++ b/src/proto/streams/prioritize.rs
@@ -744,6 +744,14 @@ impl Prioritize {
                             // capacity at this point.
                             debug_assert!(len <= self.flow.window_size());
 
+                            // Check if the stream level window the peer knows is available. In some
+                            // scenarios, maybe the window we know is available but the window which
+                            // peer knows is not.
+                            if len > 0 && len > stream.send_flow.window_size() {
+                                stream.pending_send.push_front(buffer, frame.into());
+                                continue;
+                            }
+
                             tracing::trace!(len, "sending data frame");
 
                             // Update the flow control


aftersnow:
#607

src/proto/streams/flow_control.rs
```diff
@@ -173,12 +178,15 @@ impl FlowControl {
             self.available
         );
 
-        // Ensure that the argument is correct
-        assert!(self.window_size >= sz as usize);
+        // If send size is zero it's meaningless to update flow control window
```
seanmonstar, line 181: Is this change truly needed? If the value is zero, the assert and subtractions shouldn't do anything harmful, right?

src/proto/streams/flow_control.rs
```diff
@@ -61,6 +61,11 @@ impl FlowControl {
         self.window_size.as_size()
     }
 
+    /// Check if send window is available
+    pub fn window_available(&mut self, sz: WindowSize) -> bool {
```
seanmonstar, line 65: The name of this function could be confusing, since there is both `window_size()` and `available()`, so a function `window_available()` sounds like a combination, while at the same time possible a question. I'd suggest leaving this function off, and just doing `if flow.window_size() >= sz` wherever it's needed.

src/proto/streams/prioritize.rs
```diff
@@ -744,6 +744,14 @@ impl Prioritize {
                             // capacity at this point.
                             debug_assert!(len <= self.flow.window_size());
 
+                            // Check if the stream level window the peer knows is available. In some
+                            // scenarios, maybe the window we know is available but the window which
+                            // peer knows is not.
+                            if len > 0 && !stream.send_flow.window_available(len) {
```
seanmonstar, line 750: I'd probably lean towards removing the `len > 0` part.

I see that `stream.send_flow.available()` is checked a few lines above, so the problem is that there's "available", but not enough `window_size`? Am I understanding the problem here?

(Wow, our choice of naming those fields could have been better. The current names don't distinguish what they really mean that well.)

seanmonstar (changes requested):
Thanks so so much for sending a PR! It's always appreciated when people try to help fix the problems they find :)

src/proto/streams/prioritize.rs
aftersnow: The "len > 0" part is for prevent zero-length frame from being blocked, which should not.

src/proto/streams/flow_control.rs
seanmonstar: Oh true, very good catch. I had forgotten about negative sizes.

aftersnow:
@seanmonstar Hi, have changed the PR, could you review the PR again please?

seanmonstar:
Sorry for forgetting about this. I think fixing this issue is a really nice improvement. I wonder if adding a unit test making sure we never regress this would be worth doing. I suppose the instance is a case where perhaps the connection window size is larger than the accumulative stream windows, right? What do you think? Does it look manageable to add a test case?

Have you, by chance, tried out this patch in your application to see it fixed the panics?

aftersnow:
> 

OK, I fixed the test error: If the frame length equals the windows size, it should not be pushed front. Now all tests passed. (Maybe the current UTs are enough?)

> I suppose the instance is a case where perhaps the connection window size is larger than the accumulative stream windows, right?

Yes, I found the error in our production environment (see the commit message). But I am not familiar with the http protocol, so I cannot explain it deeply. Do you have any suggestions for me?

> Does it look manageable to add a test case?

It would be better to add a test case, but I don't know how to reproduce the case...

> Have you, by chance, tried out this patch in your application to see it fixed the panics?

Yes, the initial commit of this PR has been tested in our production environment, and fixed the panic.

seanmonstar (approved):
Thanks for keeping up with this, LGTM!

PR #668, opened 2023-04-13
fix: limit the amount of pending-accept reset streams

seanmonstar:
Streams that have been received by the peer, but not accepted by the user, can also receive a RST_STREAM. This is a legitimate pattern: one could send a request and then shortly after, realize it is not needed, sending a CANCEL.

However, since those streams are now "closed", they don't count towards the max concurrent streams. So, they will sit in the accept queue, using memory.

In most cases, the user is calling `accept` in a loop, and they can accept requests that have been reset fast enough that this isn't an issue in practice.

But if the peer is able to flood the network faster than the server accept loop can run (simply accepting, not processing requests; that tends to happen in a separate task), the memory could grow.

So, this introduces a maximum count for streams in the pending-accept but remotely-reset state. If the maximum is reached, a GOAWAY frame with the error code of ENHANCE_YOUR_CALM is sent, and the connection marks itself as errored.

Closes https://github.com/hyperium/hyper/issues/2877 (we assume)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -326,6 +326,10 @@ pub struct Builder {
     /// Maximum number of locally reset streams to keep at a time.
     reset_stream_max: usize,
 
+    /// Maximum number of remotely reset streams to allow in the pending
+    /// accept queue.
+    pending_accept_reset_stream_max: usize,
+
     /// Initial `Settings` frame to send as part of the handshake.
     settings: Settings,
 
@@ -634,6 +638,7 @@ impl Builder {
             max_send_buffer_size: proto::DEFAULT_MAX_SEND_BUFFER_SIZE,
             reset_stream_duration: Duration::from_secs(proto::DEFAULT_RESET_STREAM_SECS),
             reset_stream_max: proto::DEFAULT_RESET_STREAM_MAX,
+            pending_accept_reset_stream_max: proto::DEFAULT_REMOTE_RESET_STREAM_MAX,
             initial_target_connection_window_size: None,
             initial_max_send_streams: usize::MAX,
             settings: Default::default(),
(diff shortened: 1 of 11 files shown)


seanmonstar:
I started to make the limit configurable, but then paused, wanting to get something working faster. If it seems worth the delay, I can add in the public API to set these limits yourself.

LPardue:
Since this seems like something that occurs due to several dynamic factors or behaviours, a single hard coded limit _on paper_ seems too restrictive. Not least because it isn't immediately obvious where the value `20` comes from. This might be an entirely reasonable value but without some wider testing, who knows? A default value that can be tweaked via API seems like a decent middleground to me.

Noah-Kennedy (approved):
I'm in favor of merging this now as is, but we should in the future make the threshold configurable.

seanmonstar:
Added another commit making the options part of the public API.

src/client.rs
```diff
@@ -966,6 +971,37 @@ impl Builder {
         self
     }
 
+    /// Sets the maximum number of pending-accept remotely-reset streams.
+    ///
+    /// When a
```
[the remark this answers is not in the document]
seanmonstar: oh woops, I was supposed to copy the explanation from the server builder! 😴

Noah-Kennedy:
I'm summarizing a conversation @seanmonstar and I had in Discord here:

There is a non-zero chance that the threshold we set does trip on real (and non-malicious) traffic. We have a log line in here for if this goes off, but it might be worthwhile for us to expose an actual metric for these GOAWAYs so that users of the library can better understand how frequently this trips.

seanmonstar:
Given that I've been at this for 14 hours, I'm going to go knock out. Since the assumed problem is quite hard to trigger, and that the fix here is a guess, and that there's always some amount of risk when publishing new versions, I'm opting for leaving this open for the night.

Noah-Kennedy:
This will also give our peers outside of North America a chance to look at this change before we release it.

If anyone has any concerns about this change, please make them heard.

edevil (approved):

src/proto/streams/recv.rs
```diff
@@ -741,12 +741,39 @@ impl Recv {
...
+        // See https://github.com/hyperium/hyper/issues/2877
+        if stream.is_pending_accept {
+            if counts.can_inc_num_remote_reset_streams() {
+                counts.inc_num_remote_reset_streams();
+            } else {
+                tracing::warn!(
+                    "recv_reset; remotely-reset pending-accept streams reached limit ({:?})",
+                    counts.max_remote_reset_streams(),
+                );
+                return Err(Error::library_go_away(Reason::ENHANCE_YOUR_CALM));
```
nox, line 766: Can you drive-by expose `Error::is_library` or something like that, so we can check for `error.is_library() && error.reason() == Some(Reason::ENHANCE_YOUR_CALM)` for logging purposes?

seanmonstar: Would checking `!error.is_remote() && error.is_go_away()` be fine?

nox: There are 3 error initiators: remote, user and library, and there is no way to check for the user one either. If you don't want to expose `is_library` I'm fine with checking `!error.is_remote() && !error.is_user() && error.is_go_away() && error.reason() == Some(Reason::ENHANCE_YOUR_CALM)`

seanmonstar: We _can_ add the method. I was just wondering if what exists is enough.

As an alternative (or in addition to), we could add some debug data to GOAWAY frame, maybe `too_many_resets` or something. Oh, just looking, a method doesn't exist to access that either...

seanmonstar: Made in a separate PR: https://github.com/hyperium/h2/pull/669

PR #673, opened 2023-04-14
#3000: Serialize debug_data when present in GOAWAY frames

Herbstein:
Fixes hyperium/hyper#3200

Debug data in GOAWAY frames is opaque information meant for diagnostic purposes. The data was previously read and stored when being received, but was never written during encoding.

This pull request includes the data when encoding the frame but does not add a (stable) way of setting the data. If it is desired I am happy to include it too. I suspect the function added for testing purposes wouldn't be fit for a stable, public API.

Additionally, I've included an integration test as described in the original issue. However, I'm quite sure there is a better way of creating the GOAWAY frame, I'm just not familiar enough with the structure of the integration tests to find it at the moment. With a few pointers I'd like to rewrite how it is done, if needed.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -544,6 +544,12 @@ where
         self.connection.go_away_gracefully();
     }
 
+    #[doc(hidden)]
+    #[cfg(feature = "unstable")]
+    pub fn debug_data_shutdown(&mut self) {
+        self.connection.go_away_debug_data();
+    }
+
     /// Takes a `PingPong` instance from the connection.
     ///
     /// # Note
--- a/src/frame/go_away.rs
+++ b/src/frame/go_away.rs
@@ -8,7 +8,6 @@ use crate::frame::{self, Error, Head, Kind, Reason, StreamId};
 pub struct GoAway {
     last_stream_id: StreamId,
     error_code: Reason,
-    #[allow(unused)]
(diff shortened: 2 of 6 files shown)


src/server.rs
```diff
@@ -544,6 +544,12 @@ where
         self.connection.go_away_gracefully();
     }
 
+    #[doc(hidden)]
+    #[cfg(feature = "unstable")]
+    pub fn debug_data_shutdown(&mut self) {
```
seanmonstar, line 549: I realized right as I was finishing up the checklist in that issue that coming up with a unit test would be harder, since there would be no real way to externally test it _works_. So I appreciate trying to come up with one 😅  And that this existing to begin with as "unstable" for testing purposes seems fine.

I meant to file a follow-up issue after, adding debug data to a specific GOAWAY case, which in that once a test exists, we could remove this stuff, I suppose?

Herbstein: Adding the test wasn't too hard, to be honest. It was just copying what was already done and modifying as needed. The hardest part was figuring out _where_ the test should be, and how to run it, and that didn't take long either.

As soon as an actual use of debug data is added I agree that all of that stuff should be ripped out again, in favor of code paths

seanmonstar (approved):

PR #675, opened 2023-04-16
use HeaderName::from_static in Send::check_headers

srijs:
Prevents parsing the header names (due to the `AsHeaderName` impl on `&str`) over and over again at runtime. It's just a small optimization in my case (sub-1% cpu time) but given the simplicity of the change seemed worth making in any case.


hawkw (approved):
👍

srijs:
@hawkw thanks for the quick approve! I updated the branch based on the failed build, but looks like this needs another approval to re-run the tests?

hawkw:
re-approved, hopefully everything passes this time!

Issue #548, opened 2021-07-23
"protocol error: unspecific protocol error detected" making http2 request to go.googlesource.com


kevinburke:
I compiled curl with rustls-ffi and Hyper. I'm attempting to make a request to https://go.googlesource.com. Here's what I get when I try that:

```
$ curl --version
curl 7.78.0 (x86_64-apple-darwin19.6.0) libcurl/7.78.0 crustls/0.7.1/rustls/0.19.0 zlib/1.2.11 brotli/1.0.9 zstd/1.5.0 libidn2/2.3.2 librtmp/2.3 Hyper/0.14.11 OpenLDAP/2.5.5
Release-Date: 2021-07-21
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtmp smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli Debug GSS-API HSTS HTTP2 IDN IPv6 Kerberos Largefile libz SPNEGO SSL TrackMemory UnixSockets zstd

$ curl https://go.googlesource.com
curl: (56) Hyper: [1] error reading a body from connection: protocol error: unspecific protocol error detected
```

Full `-vvv` logs are available here: https://gist.github.com/kevinburke/c2ac5cf3ed7f59c650263288010ccdce

I'm compiling curl on Mac using the Homebrew formula available at github.com/meterup/homebrew-safe. You can install using `brew install meterup/safe/curl`.

I tracked down that "unspecific protocol error" message to this library, so I'm hopeful that this is the right place to file a ticket. I'm also hopeful that you are able to reproduce the issue.

I'm not sure what your policy is on h2 handling HTTP/2 implementations in the wild... it's possible that the server is doing something wrong, but I don't know enough about HTTP/2 to say for certain. The server is running Gitiles (https://gerrit.googlesource.com/gitiles/), though I'm not sure if that service is also terminating TLS, or if there's some other layer in between that's handling that.

Updates rustls/rustls-ffi#129.

seanmonstar:
I'm all for improving interop with HTTP2 implementations. So, first thing I did was try this using a pure Rust implementation. I used [reqwest](https://github.com/seanmonstar/reqwest)'s simple example, with default-tls disabled and rustls enabled, so it would use ALPN and HTTP2. It works correctly for me... I get a `200 OK` response, and the full HTML body.

Does this happen with other hosts that use HTTP2?

kevinburke:
[...]

PR #679, opened 2023-04-27
fix: avoid protocol error for request with host header

everpcpc:
ref: #548
https://github.com/seanmonstar/reqwest/issues/1809

http2 server in google is strict with headers, and a request with `HOST` in header fields would result in `PROTOCOL_ERROR`, which is not able to be handled well in hyper and other upper http client libraries.

> The ":authority" pseudo-header field conveys the authority portion ([Section 3.2](https://www.rfc-editor.org/rfc/rfc3986.html#section-3.2) of [[RFC3986]](https://httpwg.org/specs/rfc9113.html#RFC3986)) of the target URI ([Section 7.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-7.1) of [[HTTP]](https://httpwg.org/specs/rfc9113.html#RFC9110)). The recipient of an HTTP/2 request MUST NOT use the Host header field to determine the target URI if ":authority" is present.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -138,6 +138,7 @@
 use crate::codec::{Codec, SendError, UserError};
 use crate::ext::Protocol;
 use crate::frame::{Headers, Pseudo, Reason, Settings, StreamId};
+use crate::hpack::BytesStr;
 use crate::proto::{self, Error};
 use crate::{FlowControl, PingPong, RecvStream, SendStream};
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR!

As I mentioned in the linked reqwest issue, since it's legal to send both an `:authority` and `host` header, with the spec even explicitly saying "intermediaries MAY forward them on", then `h2` should **not** remove them.

PR #682, opened 2023-05-10
Fix Counts::dec_num_remote_reset_streams assertion failure

nox:


nox:
Cc @Noah-Kennedy

nox:
I can't manage to write a test that fails without the fix though :(

nox:
Added a test, the key to it was to keep around an ongoing request so `h2.poll_closed()` doesn't actually close the connection.

seanmonstar (approved):

PR #687, opened 2023-06-05
fix: do not ignore result of `ensure_recv_open`

DDtKey:
It may cause hanging, because result isn't checked
Noticed within #600

### DIFF
--- a/src/proto/streams/recv.rs
+++ b/src/proto/streams/recv.rs
@@ -308,7 +308,9 @@ impl Recv {
             Some(Event::Headers(Client(response))) => Poll::Ready(Ok(response)),
             Some(_) => panic!("poll_response called after response returned"),
             None => {
-                stream.state.ensure_recv_open()?;
+                if !stream.state.ensure_recv_open()? {
+                    return Poll::Ready(Ok(Response::new(())))
+                }
 
                 stream.recv_task = Some(cx.waker().clone());
                 Poll::Pending
--- a/src/proto/streams/streams.rs
+++ b/src/proto/streams/streams.rs
@@ -726,7 +726,10 @@ impl Inner {
                 }
 
                 // The stream must be receive open
-                stream.state.ensure_recv_open()?;
+                if !stream.state.ensure_recv_open()? {
+                    return Ok(())
+                }
+
                 stream.key()
             }
             None => {


src/proto/streams/recv.rs
```diff
@@ -308,7 +308,9 @@ impl Recv {
             Some(Event::Headers(Client(response))) => Poll::Ready(Ok(response)),
             Some(_) => panic!("poll_response called after response returned"),
             None => {
-                stream.state.ensure_recv_open()?;
+                if !stream.state.ensure_recv_open()? {
+                    return Poll::Ready(Ok(Response::new(())));
+                }
```
seanmonstar, line 313: So, if recv is NOT open, should it be returning a new empty response? Or is that an error in this context? It would be good to include a code comment here to remind us why, in either case.

src/proto/streams/streams.rs
```diff
@@ -726,7 +726,10 @@ impl Inner {
                 }
 
                 // The stream must be receive open
-                stream.state.ensure_recv_open()?;
+                if !stream.state.ensure_recv_open()? {
```
seanmonstar, line 729: Similar question here.

seanmonstar (commented):
(Wow, a function returning a `Result<bool>` is extremely confusing XD)

src/proto/streams/recv.rs
seanmonstar: Hm. Is there a way to build a test for this? That might help make it clearer what should actually be done.

DDtKey: Yeah, that would be nice, however it looks a bit complicated to me.

These two remaining methods are related to client-side and they just ignored, while server-side ones are handled but in the same way as I did in the PR for client ones.

Changes in #600 makes [this test](https://github.com/hyperium/h2/blob/6a75f232330374d5f329aaae91afc2dee7ed2b1f/tests/h2-tests/tests/stream_states.rs#L1042C10-L1042C37) failing without this PR. 

All this logic looks too coupled & fragile, however current code is 100% erroneous. As I mentioned this probably needs a deeper refactoring (even [this line](https://github.com/hyperium/h2/blob/6a75f232330374d5f329aaae91afc2dee7ed2b1f/src/proto/streams/state.rs#L434) is a call to action).
I'd be happy to help, but I seem to be a little stuck 😞 At least I don't have the bandwidth to go as deep as I need to.

Btw, there is a little mess in [stream_states](https://github.com/hyperium/h2/blob/master/tests/h2-tests/tests/stream_states.rs) test-module, some names doesn't looks like they fit the logic and 3 tests contains the same description:
```rs
    // Data is buffered in `FramedWrite` and the stream is reset locally before
    // the data is fully flushed. Given that resetting a stream requires
    // clearing all associated state for that stream, this test ensures that the
    // buffered up frame is correctly handled
```

DDtKey: @seanmonstar hey, sorry for delay.
Seems you was actually right - it should be an error. `h2spec` tests highlighted the wrong behavior and I update to correct errors.

Please, can you take a look again?

seanmonstar (approved):

PR #692, opened 2023-06-13
Fix for a fuzzer-discovered integer underflow of the flow control window size

f0rki:
Fuzzing discovered a integer underflow in the flow control handling that can be triggered by a certain sequence of frames. In release builds this would lead to wrap-around of the negative window. This seems incorrect to me. I did not find anything definitive in the http2 spec.

Removed the SubAssign, etc. syntactic sugar functions and switched to return Result on over/underflow

Whenever possible, switched to returning a library GoAway protocol error. Otherwise we check for over/underflow only with `debug_assert!`, assuming that those code paths do not over/underflow. In this case I left `TODO:` comments.

### DIFF
--- a/src/proto/connection.rs
+++ b/src/proto/connection.rs
@@ -145,7 +145,9 @@ where
 
     /// connection flow control
     pub(crate) fn set_target_window_size(&mut self, size: WindowSize) {
-        self.inner.streams.set_target_connection_window_size(size);
+        let _res = self.inner.streams.set_target_connection_window_size(size);
+        // TODO: proper error handling
+        debug_assert!(_res.is_ok());
     }
 
     /// Send a new SETTINGS frame with an updated initial window size.
--- a/src/proto/streams/flow_control.rs
+++ b/src/proto/streams/flow_control.rs
@@ -182,7 +182,7 @@ impl FlowControl {
 
             // Update values
             self.window_size.decrease_by(sz)?;
-            self.available.increase_by(sz)?;
+            self.available.decrease_by(sz)?;
         }
         Ok(())
     }
@@ -222,12 +222,9 @@ impl Window {
     }
 
     pub fn increase_by(&mut self, other: WindowSize) -> Result<(), Reason> {
-        if let Some(v) = self.0.checked_add(other as i32) {
-            self.0 = v;
-            Ok(())
-        } else {
-            Err(Reason::FLOW_CONTROL_ERROR)
-        }
+        let other = self.add(other)?;
+        self.0 = other.0;
+        Ok(())
     }
 
     pub fn add(&self, other: WindowSize) -> Result<Self, Reason> {
--- a/src/proto/streams/flow_control.rs
+++ b/src/proto/streams/flow_control.rs
@@ -75,12 +75,12 @@ impl FlowControl {
         self.window_size > self.available
     }
 
-    pub fn claim_capacity(&mut self, capacity: WindowSize) {
-        self.available -= capacity;
+    pub fn claim_capacity(&mut self, capacity: WindowSize) -> Result<(), Reason> {
+        self.available.decrease_by(capacity)
     }
 
-    pub fn assign_capacity(&mut self, capacity: WindowSize) {
-        self.available += capacity;
+    pub fn assign_capacity(&mut self, capacity: WindowSize) -> Result<(), Reason> {
+        self.available.increase_by(capacity)
     }
 
     /// If a WINDOW_UPDATE frame should be sent, returns a positive number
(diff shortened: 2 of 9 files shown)


src/proto/connection.rs
```diff
@@ -145,7 +145,9 @@ where
 
     /// connection flow control
     pub(crate) fn set_target_window_size(&mut self, size: WindowSize) {
-        self.inner.streams.set_target_connection_window_size(size);
+        let _res = self.inner.streams.set_target_connection_window_size(size);
+        // TODO: proper error handling
+        debug_assert!(_res.is_ok());
```
82marbag, line 150: There are some debug left around here and in other files. Are these on purpose?

src/proto/streams/flow_control.rs
```diff
@@ -176,12 +178,15 @@ impl FlowControl {
         // If send size is zero it's meaningless to update flow control window
         if sz > 0 {
             // Ensure that the argument is correct
-            assert!(self.window_size >= sz as usize);
+            assert!(self.window_size.0 >= sz as i32);
 
             // Update values
-            self.window_size -= sz;
-            self.available -= sz;
+            // debug_assert!(self.window_size.0 >= sz as i32);
```
82marbag, line 184: On purpose?

f0rki: nope this one is not.

82marbag, line 184:
```suggestion

```

src/proto/connection.rs
f0rki: Yes. Whenever it was obvious on how to handle a potential overflow, I did. However, many of the call sites do not have a `Result` return type, so it is not easy to propagate the Error. I assume that those functions should not be able not fail. I did not want to incur additional overhead in release builds, so I stuck with `debug_assert!` and left the `TODO` comment. For those TODOs someone with more experience with the code should check whether they are fine to be left as `debug_assert!`, whether they should be checked with `assert!` or whether error propagation is needed.

src/proto/streams/recv.rs
```diff
@@ -90,7 +90,7 @@ impl Recv {
         // settings
         flow.inc_window(DEFAULT_INITIAL_WINDOW_SIZE)
             .expect("invalid initial remote window size");
-        flow.assign_capacity(DEFAULT_INITIAL_WINDOW_SIZE);
+        flow.assign_capacity(DEFAULT_INITIAL_WINDOW_SIZE).unwrap();
```
82marbag, line 93: Is this ok to unwrap?

src/proto/streams/flow_control.rs
```diff
@@ -176,12 +178,13 @@ impl FlowControl {
...
         if sz > 0 {
             // Ensure that the argument is correct
-            assert!(self.window_size >= sz as usize);
+            assert!(self.window_size.0 >= sz as i32);
 
             // Update values
-            self.window_size -= sz;
-            self.available -= sz;
+            self.window_size.decrease_by(sz)?;
+            self.available.increase_by(sz)?;
```
82marbag: Now I understand why a test was failing :)

82marbag:
@f0rki is it ready for review?

src/proto/streams/recv.rs
f0rki: I would suggest doing a `debug_assert!` on the result?

src/proto/streams/flow_control.rs
f0rki: oh no. what a stupid mistake...

f0rki:
> @f0rki is it ready for review?

looks good to me @82marbag @seanmonstar

src/proto/connection.rs
f0rki: FYI: I did more fuzzing runs with this patch and none of the `debug_assert!` are triggered, so that gives some assurance that the they are indeed not reachable. So I think we could remove the todo comments, but I would leave the `debug_assert!` as is.

seanmonstar (approved):
Excellent work, thank you!

nox:
In an effort to understand a bit more what's going on in https://github.com/hyperium/h2/issues/607, I've been looking at the test from this PR to see if the root cause could be the same here.

I've managed to simplify the test there a great deal, which led me to this (I'll make a PR simplifying it, separately):

```rust
#[tokio::test]
async fn window_size_decremented_past_zero() {
    h2_support::trace_init!();
    let (io, mut client) = mock::new();

    let client = async move {
        let settings = client.assert_server_handshake().await;
        assert_default_settings!(settings);

        // HEADERS with invalid hpack stuff.
        client
            .send_bytes(&[
                0, 0, 23, 1, 1, 0, 0, 0, 1, 131, 1, 1, 1, 70, 1, 1, 1, 1, 65, 1, 1, 65, 1, 1,
                65, 1, 1, 1, 1, 1, 1, 190,
            ])
            .await;

        client.send_frame(frames::settings().initial_window_size(666413898)).await;
        client.send_frame(frames::settings().initial_window_size(3809661)).await;
        client.send_frame(frames::settings().initial_window_size(666413898)).await;
        client.send_frame(frames::settings().initial_window_size(3809661)).await;
        client.send_frame(frames::settings().initial_window_size(1467177332)).await;
        client.send_frame(frames::settings().initial_window_size(3844989)).await;
    };

    let srv = async move {
        let builder = server::Builder::new();

        // just keep it open
        let res = poll_fn(move |cx| srv.poll_closed(cx)).await;
        tracing::debug!("{:?}", res);
    };

    join(client, srv).await;
}
```

The test fails if we put back an assertion in `FlowControl::dec_send_window`, as expected. If I remove any of the frames, the test passes.

So I looked into what happens when the initial window size setting is incremented or decremented, and I've found something suspicious.

https://github.com/hyperium/h2/blob/3bce93e9b0dad742ffbf62d25f9607ef7651a70c/src/proto/streams/send.rs#L469-L480

When a stream is already closed and the initial window size setting is incremented, nothing happens to the stream because it is already closed.

https://github.com/hyperium/h2/blob/3bce93e9b0dad742ffbf62d25f9607ef7651a70c/src/proto/streams/prioritize.rs#L306-L309

But when the initial window size setting is decremented, the stream window size is also decremented, so there is an imbalance there.

To double check my finding, I managed to reduce the number of frames to cause the assertion failure with:

```rust
        client.send_frame(frames::settings().initial_window_size(1329018135)).await;
        client.send_frame(frames::settings().initial_window_size(3809661)).await;
        client.send_frame(frames::settings().initial_window_size(1467177332)).await;
        client.send_frame(frames::settings().initial_window_size(3844989)).await;
```

Note that `1329018135 = 666413898 * 2 - 3809661`, so after the second frame, the stream's window size is the same negative value as after the 4th one in the complete test above.

PR #695, opened 2023-06-15
Fix for invalid header panic corrected

f0rki:
Corrected fix in #690 - the root cause was that informational header frames are ignored, even if received by the server. However, the stream was still scheduled in `pending_accept`. This then triggers the `take_request`, while nothing is put in the `pending_recv` queue, which then causes the panic.

### DIFF
--- a/src/proto/streams/recv.rs
+++ b/src/proto/streams/recv.rs
@@ -263,7 +263,9 @@ impl Recv {
 
         match stream.pending_recv.pop_front(&mut self.buffer) {
             Some(Event::Headers(Server(request))) => request,
-            _ => unreachable!("This panic should not be reached. Cut an issue at https://github.com/hyperium/h2"),
+            _ => unreachable!(
+                "This panic should not be reached. Cut an issue at https://github.com/hyperium/h2"
+            ),
         }
     }
 
--- a/src/proto/streams/recv.rs
+++ b/src/proto/streams/recv.rs
@@ -248,6 +248,8 @@ impl Recv {
             // Only servers can receive a headers frame that initiates the stream.
             // This is verified in `Streams` before calling this function.
             if counts.peer().is_server() {
+                // Correctness: never push a stream to `pending_accept` without having the
+                // corresponding headers frame pushed to `stream.pending_recv`.
                 self.pending_accept.push(stream);
             }
         }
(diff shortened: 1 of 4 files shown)


src/proto/streams/recv.rs
```diff
@@ -239,27 +244,26 @@ impl Recv {
...
         use super::peer::PollMessage::*;
 
         match stream.pending_recv.pop_front(&mut self.buffer) {
-            Some(Event::Headers(Server(request))) => Ok(request),
-            _ => {
-                proto_err!(stream: "received invalid request; stream={:?}", stream.id);
-                Err(Error::library_reset(stream.id, Reason::PROTOCOL_ERROR))
-            }
+            Some(Event::Headers(Server(request))) => request,
+            _ => panic!(),
```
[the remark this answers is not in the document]
f0rki: this panic should not be reachable anymore.

82marbag, line 266:
```suggestion
            _ => panic!("This panic should not be reached. Cut an issue at https://github.com/hyperium/h2"),
```

f0rki: I think the `unreachable!` macro is better suited, no?

82marbag: Edited. It's for clarity, as it is implemented as `panic!`

src/proto/streams/recv.rs
```diff
@@ -239,27 +244,33 @@ impl Recv {
...
 
         match stream.pending_recv.pop_front(&mut self.buffer) {
-            Some(Event::Headers(Server(request))) => Ok(request),
-            _ => {
-                proto_err!(stream: "received invalid request; stream={:?}", stream.id);
-                Err(Error::library_reset(stream.id, Reason::PROTOCOL_ERROR))
-            }
+            Some(Event::Headers(Server(request))) => request,
+            _ => unreachable!(
+                "This panic should not be reached. Cut an issue at https://github.com/hyperium/h2"
```
seanmonstar, line 272: `unreachable!` will already say it shouldn't be reached. It's nice to instead include in the string why, like `unreachable!("server stream queue should start with Headers")` (or whatever message makes the most sense.)

seanmonstar (approved):

PR #696, opened 2023-06-20
DRAFT: disabled two assertions while fuzzing

f0rki:


src/proto/connection.rs
```diff
@@ -448,6 +448,7 @@ where
             // This is handled by resetting the frame then trying to read
             // another frame.
             Err(Error::Reset(id, reason, initiator)) => {
+                #[cfg(not(fuzzing))]
                 debug_assert_eq!(initiator, Initiator::Library);
```
seanmonstar, line 452: @nox do you recall why this is asserted? If it's `Remote`, should something else be done? This is getting trigger in fuzzing.

nox: It is done because from my understanding of the code back then this path should never be taken with `Initiator::Remote`, can you repro how it does happen?

PR #698, opened 2023-06-26
Update indexmap to version 2

djc:


djc:
Hmm, I guess you might not be eager to bump the MSRV to 1.64 yet.

jplatte:
1.64 would be in line with `hyper`'s [MSRV policy](https://github.com/hyperium/hyper/blob/0.14.x/docs/MSRV.md) of the MSRV always being at least 6 months old (1.64.0 was released on Sept. 22, 2022).

I would appreciate this being merged since it removes duplicate versions of indexmap & hashbrown from some dependency trees.

djc:
@seanmonstar any opinions?

khuey:
Likewise I'd also benefit from this. h2 is the last crate in my dependency tree using indexmap 1.x and hashbrown 0.12.x

seanmonstar:
I recently bumped hyper to 1.63 (mostly because Tokio required it, not because hyper needs anything). While it would be within policy to go one more for 1.64, there's one thing that makes me hesitate: Debian stable is currently using 1.63. 🤔

cuviper:
I don't think hashbrown really *needs* 1.64 -- only recent versions of its optional `allocator-api2` dep need that, but its min-dep 0.2.9 does work with 1.63. And I can speak for `indexmap` that I only set 1.64 to keep up with `hashbrown`. These could possibly be relaxed!

djc:
AIUI Debian stable will be on 1.63 for 2 years, but I don't think it will be tenable for this crate to stick to that for that kind of period. So what's the value in sticking to what Debian stable is today?

cuviper:
At least for `indexmap`, it would be nice to let Debian users reach that `v2` semver upgrade, even though `2.1` etc. may later raise the MSRV beyond 1.63. In this case, there wouldn't be any real sacrifice to offer that carrot.

ds-cbo:
`indexmap` just released version [v2.0.1](https://github.com/bluss/indexmap/pull/276) which supports MSRV 1.63.0. Could this PR get another CI-run?

seanmonstar:
Seems like to time period to restart CI builds has expired, and I can't make them run again without a push.

jplatte:
Close & reopen also triggers CI.

ds-cbo:
Ah, too bad, we'll have to wait for https://github.com/rust-lang/hashbrown/pull/457 to be released as hashbrown 0.14.1/0.15 before this will work. They seem to be on a quarterly release schedule with the next release guessed to be around October, so we'll need a bit more patience. Thanks for the retry anyway 😊

cuviper:
Rust 1.63 should be sorted out with `indexmap v2.02` and `hashbrown v0.14.1`.

cuviper:
To be clear, those 1.63-compatible `hashbrown` and `indexmap` versions are already available.

ds-cbo:
@seanmonstar Would you mind closing and re-opening the issue again to re-trigger CI? 😙 Should work this time

djc:
Pushed a rebased version including some additional dev-dependency bumps.

seanmonstar (approved):

Issue #701, opened 2023-07-20
`h2spec 2.6.0` tests are failling against `h2`


DDtKey:
I tried to run the latest version of `h2spec` against master branch in `h2` and it's failing with the following error:

```
Hypertext Transfer Protocol Version 2 (HTTP/2)
  3. Starting HTTP/2
    3.5. HTTP/2 Connection Preface
      using source address 127.0.0.1:62276
      × 2: Sends invalid connection preface
        -> The endpoint MUST terminate the TCP connection.
           Expected: GOAWAY Frame (Error Code: PROTOCOL_ERROR)
                     Connection closed
             Actual: Error: read tcp 127.0.0.1:62276->127.0.0.1:5928: read: connection reset by peer

```

See #702

hawkw:
Huh, looks like we're not properly GOAWAYing the conn in this case and just immediately closing it?

DDtKey:
Seems so, and moreover, I missed one more case:

```
Generic tests for HTTP/2 server
  3. Frame Definitions
    3.8. GOAWAY
      using source address 127.0.0.1:58401
      × 1: Sends a GOAWAY frame
        -> The endpoint MUST accept GOAWAY frame.
           Expected: Connection closed
                     PING Frame (length:8, flags:0x01, stream_id:0, opaque_data:h2spec)
             Actual: Error: read tcp 127.0.0.1:58401->127.0.0.1:5928: read: connection reset by peer

```

But this seems may be closed with #687 - it's not reproducible to me with this fix.

But the first error is still relevant

DDtKey:
@hawkw 
Well, now I don't think so, it seems related to https://github.com/summerwind/h2spec/issues/81

And the fix of `h2spec` was here: https://github.com/summerwind/h2spec/pull/119 (see the last comment)

So I guess it's allowed behavior.  But there is no newer release of `h2spec`, only build from `main` branch probably.

hawkw:
Ah, yup, it looks like our current behavior is permitted but `h2spec` incorrectly *requires* us to send a GOAWAY, based on the issue you linked.

For now, I think the best thing to do is to close this issue and wait for a new version to be released before updating `h2spec`.

PR #702, opened 2023-07-20
test: update `h2spec` version

DDtKey:
But it fails against the current `h2` version, see #701


DDtKey:
I'll try to fix

DDtKey:
Postponed until the new version of `h2spec` with fix

PR #703, opened 2023-07-21
test: early server response with data

DDtKey:
- fix of the test's name after changing it in #634
- additional test that server also sends data frames correctly

This is normal scenario and I decided to extend test a bit.

### DIFF
--- a/tests/h2-tests/tests/server.rs
+++ b/tests/h2-tests/tests/server.rs
@@ -553,7 +553,7 @@ async fn recv_connection_header() {
 }
 
 #[tokio::test]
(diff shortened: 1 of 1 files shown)


DDtKey:
And also this is preliminary change to handle `NO_ERROR` correctly on client side in `hyper`, because currently such case causes:
```rs
hyper::Error(Body, Error { kind: Reset(StreamId(3), NO_ERROR, Remote) }
```
While request was handled correctly (for example, reaching max request body size) and the response was sent.

seanmonstar (approved):
Nice!

PR #705, opened 2023-08-15
Feat: Add a feature flag to ignore connection headers

eaufavor:
According to the RFC, when encountering connection headers, H2 should treat them as protocol errors. However, in reality there are servers just setting these headers but only for informational purpose.

This feature allow the receiving end to just ignore these headers without bailing the entire stream.

I'm just putting this out here to see whether it is possible for things like this to be accepted.

### DIFF
--- a/src/frame/headers.rs
+++ b/src/frame/headers.rs
@@ -871,16 +871,25 @@ impl HeaderBlock {
             match header {
                 Field { name, value } => {
                     // Connection level header fields are not supported and must
-                    // result in a protocol error.
+                    // result in a protocol error. However in reality there are
+                    // servers setting these headers. Some of these headers are
+                    // purely informational and can be simply ignored without
+                    // affecting the integrity of the stream.
 
-                    if name == header::CONNECTION
-                        || name == header::TRANSFER_ENCODING
+                    if name == header::TRANSFER_ENCODING {
+                        tracing::trace!("load_hpack; connection level header");
+                        malformed = true;
+                    } else if name == header::CONNECTION
                         || name == header::UPGRADE
                         || name == "keep-alive"
                         || name == "proxy-connection"
                     {
-                        tracing::trace!("load_hpack; connection level header");
-                        malformed = true;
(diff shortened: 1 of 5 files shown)


seanmonstar:
Are there any other widely used implementations that support this? I initially think the pressure should instead be pushed back onto those servers to change, since weakening the implementations is how HTTP/1.1 got to have so many edge cases.

eaufavor:
Good question.
I've test a few user agents:
The server: nginx conf 
```
location /test_h2_header/ {
   add_header connection "keep-alive";
   return 200 "hello"
}
```
https://origin.eaufavor.net/test_h2_header/

- Chrome via h2: accepts the connection header (the header shows in network debug tab), no error.
- Firefox via h2: ignores the header (not in the network debug tab), no error. Interestingly it also shows that "Connection: keep-alive" is sent by Firefox as a request header (if we trust it own debug tool).
- Curl: stream error.

I have not yet tested other connection headers.

nox:
Why is "CI / Check MSRV (pull_request)" failing?

eaufavor:
This https://github.com/hyperium/h2/pull/708?

nox:
Ah, could you rebase then?

src/frame/headers.rs
```diff
@@ -871,16 +871,25 @@ impl HeaderBlock {
...
+                        tracing::trace!("load_hpack; connection level header");
+                        malformed = true;
+                    } else if name == header::CONNECTION
                         || name == header::UPGRADE
                         || name == "keep-alive"
                         || name == "proxy-connection"
                     {
-                        tracing::trace!("load_hpack; connection level header");
-                        malformed = true;
+                        if cfg!(feature = "ignore_connection_header") {
```
nox, line 887: I don't think this should be a cargo feature, instead it should be a runtime option, so that each part of a large system can decide whether or not they want to enable it.

eaufavor: Sounds good. I can try to update (no promises when). Just wanted to make sure there is a chance for features like this to be accepted.

PR #707, opened 2023-08-21
i704 max concurrent streams mid connection

### LINKED ISSUE #704: Settings frames are not applied correctly

seanmonstar:
I'm just opening a separate pull request that I can push commits to, more details in the linked items.

Fixes #704 
Builds on #706


tests/h2-tests/tests/server.rs
```diff
@@ -296,10 +296,10 @@ async fn push_request_against_concurrency() {
             .await;
         client.recv_frame(frames::data(2, &b""[..]).eos()).await;
         client
-            .recv_frame(frames::headers(1).response(200).eos())
+            .recv_frame(frames::headers(4).response(200).eos())
             .await;
         client
-            .recv_frame(frames::headers(4).response(200).eos())
+            .recv_frame(frames::headers(1).response(200).eos())
```
seanmonstar, line 302: @jwilm this was an interesting change in behavior... I'm guessing this is because of the `push_front` vs `push_back` that you added?

jwilm: That sounds a likely cause. There's another option with how the pending_open/pending_send queues are managed where we could prefer to pop data frames before popping new streams. I think that would actually be a bit closer to the old behavior, but it would make the `poll_complete` implementation more complex.

Issue #711, opened 2023-08-31
Sending data frame results in transmission of more tcp segments than needed


xiaoyawei:
## What is the issue?

I was using `tonic` (which depends on `h2` for http2) for a bidi-streaming gRPC service. With some packet sniffing, I noticed that when one side is sending data of around 1KB to the other side through the gRPC stream (which is based on an underlying http2 stream), the data is always transmitted as 2 TCP segments: the 1st one carries the http2 data frame's head , which is of 9 bytes, while the 2nd one carries the real payload.

Since `tonic` (and many other libraries) disables nagle algorithm by default, the above issue happens in a relatively wide range. For us, our application keeps transmitting data around 1KB, and this issue leads to a lot of tcp segments being transmitted, which is a big overhead and affects our applications' performance.

## What is the root cause?

Yes, `h2` is agnostic to the underlying transport protocol, but I assume that most of use cases are built on top of TCP; the problematic code is here

https://github.com/hyperium/h2/blob/da38b1c49c39ccf7e2e0dca51ebf8505d6905597/src/codec/framed_write.rs#L243-L251

When sending a data frame with payload size over `CHAIN_THRESHOLD` (which is 256 now), the frame head (9 bytes) is written to buffer, while the entire payload is left in `self.next`; then when data is being transmitted over the transport,

https://github.com/hyperium/h2/blob/da38b1c49c39ccf7e2e0dca51ebf8505d6905597/src/codec/framed_write.rs#L125-L135

the buffer is chained with the next token, which is finally written into the transport via 

https://github.com/hyperium/h2/blob/da38b1c49c39ccf7e2e0dca51ebf8505d6905597/src/codec/framed_write.rs#L186-L186

here `buf` is of type `bytes::buf::Chain` in `bytes` crate, whose `chunk()` method is implemented [as follows](https://github.com/tokio-rs/bytes/-/blob/64c4fa286771ad9e522ffbefc576bcf7b76933d0/src/buf/chain.rs#L141-147)

```rust
    fn chunk(&self) -> &[u8] {
        if self.a.has_remaining() {
            self.a.chunk()
        } else {
            self.b.chunk()
        }
    }
```

Therefore, the buffer, which only contains the data frame's head, is first written to the transport (which is tcp in most use cases), and the payload is then written separately. In the scenarios where nagel algorithm is disabled, it'll lead to the that data frame's head and payload are transmitted in separate tcp segments, which causes some overhead.

seanmonstar:
[...]

PR #712, opened 2023-09-04
Improve throughput when vectored IO is not enabled

xiaoyawei:
As discussed in https://github.com/hyperium/h2/issues/711, the current implementation of sending data is suboptimal when vectored I/O is not enabled: data frame's head is likely to be sent in a separate TCP segment, whose payload is of only 9 bytes.

This PR adds some specialized implementaton for non-vectored I/O case. In short, it sets a larget chain threhold, and also makes sure a data frame's head is sent along with the beginning part of the real data payload.

All existing unit tests passed. Also I take a look at the e2e https://github.com/hyperium/hyper/blob/0.14.x/benches/end_to_end.rs but realize that all the benchmarks there are for the case of vectored I/O if the OS supports vectored I/O. There isn't a specific case for non-vectored I/O so I am not sure how to proceed with benchmark for performance evaluations.  @seanmonstar Would appreciate if there's some advice on how to do benchmarks for this change.

This PR also includes a trivial bug fix, and addressing a straightforward TODO comment


seanmonstar:
> There isn't a specific case for non-vectored I/O so I am not sure how to proceed with benchmark for performance evaluations.

For comparison purposes, you could write a `struct NonVecIo<T>(T)`, and implement `AsyncRead` and `AsyncWrite` for it, forwarding each method on, except for `is_write_vectored()`. You could do that in the `end_to_end.rs` file, and compare before and after. I don't think we need commit that change, though...

xiaoyawei:
@seanmonstar I forked `hyper` and put together a commit https://github.com/hyperium/hyper/commit/6f3c37939236899a23e8d4f694d00cc16b979f2c to benchmark vectored io. In my test setup, parallelism is `10`, both request and response payload is of 500 bytes. I run the benchmark with h2 `0.3.9` and this fix 3 times, and the result is as follows

### `0.3.9`

* 242576 ns/iter (+/- 32535)
* 257091 ns/iter (+/- 82970)
* 258892 ns/iter (+/- 216218)

### `w/ this fix`

* 184308 ns/iter (+/- 57584)
* 181720 ns/iter (+/- 51599)
* 179991 ns/iter (+/- 38387)

Looks like the perf improvement is pretty solid

seanmonstar:
Nice! Thanks for putting that together.

@nox or @Noah-Kennedy, does this seem good in your cases too?

Noah-Kennedy:
I'll take a look today. Not sure if @nox has opinions or not.

.gitignore
```diff
@@ -8,3 +8,5 @@ h2spec
 # Files generated by honggfuzz
 hfuzz_target
 hfuzz_workspace
+
+**/.idea
```
Noah-Kennedy, line 12: @seanmonstar do you care if there are IDE-specific ignore rules?

seanmonstar: Hm, so far we haven't, since this can be one on an individual's machine, by putting them in `~/.gitignore`.

xiaoyawei: I can move this out. Will address it along with other comments once all stackholders' opinions are received

xiaoyawei:
@seanmonstar @Noah-Kennedy 

I removed the gitignore rules; let me know if there are other feedback, thanks ;)

seanmonstar (approved):

seanmonstar:
Excellent work, thank you for the benchmark results, they certainly help in feeling confident with the change.

PR #735, opened 2024-01-11
Make the hpack a public module

ymesika:
This PR is making the hpack module with its encoder/decoder as public for other projects which have the need for this low-level capablities specifically.

### DIFF
--- a/src/hpack/header.rs
+++ b/src/hpack/header.rs
@@ -117,6 +117,16 @@ impl Header {
         }
     }
 
+    pub fn is_empty(&self) -> bool {
+        match *self {
+            Header::Field {
+                ref name,
+                ref value,
+            } => len(name, value) == 0,
+            _ => false,
+        }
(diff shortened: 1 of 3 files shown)


seanmonstar:
Thanks for the PR! While I'm not against exposing hpack, theoretically, I think it's something that should be done with care. Once the types are made public, we can't change them internally if we want to optimize.

What's the use case for needing hpack? Is it to implement a different HTTP/2 library, while reusing our hpack module?

ymesika:
@seanmonstar thanks for reviewing.
I understand your concern exposing the interfaces and I found a crate which covers our needs at the moment.
Unfortunately all other hpack crates seem to be unmaintained (and maybe it's because there is nothing to maintain there) but have you thought making the internal hpack impl in this crate as a separate one?

Our needs are more lower level adding headers sniffing in a TCP proxy.

PR #737, opened 2024-01-17
streams: limit error resets for misbehaving connections

Noah-Kennedy:
This change causes GOAWAYs to be issued to misbehaving connections which for one reason or another cause us to emit lots of error resets.

Error resets are not generally expected from valid implementations anyways.

The threshold after which we issue GOAWAYs is tunable, and will default to 1024.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -336,6 +336,12 @@ pub struct Builder {
     /// The stream ID of the first (lowest) stream. Subsequent streams will use
     /// monotonically increasing stream IDs.
     stream_id: StreamId,
+
+    /// Maximum number of locally reset streams due to protocol error across
+    /// the lifetime of the connection.
+    ///
+    /// When this gets exceeded, we issue GOAWAYs.
+    local_max_error_reset_streams: Option<usize>,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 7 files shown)


src/client.rs
```diff
@@ -973,6 +980,23 @@ impl Builder {
...
+    /// Too many of these often indicate a malicious client, and there are attacks which can abuse this to DOS servers.
+    /// This limit protects against these DOS attacks by limiting the amount of resets we can be forced to generate.
+    ///
+    /// When the number of local resets exceeds this threshold, the client will close the connection.
+    ///
+    /// If you really want to disable this, supply [`Option::None`] here.
+    /// Disabling this is not recommended and may expose you to DOS attacks.
+    ///
+    /// The default value is currently 1024, but could change.
+    pub fn max_local_error_reset_streams(&mut self, max: Option<usize>) -> &mut Self {
```
seanmonstar, line 995: Should this be an option? Just to argue the other side, it can also essentially be disabled by setting the max to be the `1 << 31`, right?

I think either choice could be right, let's just discuss briefly right here.

Noah-Kennedy: I figured it would be cleaner and to just make it optional rather than to make people try and remember the highest usize.

seanmonstar (approved):

PR #742, opened 2024-01-18
streams: add a second check in front of reset flood mitigations

Noah-Kennedy:
Some clients occasionally send garbage, so let's also only mitigate if they actually start inflating the queue to try and attack us.


Noah-Kennedy:
marking as draft while i figure out if this is even useful

PR #743, opened 2024-01-19
feat: not returning UnexpectedEof when client drop without close_notify

dswij:
closes https://github.com/hyperium/hyper/issues/3427

This PR makes it so that if we encountered `UnexpectedEof` as a server, and when there is nothing else to send (i.e. we sent what the client wanted), `Connection::poll` will not return an `Err`.

Note that this does not change the client behaviour when the server closed the connection abruptly.

---

Something to consider is whether we should handle this in hyper instead. But I don't see a good reason why we shouldn't put it in `h2`

### DIFF
--- a/src/proto/connection.rs
+++ b/src/proto/connection.rs
@@ -461,12 +461,26 @@ where
             // active streams must be reset.
             //
             // TODO: Are I/O errors recoverable?
-            Err(Error::Io(e, inner)) => {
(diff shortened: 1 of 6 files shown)


Noah-Kennedy (approved):

seanmonstar (approved):

cablehead:
🙌🏻 💚

PR #746, opened 2024-02-01
fix: stream flow control insufficient size before ack

### LINKED ISSUE #630: FlowControl problem when the initial_stream_window_size is smller than 65535
If I start an HTTP2 service as follows,  the local_init_window_sz will be  is set to 32767 when a new connection is established. Before client receive the server initial settings frame of connection preface, it sends the data frame using the default window size 65535. In this case, the data received by the service may exceed the tream window size.
``` rust
let server_fut = Builder::new().initial_window_size(32767);
```
The default value should probably be used when the connection is established and updated to the custom value after the ACK settings frame is received. Another way is to limit the custom size to no less than 65535.

dswij:
closes #630 

This PR makes the Recv stream use the default window size before receiving ACK from the other party.

This matters as a server when we want the stream `initial_window_size` <= default settings. In some cases, the client will eagerly send some data frames before ACK-ing settings, thus resulting in a flow control error if the size of data frames is greater than the set `initial_window_size`.

### DIFF
--- a/src/proto/connection.rs
+++ b/src/proto/connection.rs
@@ -106,10 +106,6 @@ where
     pub fn new(codec: Codec<T, Prioritized<B>>, config: Config) -> Connection<T, P, B> {
         fn streams_config(config: &Config) -> streams::Config {
             streams::Config {
-                local_init_window_sz: config
-                    .settings
-                    .initial_window_size()
-                    .unwrap_or(DEFAULT_INITIAL_WINDOW_SIZE),
                 initial_max_send_streams: config.initial_max_send_streams,
                 local_max_buffer_size: config.max_send_buffer_size,
                 local_next_stream_id: config.next_stream_id,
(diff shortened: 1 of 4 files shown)


src/proto/connection.rs
```diff
@@ -106,10 +106,6 @@ where
     pub fn new(codec: Codec<T, Prioritized<B>>, config: Config) -> Connection<T, P, B> {
         fn streams_config(config: &Config) -> streams::Config {
             streams::Config {
-                local_init_window_sz: config
-                    .settings
-                    .initial_window_size()
-                    .unwrap_or(DEFAULT_INITIAL_WINDOW_SIZE),
```
seanmonstar, line 112: Oh good catch. This gets applied after the peer sends back an ACK, right?

dswij: Yes, it will: 

https://github.com/hyperium/h2/blob/dd67adc19268a9780b09502ecc660be8363f008f/src/proto/streams/recv.rs#L489

We're still negotiating, but just not eagerly opening streams with non-default settings.

seanmonstar (approved):

PR #750, opened 2024-02-21
perf: optimize header list size calculations

Noah-Kennedy:
This speeds up loading blocks in cases where we have many headers already.

### DIFF
--- a/src/frame/headers.rs
+++ b/src/frame/headers.rs
@@ -12,6 +12,7 @@ use std::fmt;
 use std::io::Cursor;
 
 type EncodeBuf<'a> = bytes::buf::Limit<&'a mut BytesMut>;
+
 /// Header frame
 ///
 /// This could be either a request or a response.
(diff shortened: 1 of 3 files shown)


Noah-Kennedy:
Accidentally targeted it at 0.3, though I think this is enough of a perf improvement to justify a backport.

We can just forwardport this to 0.4 after I guess?

src/frame/headers.rs
```diff
@@ -892,6 +901,8 @@ impl HeaderBlock {
 
                         headers_size += decoded_header_size(name.as_str().len(), value.len());
                         if headers_size < max_header_list_size {
+                            self.field_size +=
+                                decoded_header_size(name.as_str().len(), value.len());
```
rushilmehra, line 905: Should we check that `decoded_header_size` won't exceed `max_header_list_sized`?

Noah-Kennedy: that is covered under other checks elsewhere

rushilmehra: Nvm, that's exactly what this does

seanmonstar (approved):

rushilmehra (approved):

src/frame/headers.rs
Noah-Kennedy: Yeah I don't know why I said that lol, it's literally checked right here.

PR #751, opened 2024-02-22
set rust `edition` to `2021`

smndtrl:
msrv is already newer than `2021`. However `2021` also enables the v2 resolver


seanmonstar:
Thanks for the PR! Seems a couple imports have become redundant, want to fix them up?

smndtrl:
For reference https://github.com/rust-lang/rust/issues/121315

seanmonstar (approved):

PR #765, opened 2024-04-13
Switch to trace logging for frame recieved messages

orf:
Enabling `DEBUG` tracing/logging when using `h2` results in a very, very large volume of messages being logged.

These feel like `trace`-level messages, not debug messages. This PR switches them to `trace` instead.

Screenshot 2024-04-13 at 20 15 12

### DIFF
--- a/src/codec/framed_read.rs
+++ b/src/codec/framed_read.rs
@@ -402,7 +402,7 @@ where
                 partial,
                 bytes,
             )? {
-                tracing::debug!(?frame, "received");
+                tracing::trace!(?frame, "received");
                 return Poll::Ready(Some(Ok(frame)));
             }
         }


seanmonstar:
I get that it's a lot of frames, and they might not be that interesting, but then again, I do think the frame events are more important than all the other events emitted at trace level. Especially if you're trying to debug what the communication to the peer was.

orf:
Perhaps there’s a middle ground here? “Uninteresting” frames, like data frames, could be logged with trace, but less-common ones could be logged as debug messages?

Personally, and with someone with little http2 experience, I’m not sure how these messages would help me debug communication issues. It only contains the stream ID and not other relevant information like the size of the frame, which I’d assume would be useful to debug issues?

seanmonstar:
The length would be a valuable thing to add.

PR #776, opened 2024-05-08
chore: make tracing optional

### LINKED ISSUE #713: Make tracing an optional feature
Make `tracing` available as a Cargo feature, probably default off.

tottoto:
Resolves #713.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -51,6 +51,9 @@ jobs:
       - name: Check with unstable flag
         run: cargo check --features unstable
 
+      - name: Check with tracing feature
+        run: cargo check --features tracing
+
       - name: Run lib tests and doc tests
         run: cargo test
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,9 @@ rust-version = "1.63"
 # Enables `futures::Stream` implementations for various types.
 stream = []
 
+# Enables tracing.
+tracing = ["dep:tracing"]
+
 # Enables **unstable** APIs. Any API exposed by this feature has no backwards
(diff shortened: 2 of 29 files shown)


gituser-rs:
is there any chance of approving this?

seanmonstar:
It will be, it's just a breaking change.

Incisakura:
Waiting for this PR to be merged.

If anyone is annoyed by h2 (or any other crate) emitting tracing logs but actually useless to them, a temporary workaround is to set the tracing feature "max_level_off," which would prevent the generation of the event log code, potentially improving runtime performance and reducing the binary size.

``` toml
[dependencies]
# your dependencies here

tracing = { version = "*", default-features = false, features = ["max_level_off", "release_max_level_off"] }
```

papa-zakari:
> @seanmonstar is there anything that external contributors can do to help move this forward?
> 
> Does it makes sense to make it an on-by-default feature to avoid the backwards compatibility issues?

While I can't speak on behalf of the org, I think that this is a great idea @gaynor-anthropic.

Hyper could use `default-features = false` and enable tracing specifically for the h2 crate with an "internal" feature flag as part of the suite of features enabled as part of `http2`. This would imply introducing an additional, user-facing feature flag that would enable http2 without tracing.

---

-----BEGIN PHILOSOPHIC VALUE PROPOSITION-----

Such a change would make a big difference for anyone hesitating to use hyper for anything from telehealth to military communications. Telemetry is often a non-starter for these purposes. Audit processes often start with looking through the dependency graph of an application.

I understand that it may appear as if I'm introducing urgency in the form of moral obligation. That is a fair observation.

-----END PHILOSOPHIC VALUE PROPOSITION-----

PR #792, opened 2024-08-05
fix: send PROTOCOL_ERROR instead of REFUSED_STREAM for oversized headers

seanmonstar:
Noticed during #791 

cc @ajwerner


ajwerner (approved):
LGTM

PR #794, opened 2024-08-06
h2-support,h2-tests: add tools to ensure wake 

ajwerner:
This commit adds wrappers around futures::future helpers and augments
TestFuture to ensure that the underlying futures are notified before
they are polled. This helps to catch bugs where there are missing notify
calls or bad handling of the waker.

The commit then extends the tests to use these helpers instead of the
library functions from futures.

It also ammends the client_requests::recv_too_big_headers test to no
longer use the tokio spawned tasks that were added in https://github.com/hyperium/h2/pull/791.

### DIFF
--- a/tests/h2-support/src/future_ext.rs
+++ b/tests/h2-support/src/future_ext.rs
@@ -1,7 +1,9 @@
-use futures::FutureExt;
+use futures::{FutureExt, TryFuture};
 use std::future::Future;
 use std::pin::Pin;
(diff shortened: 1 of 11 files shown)


ajwerner:
First commit is https://github.com/hyperium/h2/pull/793, I'll rebase it away when that PR merges.

ajwerner:
Feel free to nitpick this, it was done hastily.

seanmonstar (approved):

PR #809, opened 2024-11-04
improve ci/h2spec.sh (macOS compat, /tmp dir and overwrite)

GlenDC:
- detect if run on MacOS, so we download h2spec macos build in that case
- support overwriting h2spec detection so we anyway download new file, useful in case you switch to new version for example
- move h2spec, archive and log all to /tmp dir as to not polute the repo dir

Note one or more follow up tasks might be required because this is also using an out of date h2spec archive. We are already a handful of major versions further. I tried to switch it over but there were a lot of failures. Is that because of:

- a) how we use `h2spec`
- b) how we define the `h2-server` example
- c) bugs in the `h2` crate?

Some guidance is appreciated here and if you have me I could help contribute some or all fixes for this? At the very least this hopefully gets the ball rolling on this as I do have intuitively the feeling that we might want to stay up to date with a tool we use, especially when it tests for h2spec compatibility / correctness?

This is however separate from this patch as this are just some improvements that allow it to be run on my dev machine as well as in CI, it should not affect the existing usage by (linux) users.

### DIFF
--- a/ci/h2spec.sh
+++ b/ci/h2spec.sh
@@ -1,10 +1,25 @@
 #!/bin/bash
-LOGFILE="h2server.log"
+LOGFILE="/tmp/h2server.log"
 
-if ! [ -e "h2spec" ] ; then
+override_h2spec=false
+
+# Check for optional flag
+while getopts "F" opt; do
+    case $opt in
+        F) override_h2spec=true ;;
+        *) echo "Usage: $0 [-o]"; exit 1 ;;
+    esac
+done
+
+if ! [ -e "/tmp/h2spec" ] || $override_h2spec ; then
     # if we don't already have a h2spec executable, wget it from github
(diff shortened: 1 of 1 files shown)


seanmonstar:
We just haven't remembered to keep up to date. What are the failures?

seanmonstar:
Sure, separate PR fixing those and upgrading would be great! I don't think they're concerning, it looks like differences on error conditions. Would you want to investigate and try to fix?

seanmonstar (approved):

GlenDC:
> Sure, separate PR fixing those and upgrading would be great! I don't think they're concerning, it looks like differences on error conditions. Would you want to investigate and try to fix?

Sure thing, added it to my whiteboard backlog, for a little snack this week. Will follow up. Thx!

PR #810, opened 2024-11-04
Fix RecvStream::is_end_stream(): return true only when END_STREAM is received

### LINKED ISSUE #806: RecvStream::is_end_stream() shouldn't return true after encountering error
From the [doc](https://docs.rs/h2/latest/h2/struct.RecvStream.html#method.is_end_stream)

> Returns true if the receive half has reached the end of stream.
> A return value of true means that calls to poll and poll_trailers will both return None.

However in reality, it would return true also after the connection / stream encounters errors because 
https://github.com/hyperium/h2/blob/77be6648e5c6da22f51e5049347496335a032288/src/proto/streams/recv.rs#L560-L562
and `is_recv_closed` is true regardless of the actual cause of the closure
https://github.com/hyperium/h2/blob/77be6648e5c6da22f51e5049347496335a032288/src/proto/streams/state.rs#L416-L420

The problem of `RecvStream::is_end_stream()` returning true after encountering error is that it hints the user that the http request finishes successfully. So complete responses might be interpreted as complete ones.

My suggestion is that this function should only return true after seeing END_STREAM flag.

eaufavor:
Before this change, it returned true on other types of disconnection as well.

Fixes #806

Removed unused `is_recv_closed()`.

### DIFF
--- a/src/proto/streams/state.rs
+++ b/src/proto/streams/state.rs
@@ -409,15 +409,13 @@ impl State {
         )
     }
 
-    pub fn is_closed(&self) -> bool {
-        matches!(self.inner, Closed(_))
+    pub fn is_recv_end_stream(&self) -> bool {
+        // In either case END_STREAM has been received
+        matches!(self.inner, Closed(Cause::EndStream) | HalfClosedRemote(..))
     }
 
-    pub fn is_recv_closed(&self) -> bool {
-        matches!(
-            self.inner,
-            Closed(..) | HalfClosedRemote(..) | ReservedLocal
-        )
+    pub fn is_closed(&self) -> bool {
+        matches!(self.inner, Closed(_))
     }
 
     pub fn is_send_closed(&self) -> bool {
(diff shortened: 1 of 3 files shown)


src/proto/streams/state.rs
```diff
@@ -413,11 +413,8 @@ impl State {
         matches!(self.inner, Closed(_))
     }
 
-    pub fn is_recv_closed(&self) -> bool {
-        matches!(
-            self.inner,
-            Closed(..) | HalfClosedRemote(..) | ReservedLocal
```
seanmonstar, line 419: I've given this some thought, and I'm less sure that we want the check to be so strict. If it is half-closed successfully, I think we _do_ want to say that the `RecvStream` is finished. If there is later an error on the sending side, I don't think that should be a problem, the `SendStream` will notice.

eaufavor: Good point. Updated to check both `Closed(Cause::EndStream)` and  `HalfClosedRemote(..)`. Also updated the function name accordingly.

seanmonstar (approved):

PR #813, opened 2024-11-11
Header frame with non-zero content-length and END_STREAM is malformed

eaufavor:
Before this change, content-length underflow is only checked when receiving date frames.

This change adds similar check for headers frames.


seanmonstar (approved):
Great catch, thanks!

seanmonstar:
I'm fixing the MSRV job in #814.

PR #820, opened 2024-11-25
tone down debugs from codec modules

lthiery:
Some logging frameworks don't really tolerate module level configurations well. I have quite a broad project but h2 seems to be the only dependency that has such chatty debug settings.

Would you be open-minded to removing all `debug!` in favor of `trace!`?

### DIFF
--- a/src/codec/framed_read.rs
+++ b/src/codec/framed_read.rs
@@ -401,7 +401,7 @@ where
                 partial,
                 bytes,
             )? {
-                tracing::debug!(?frame, "received");
+                tracing::trace!(?frame, "received");
                 return Poll::Ready(Some(Ok(frame)));
             }
         }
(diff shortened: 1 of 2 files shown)


seanmonstar:
This library already has a lot of noise at the trace level, the things at debug level are indeed helpful for many people to debug without needing the full firehose.

I don't think the solution is to reduce the helpfulness of the logs for those people, but to use a logger that allows you to adjust specific libraries if you don't want their logs.

lthiery:
I'm still trying to fix that but this is also the only crate that's causing so much chattiness. Perhaps if just the "send" in `frame_write` and "read" in `frame_read` could be downgraded?

lthiery:
@seanmonstar checking in to see if this level of change would acceptable? the scope has been narrowed to only the messages that I believe are very chatty. I do agree that a logging system should be capable of tuning logs per module, but I also believe that debug logging should be a little less active than h2 is with these two instances. 

Of course, up to you and will close and move on if you're not into it. Thanks for your work!

seanmonstar:
I still feel the same as [my previous comment](https://github.com/hyperium/h2/pull/820#issuecomment-2498787559).

PR #827, opened 2024-12-24
feat: Add keepalive timeout when connection idle

### LINKED ISSUE #826: Is there a way to configure h2 keepalive_timeout?
We want close non-active connection to avoid fd run out. https://nginx.org/en/docs/http/ngx_http_core_module.html#keepalive_timeout

I can open a PR if you accept.

1. increment number when there has stream https://github.com/nginx/nginx/blob/43191be96fb5f425710400e0064c0f2d285a352d/src/http/v2/ngx_http_v2.c#L3097
2. when hit timeout, check whether has stream active, or close h2 connection. https://github.com/nginx/nginx/blob/43191be96fb5f425710400e0064c0f2d285a352d/src/http/v2/ngx_http_v2.c#L629

taikulawo:
fix https://github.com/hyperium/h2/issues/826

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ atomic-waker = "1.0.0"
 futures-core = { version = "0.3", default-features = false }
 futures-sink = { version = "0.3", default-features = false }
 tokio-util = { version = "0.7.1", features = ["codec", "io"] }
-tokio = { version = "1", features = ["io-util"] }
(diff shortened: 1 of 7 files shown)


winrid:
Is there anything we can do to get this merged? Also this is just for http1.1 right, http2 has no concept of keepalive and we should have an idle timeout instead.

PR #830, opened 2025-01-16
Fix window size decrement of send-closed streams

nox:
Send-closed streams should be skipped when decreasing window size, as they are skipped when increasing it.

### DIFF
--- a/tests/h2-tests/tests/flow_control.rs
+++ b/tests/h2-tests/tests/flow_control.rs
@@ -1859,7 +1859,7 @@ async fn poll_capacity_wakeup_after_window_update() {
 }
 
 #[tokio::test]
-async fn window_size_decremented_past_zero() {
+async fn window_size_does_not_underflow() {
     h2_support::trace_init!();
     let (io, mut client) = mock::new();
 
@@ -1891,9 +1891,7 @@ async fn window_size_decremented_past_zero() {
         let builder = server::Builder::new();
         let mut srv = builder.handshake::<_, Bytes>(io).await.expect("handshake");
 
-        // just keep it open
-        let res = poll_fn(move |cx| srv.poll_closed(cx)).await;
-        tracing::debug!("{:?}", res);
+        poll_fn(move |cx| srv.poll_closed(cx)).await.unwrap();
     };
 
     join(client, srv).await;
--- a/src/proto/streams/send.rs
+++ b/src/proto/streams/send.rs
@@ -466,6 +466,16 @@ impl Send {
                     store.try_for_each(|mut stream| {
                         let stream = &mut *stream;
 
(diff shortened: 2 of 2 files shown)


src/proto/streams/flow_control.rs
```diff
@@ -136,16 +136,16 @@ impl FlowControl {
...
             sz,
             self.window_size,
             self.available
         );
-        // ~~This should not be able to overflow `window_size` from the bottom.~~ wrong. it can.
-        self.window_size.decrease_by(sz)?;
-        Ok(())
+        self.window_size
+            .decrease_by(sz)
+            .expect("window_size should be greater than decrement");
```
seanmonstar, line 148: Why the change to make this method panic instead of returning an error?

nox: Well it's not supposed to happen ever, and it used to be an assertion inside. At the very least it shouldn't be `FLOW_CONTROL_ERROR`.

nox: Ideally most of those `?` in flow control internals would be `INTERNAL_ERROR` with debug payloads explaining that a bug happened, imo.

nox: Would you prefer I put the `expect` at the call site, and not change the signature of `dec_send_window`?

seanmonstar: Well, looking at the changes here, it still seems like it's possible. It might not be _right_ to happen, but if it does, shouldn't a connection close be better than a panic?

Or is the change in `send.rs` truly the last place that could ever make this be wrong?

nox: I made it return an `INTERNAL_ERROR`, we can revisit later all of that to panic in debug builds.

tests/h2-tests/tests/flow_control.rs
```diff
@@ -1891,9 +1891,7 @@ async fn window_size_decremented_past_zero() {
         let builder = server::Builder::new();
         let mut srv = builder.handshake::<_, Bytes>(io).await.expect("handshake");
 
-        // just keep it open
-        let res = poll_fn(move |cx| srv.poll_closed(cx)).await;
```
olix0r, line 1895: Is this bug covered by this test? If not, is it possible to cover this bug in a test?

nox: Yes, it returned an error prior to my changes, now it doesn’t and it checks that it doesn’t.

seanmonstar (approved):

src/proto/streams/flow_control.rs
nox: Turns out I forgot to push the part where I made it `INTERNAL_ERROR`, oh well.

PR #833, opened 2025-01-20
Handle implicit resets at the right time

nox:
A stream whose ref count reaches zero while open should not immediately decrease the number of active streams, otherwise MAX_CONCURRENT_STREAMS isn't respected anymore.

### DIFF
--- a/src/proto/streams/counts.rs
+++ b/src/proto/streams/counts.rs
@@ -229,7 +229,7 @@ impl Counts {
                 }
             }
 
-            if stream.is_counted {
+            if !stream.state.is_scheduled_reset() && stream.is_counted {
                 tracing::trace!("dec_num_streams; stream={:?}", stream.id);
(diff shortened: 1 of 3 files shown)


src/proto/streams/recv.rs
```diff
@@ -914,7 +914,7 @@ impl Recv {
 
         tracing::trace!("enqueue_reset_expiration; {:?}", stream.id);
 
-        if counts.can_inc_num_reset_streams() {
+        if dbg!(counts.can_inc_num_reset_streams()) {
```
seanmonstar, line 917: Extra `dbg!` left in here.

nox:
I simplified the change a lot by just gating the `dec_num_streams` call in `transition_after` with `!stream.state.is_scheduled_reset()`.

seanmonstar (approved):

PR #838, opened 2025-02-20
Try to support Keepalive Timeout, expose 'streams check' for `Connection`

BamLubi:
Expose "streams check" to support keepalive timeout in external project.
We have test the idea in our project and found it success.

maybe solve issue #826


seanmonstar (approved):
LGTM, thanks!

PR #840, opened 2025-04-09
fix: `Stream` debug impl to not print `Waker`

LucioFranco:
This implements a custom `Debug` impl for `Stream` so that the `Waker` type stored in the `*_task` fields does not get printed. This is because the `Debug` impl of `Waker` prints a pointer which is non-deterministic. My use-case here is that I am trying to make turmoil deterministic and testing it with h2 via comparing trace logs between two same seeded runs.

### DIFF
--- a/src/proto/streams/stream.rs
+++ b/src/proto/streams/stream.rs
@@ -2,6 +2,7 @@ use crate::Reason;
 
 use super::*;
 
+use std::fmt;
 use std::task::{Context, Waker};
 use std::time::Instant;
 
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):
Wow, it'd be really nice to finally get `#[debug(hidden)]` etc attributes...

LucioFranco:
Attempted CI fix for msrv https://github.com/hyperium/h2/pull/841

LucioFranco:
> Wow, it'd be really nice to finally get #[debug(hidden)] etc attributes...

Maybe in Rust 2.0? 😂

Issue #853, opened 2025-07-22
Logical connection deadlock occurs under high concurrency - outgoing data transmission stops


sandersaares:
Under load test scenarios, I observe that our client app enters a state where outgoing traffic stops flowing. Here is a PR with a test that reproduces the behavior: #852

The test is timing-dependent but reproduces the problem 100% reliably on my PC, just gets stuck at different points each time I run it.

Typical output:

```text
Requests completed: 2 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000
Requests completed: 32 of 10000

thread 'logical_deadlock' panicked at tests\h2-tests\tests\deadlock.rs:117:17:
No requests completed in the last 2s, deadlock likely occurred
```

My best-effort investigation of the logs leading up to this suggests that when the timing aligns just right, we end up in a situation where flow control logic has allocated all connection capacity to requests that are in `pending_open` state but which never leave the `pending_open` state.

The leading theory is that this leads to a logical deadlock where:

* None of the open streams have any capacity allocated to them, so their data is not being sent. They cannot get more capacity because the connection's send window is exhausted (`conn=0`).
* All the streams that have assigned capacity are in `pending_open` state and are never leaving it (presumably because `max_concurrent_streams` has been reached).

PR #860, opened 2025-08-27
fix: do not assign capacity for pending streams

benjaminp:
`Prioritize::send_data` has a check to prevent assigning capacity to streams that are not yet open. Assigning flow control window to pending streams could starve already open streams.

This change moves the `stream.is_pending_open` check into `Prioritize::try_assign_capacity`. This prevents capacity from ever being assigned to pending streams. In particular, neither a client's `reserve_capacity` call nor a remote's initial window size adjustment will assign capacity to pending streams.

Tests `capacity_not_assigned_to_unopened_streams` and `new_initial_window_size_capacity_not_assigned_to_unopened_streams` in `flow_control.rs` demonstrates the fix.

A number of other tests must be changed because they were assuming that pending streams immediately received connection capacity.

This may be related to https://github.com/hyperium/h2/issues/853.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -90,14 +90,15 @@ jobs:
 
       - name: Pin some dependencies for MSRV
         run: |
-          cargo update --package tokio --precise 1.38.1
           cargo update --package tokio-util --precise 0.7.11
+          cargo update --package tokio --precise 1.38.1
           cargo update --package indexmap --precise 2.11.3
           cargo update --package hashbrown --precise 0.15.0
           cargo update --package once_cell --precise 1.20.3
           cargo update --package tracing --precise 0.1.41
           cargo update --package tracing-subscriber --precise 0.3.19
           cargo update --package tracing-core --precise 0.1.33
+          cargo update --package itoa --precise 1.0.15
 
 
       - run: cargo check -p h2


bvinc:
I am hitting this problem in production, and I've spent weeks tracking it down.  I am running a GRPC client that is creating thousands of streams to a server that limits streams to 400, and I keep hitting a deadlock.  I caught logs where h2 gives capacity to streams that are pending_open, and it triggers a deadlock where no more progress is made. I ended up with a patch to fix it that looks almost exactly like this, which lead to me finding this PR.  I support this PR.

Here's my question though:  Is it always a bug to assign capacity to a stream that is pending_open?  If so, do you think there should be a check inside of `try_assign_capacity` that returns early if the stream is pending_open?  If not, maybe there should at lest be a `debug_assert`?

vaibhav-shah:
We've had some folks use this patch in production, and it is behaving well.

rogerhu:
Any update here?

seanmonstar (approved):
OK, I see what this is doing: there used to be _one_ place that would prevent pending streams from getting capacity, but it was (let's say) forgotten in every other code path. So this fixes up that oversight by putting the conditional within the `try_assign_capacity` call itself.

That makes the change pretty clear to me, this seems like it was an oversight.

Thanks for investigating and submitting the fix, and the tests to show its behavior, and the patience waiting for a maintainer to have more attention (🙏 ). I'll merge and release this, and it should be an improvement for all, but we can handle any future bug reports that come in.

PR #865, opened 2025-11-19
Implement HTTP/2 informational responses (1xx) support

apu031:
Add support for HTTP/2 informational responses, enabling servers to send multiple 1xx status code responses before the final response and clients to receive them separately.

### Client-side Support
- Add `poll_informational()` method to `ResponseFuture` for polling 1xx responses
- Modify response polling logic to properly handle and skip informational headers
- Ensure informational responses don't interfere with main response flow

### Server-side Support
- Add `send_informational()` method to `SendResponse` with documentation
- Support multiple informational responses per request (e.g., 100 Continue, 103 Early Hints)
- Validate that only 1xx status codes are sent as informational responses
- Gracefully handle invalid informational responses without breaking the stream

### Protocol Implementation
- Add `InformationalHeaders` event type to distinguish from regular headers
- Implement `send_informational_headers_direct()` for sending without state changes
- Modify header processing to route informational responses correctly
- Ensure informational responses don't affect HTTP/2 stream state transitions

## Changed Files
- **Cargo.toml**: Switch to local http crate dependency for extended status code support
- **src/client.rs**: Add client-side informational response polling API
- **src/proto/streams/recv.rs**: Implement informational header event handling and polling
- **src/proto/streams/send.rs**: Add direct informational header sending capability
- **src/proto/streams/streams.rs**: Integrate informational response methods into stream APIs
- **src/server.rs**: Add server-side informational response API
- **tests/h2-tests/tests/informational_responses.rs**: Add 6 unit tests
      * send_100_continue: Basic 100 Continue functionality
      * send_103_early_hints: 103 Early Hints with Link headers
      * send_multiple_informational_responses: Multiple 1xx responses in sequence
      * invalid_informational_status_returns_error: Invalid status code handling
      * client_poll_informational_responses: Client-side polling API
      * informational_responses_with_body_streaming: Streaming scenarios

## Use Cases Supported
- **100 Continue**: Allow clients to continue sending request body
- **103 Early Hints**: Provide early resource hints for client preloading

## Technical Notes
- Informational responses are sent without the END_STREAM flag
- Stream state remains unchanged when sending informational responses
- Multiple informational responses can be sent before the final response
- Client polling separates informational responses from the main response flow
- Backward compatibility maintained - existing code continues to work unchanged

Resolves support for HTTP 103 Early Hints and 100 Continue HTTP/2 informational responses.

### DIFF
--- a/src/proto/streams/recv.rs
+++ b/src/proto/streams/recv.rs
@@ -66,6 +66,7 @@ pub(super) enum Event {
     Headers(peer::PollMessage),
     Data(Bytes),
     Trailers(HeaderMap),
+    InformationalHeaders(peer::PollMessage),
 }
 
 #[derive(Debug)]
@@ -264,6 +265,21 @@ impl Recv {
                 // corresponding headers frame pushed to `stream.pending_recv`.
                 self.pending_accept.push(stream);
             }
+        } else {
+            // This is an informational response (1xx status code)
+            // Convert to response and store it for polling
+            let message = counts
+                .peer()
+                .convert_poll_message(pseudo, fields, stream_id)?;
+
+            tracing::trace!("Received informational response: stream_id={:?}", stream_id);
+
+            // Push the informational response onto the stream's recv buffer
+            // with a special event type so it can be polled separately
+            stream
+                .pending_recv
+                .push_back(&mut self.buffer, Event::InformationalHeaders(message));
+            stream.notify_recv();
         }
 
         Ok(())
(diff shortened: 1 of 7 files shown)


src/proto/streams/recv.rs
```diff
@@ -324,24 +340,63 @@ impl Recv {
...
-                        stream.id,
-                        Reason::PROTOCOL_ERROR,
-                    )));
+        // Skip over any informational headers to find the main response
+        loop {
+            match stream.pending_recv.pop_front(&mut self.buffer) {
+                Some(Event::Headers(Client(response))) => return Poll::Ready(Ok(response)),
+                Some(Event::InformationalHeaders(_)) => {
+                    // Skip informational headers - they should be consumed by poll_informational
+                    continue;
```
pablocm: Is this backwards compatible? How did the h2 client behave before when a 1xx was received?

src/server.rs
```diff
@@ -1102,6 +1102,95 @@ impl Default for Builder {
...
+    /// use h2::server;
+    /// use http::{Response, StatusCode};
+    ///
+    /// // Send 100 Continue before processing request body
+    /// let continue_response = Response::builder()
+    ///     .status(StatusCode::CONTINUE)
+    ///     .header("x-custom", "value")
+    ///     .body(())
+    ///     .unwrap();
+    /// send_response.send_informational(continue_response)?;
```
pablocm: Since interim responses are not allowed to have a body, using `.body(())` is mandatory. If a caller attempts to add a body, should h2 treat as an error/fail state?

src/server.rs
```diff
@@ -1102,6 +1102,95 @@ impl Default for Builder {
...
+    /// Send an informational response (1xx status codes)
+    ///
+    /// This method can be called multiple times before calling `send_response()`
+    /// to send the final response. Only 1xx status codes are allowed.
+    ///
+    /// Informational responses are used to provide early feedback to the client
+    /// before the final response is ready. Common examples include:
+    /// - 100 Continue: Indicates the client should continue with the request
+    /// - 102 Processing: Indicates the server is processing the request
+    /// - 103 Early Hints: Provides early hints about resources to preload
```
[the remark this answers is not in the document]
apu031: Addressed. There is not no instance of 102 in any example. Since it is deprecated, it did not work when I tried implementing it. I have updated the doc.

src/server.rs
nox: > 11\. `ignore_invalid_informational_status` - Invalid status code handling (sends 200 as informational, should be ignored)

I don't think I agree with that, there is a concept of user error in the crate and this should probably be one.

apu031: @nox and @seanmonstar . Truly appreciate the keen observation. I missed to mimic that scenario while playing with NodeJS 103 behavior. I confirmed that NodeJS also sends an error if wrong status code is sent. I have fixed the code to throw a user error, and updated the test scenario accordingly.

apu031:
@seanmonstar @nox

Thank you both for your thoughtful reviews earlier. I truly appreciate the time and expertise you've shared on this PR.

I wanted to kindly follow up to see if there are any remaining concerns or blockers that I could address to help move this forward. This PR is a critical dependency for the final pull request in the hyper repository to complete the HTTP 103 Early Hints feature implementation, so merging it would be tremendously helpful.

Please let me know if there's anything I can clarify or improve. Thank you again for your consideration and support.

seanmonstar:
Sorry for the wait, it's a larger amount of new code, and I've been swamped with finishing some other work for the end of the year. I appreciate your patience, I'm hopeful I have time to review more closely this afternoon 🤓

src/proto/streams/recv.rs
```diff
@@ -264,6 +265,21 @@ impl Recv {
...
                 self.pending_accept.push(stream);
             }
+        } else {
+            // This is an informational response (1xx status code)
+            // Convert to response and store it for polling
+            let message = counts
+                .peer()
+                .convert_poll_message(pseudo, fields, stream_id)?;
+
+            tracing::trace!("Received informational response: {:?}", message);
```
seanmonstar: A message _could_ include sensitive information in the headers, so we try not to log the contents. If the trace itself is useful, perhaps just including the stream ID and the status code?

src/server.rs
```diff
@@ -1102,6 +1102,105 @@ impl Default for Builder {
...
+        let status = response.status();
+
+        tracing::trace!(
+            "send_informational called with status: {} on stream: {:?}",
+            status,
+            stream_id
+        );
+
+        // Validate that this is an informational response (1xx status code)
+        if !response.status().is_informational() {
```
seanmonstar: I noticed that these same checks also happen in `Send::send_interim_informational_headers()`, and in `Streams::send_informational_headers()`, which is called lower in this method, but there it will also emit helpful traces.

Is this the "only" place that could call the internal methods? If so, perhaps here is where the checks, traces, and error messages should happen. And later on, it can just be `assert!(the_things)` to catch if any other code path is added that doesn't check. Because at that point, the contract is on "us", the library, to uphold the invariants.

What do you think?

PR #877, opened 2026-02-12
fix: canonicalize Host header into :authority for outbound HTTP/2

### LINKED ISSUE #876: Outbound HTTP/2 frames can emit conflicting :authority and host headers
## Problem

When a user-provided `Host` header is present on an outbound HTTP/2 request, h2 emits it as a regular header field alongside the `:authority` pseudo-header derived from the URI. If the two values differ, this produces contradictory authority metadata on the wire.

For example, sending a request with URI `https://example.net/` and header `Host: example.com` results in a HEADERS frame containing both `:authority: example.net` and `host: example.com`.

## Expected behavior

Per [RFC 9113 §8.3.1](https://www.rfc-editor.org/rfc/rfc9113#section-8.3.1), clients must not generate a request containing inconsistent value in `Host` and `:authority`. The simplest way to guarantee consistency is to never emit `host` as a regular header on the wire for HTTP/2.

Other HTTP/2 implementations do this. For instance, `curl` with an explicit Host header (`curl -H 'host: foo.net' https://bar.com`) promotes the user-supplied `Host` value to `:authority` and strips `host` from regular headers entirely (verified via Wireshark). Go's `net/http` and Python's `httpx` behave the same way.

## Additional info

This issue is evident in Deno's `fetch` implementation, which is implemented with the hyper stack (including h2). We expect the following code to send a request with `:authority = example.net` (instead of `example.com`), but what's actually sent with Deno v2.6.9 is `:authority = example.com` and `host = example.net`, violating the HTTP/2 spec.

```ts
using client = Deno.createHttpClient({ allowHost: true });
const res = await fetch("https://example.com", {
  client,
  headers: {
    host: "example.net",
  },
});
console.log(res.status);
```

magurotuna:
When sending HTTP/2 requests or push promises, promote the first valid Host header value to :authority and retain it on the wire. Duplicate Host values are collapsed to a single entry matching :authority, and unparseable Host values are stripped. This prevents emitting conflicting :authority and Host while keeping the Host header available for intermediaries that may need it ([RFC 9113 §8.3.1][rfc]).
The "Host wins" semantics align with the behavior of other HTTP/2 client implementations such as curl, Go's std, and Python's httpx.

[rfc]: https://www.rfc-editor.org/rfc/rfc9113.html#section-8.3.1

Fixes #876

### DIFF
--- a/src/frame/headers.rs
+++ b/src/frame/headers.rs
@@ -6,23 +6,6 @@ use crate::hpack::{self, BytesStr};
 use http::header::{self, HeaderName, HeaderValue};
 use http::{uri, HeaderMap, Method, Request, StatusCode, Uri};
 
-/// Canonicalize `Host` header into `:authority` pseudo-header for HTTP/2.
-///
-/// - If a `Host` header is present, attempt to parse its first value as a URI authority.
-/// - On success, override `:authority` with the parsed value.
-/// - Always remove all `Host` headers from the regular header map.
-///
-/// Callers should only invoke this in an HTTP/2 context.
-pub(crate) fn canonicalize_host_authority(pseudo: &mut Pseudo, headers: &mut HeaderMap) {
-    if let Some(host) = headers.get(header::HOST) {
-        if let Ok(authority) = uri::Authority::from_maybe_shared(host.as_bytes().to_vec()) {
-            pseudo.set_authority(BytesStr::from(authority.as_str()));
-        }
-    }
-
-    headers.remove(header::HOST);
-}
-
 use bytes::{Buf, BufMut, Bytes, BytesMut};
 
 use std::fmt;
@@ -336,6 +319,41 @@ impl fmt::Debug for Headers {
 }
 
 // ===== util =====
+
+/// Canonicalize `Host` header into `:authority` pseudo-header for HTTP/2.
+///
+/// Per [RFC 9113 §8.3.1][rfc]:
+///
+/// - Clients that generate HTTP/2 requests directly MUST use the `:authority`
+///   pseudo-header field.
+/// - Clients MUST NOT generate a request with a `Host` header field that
+///   differs from the `:authority` pseudo-header field.
+/// - An intermediary that forwards a request over HTTP/2 MAY retain any
+///   `Host` header field.
+///
+/// This function enforces consistency using "Host wins" semantics: if a
+/// `Host` header is present and parseable as a URI authority, its value
+/// overrides `:authority`. The `Host` header is retained on the wire so
+/// that intermediaries can forward it (e.g. when proxying HTTP/1.1 traffic
+/// over an HTTP/2 transport).
+///
+/// [rfc]: https://www.rfc-editor.org/rfc/rfc9113.html#section-8.3.1
+pub(crate) fn canonicalize_host_authority(pseudo: &mut Pseudo, headers: &mut HeaderMap) {
+    if let Some(host) = headers.get(header::HOST) {
+        if let Ok(authority) = uri::Authority::from_maybe_shared(host.as_bytes().to_vec()) {
+            pseudo.set_authority(BytesStr::from(authority.as_str()));
+            // Collapse any duplicate Host values into a single value that
+            // matches the newly-set :authority.
+            if let Ok(normalized) = HeaderValue::from_str(authority.as_str()) {
+                headers.insert(header::HOST, normalized);
+            }
+        } else {
+            // Host is unparseable as an authority; remove it to avoid
+            // sending a value that differs from :authority.
+            headers.remove(header::HOST);
+        }
+    }
+}
 
 #[derive(Debug, PartialEq, Eq)]
 pub struct ParseU64Error;
--- a/src/frame/headers.rs
+++ b/src/frame/headers.rs
@@ -6,6 +6,23 @@ use crate::hpack::{self, BytesStr};
 use http::header::{self, HeaderName, HeaderValue};
 use http::{uri, HeaderMap, Method, Request, StatusCode, Uri};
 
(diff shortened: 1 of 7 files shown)


0x676e67 (approved):

seanmonstar:
@cratelyn @olix0r I feel like we considered this a long time ago, and for transparent proxy purposes, chose not to normalize? But I can't remember if we decided that explicitly, or not. Checking if you have cases where this would be problematic.

cratelyn:
> @cratelyn @olix0r I feel like we considered this a long time ago, and for transparent proxy purposes, chose not to normalize? But I can't remember if we decided that explicitly, or not. Checking if you have cases where this would be problematic.

@seanmonstar i wasn't around for that earlier consideration, but i pinged Oliver to take a look at this. i suspect he'll be better positioned to remember the details here :slightly_smiling_face: thank you for checking in with us!

cratelyn:
hi @seanmonstar, thank you for your patience.

looking at what this function does...

```rust
/// Canonicalize `Host` header into `:authority` pseudo-header for HTTP/2.
///
/// - If a `Host` header is present, attempt to parse its first value as a URI authority.
/// - On success, override `:authority` with the parsed value.
/// - Always remove all `Host` headers from the regular header map.
```

i believe the relevant portions of RFC 9113 are found in section 8.3.1. the strict rules i see for _clients_ are:

* Clients that generate HTTP/2 requests directly MUST use the ":authority" pseudo-header field
* Clients MUST NOT generate a request with a Host header field that differs from the ":authority" pseudo-header field

so host headers are fine for clients to emit, _so long as_ they do not differ from the ":authority" pseudo-header. removing them unconditionally would, i think, be going a step too far.

for intermediaries, the rules i see are:

* An intermediary that forwards a request over HTTP/2 MUST construct an ":authority" pseudo-header field using the authority information from the control data of the original request, unless the original request's target URI does not contain authority information (in which case it MUST NOT generate ":authority"
* An intermediary that forwards a request over HTTP/2 MAY retain any Host header field.
* An intermediary that needs to generate a Host header field (which might be necessary to construct an HTTP/1.1 request) MUST use the value from the ":authority" pseudo-header field as the value of the Host field, unless the intermediary also changes the request target. This replaces any existing Host field to avoid potential vulnerabilities in HTTP routing.

i think it's reasonable for h2 to enforce RFC 9113 compliance, and avoiding mismatched :authority and Host values seems like a wise decision, particularly because these MUST NOT differ.

i don't think that we should _always_ remove all `Host` headers, however. nothing says clients MUST NOT generate them, and intermediaries MAY retain them.

ideally, it'd be nice to leave a way for intermediaries to retain the host header field, per https://www.rfc-editor.org/rfc/rfc9113.html#section-8.3.1-2.3.6. that's likely important for intermediaries like us (linkerd) that transport HTTP/1 traffic over an HTTP/2 intermediary layer.

src/frame/headers.rs
```diff
@@ -6,6 +6,23 @@ use crate::hpack::{self, BytesStr};
 use http::header::{self, HeaderName, HeaderValue};
 use http::{uri, HeaderMap, Method, Request, StatusCode, Uri};
 
+/// Canonicalize `Host` header into `:authority` pseudo-header for HTTP/2.
```
cratelyn: nit: this is placed in the middle of the file's `use` imports. we should move this somewhere lower down. conventionally, following the pattern of other code in this repo, this should go after the `struct` definitions (including the eoponymous `Headers` type).

src/frame/headers.rs
```diff
@@ -6,6 +6,23 @@ use crate::hpack::{self, BytesStr};
...
+///
+/// - If a `Host` header is present, attempt to parse its first value as a URI authority.
+/// - On success, override `:authority` with the parsed value.
+/// - Always remove all `Host` headers from the regular header map.
+///
+/// Callers should only invoke this in an HTTP/2 context.
```
cratelyn: for future maintainers, it would also be helpful to add mention of the specific passages of RFC 9113 that this relates to. that would help keep the rationale of this code less implicit.

magurotuna:
Thanks for the thorough review and the RFC analysis, @cratelyn! I've addressed your feedback:

- Moved `canonicalize_host_authority` to the util section, after the struct definitions and impl blocks
- Added RFC 9113 §8.3.1 references in the doc comment
- Host is now retained on the wire instead of being stripped, so intermediaries can forward it. Duplicate values are collapsed to a single entry matching :authority, and unparseable values are stripped.

src/frame/headers.rs
```diff
@@ -320,6 +320,41 @@ impl fmt::Debug for Headers {
...
+/// This function enforces consistency using "Host wins" semantics: if a
+/// `Host` header is present and parseable as a URI authority, its value
+/// overrides `:authority`. The `Host` header is retained on the wire so
+/// that intermediaries can forward it (e.g. when proxying HTTP/1.1 traffic
+/// over an HTTP/2 transport).
+///
+/// [rfc]: https://www.rfc-editor.org/rfc/rfc9113.html#section-8.3.1
+pub(crate) fn canonicalize_host_authority(pseudo: &mut Pseudo, headers: &mut HeaderMap) {
+    if let Some(host) = headers.get(header::HOST) {
+        if let Ok(authority) = uri::Authority::from_maybe_shared(host.as_bytes().to_vec()) {
```
0x676e67: It's better to use `uri::Authority::try_from(host.as_ref())` here. This prevents early memory allocation by validating the input before committing to any buffer.

cratelyn (approved):
i'd love a sign-off from @seanmonstar, but this looks great to me. thank you for the additional changes, @magurotuna 🙂

sirfz:
Hello, is there anything preventing this PR from being merged?

magurotuna:
@sirfz not that I'm aware of. I think we just need wait for @seanmonstar's green light

Issue #880, opened 2026-03-25
Implicit RST_STREAM blocked until buffered DATA is fully drained


mmishra100:
When both `ResponseFuture` and `SendStream` handles are dropped for a stream that has buffered DATA frames, `maybe_cancel()` schedules an implicit reset via `schedule_implicit_reset(CANCEL)`. However, the resulting RST_STREAM is not sent until all buffered DATA frames have been drained through flow control first. If the stream's send window is exhausted and the server never sends a WINDOW_UPDATE, the DATA never drains and the RST_STREAM is indefinitely deferred.

## How this happens

`schedule_implicit_reset()` cannot call `clear_queue()` because it is invoked from type-erased `OpaqueStreamRef::drop()` which has no access to `Buffer<Frame<B>>`. It sets `ScheduledLibraryReset` on the stream state and defers to `pop_frame()`.

In `pop_frame()`, the `ScheduledLibraryReset` state is only checked after the stream's `pending_send` queue is fully drained (the `None` arm). Before that, `pop_frame()` attempts to send each buffered DATA frame through flow control like any normal frame. The implicit RST_STREAM effectively sits behind all buffered DATA in priority, rather than preempting it.

This contrasts with explicit `send_reset()` which works correctly — it calls `clear_queue()` to discard buffered DATA, queues RST_STREAM directly, and reclaims capacity immediately.

## Reproduction

Steps:

1. Set up a server that advertises `initial_window_size = 0` (or any small value that will be exhausted)
2. Client sends a POST request with `end_of_stream = false`
3. Client buffers DATA on the stream via `send_data()`
4. Client drops both `ResponseFuture` and `SendStream`
5. Observe that RST_STREAM is never sent — it is deferred behind the buffered DATA which cannot drain through the exhausted window

Test (for `tests/h2-tests/tests/flow_control.rs`):

```rust
/// When both handles of a stream are dropped while DATA frames are
/// buffered and the send window is exhausted, the implicit RST_STREAM
/// should be sent promptly.
///
/// Currently it is deferred until the buffered DATA is fully drained
/// through flow control, which never completes when the window is zero.
#[tokio::test]
async fn implicit_rst_stream_with_buffered_data_and_zero_window() {
    h2_support::trace_init!();

    let test = async {
        let (io, mut srv) = mock::new();

        let srv = async move {
            let settings = srv
                .assert_client_handshake_with_settings(
                    frames::settings().initial_window_size(0),
                )
                .await;
[...]

PR #881, opened 2026-03-26
fix: send RST_STREAM immediately for CANCEL resets (#880)

mmishra100:
When an implicit CANCEL reset is scheduled (both ResponseFuture and SendStream dropped), pop_frame() tried to drain buffered DATA through flow control before producing the RST_STREAM. If the send window was exhausted the DATA could never drain, so the RST_STREAM was never sent.

Now pop_frame() discards buffered DATA and sends RST_STREAM immediately for CANCEL resets. NO_ERROR resets still drain their data first, since they follow a complete response (RFC 9113 §8.1).

### DIFF
--- a/src/proto/streams/prioritize.rs
+++ b/src/proto/streams/prioritize.rs
@@ -731,6 +731,27 @@ impl Prioritize {
 
                     let frame = match stream.pending_send.pop_front(buffer) {
                         Some(Frame::Data(mut frame)) => {
+                            // If a CANCEL reset is scheduled, the stream is
+                            // "no longer needed" (RFC 9113 §7) and we can
+                            // discard buffered DATA and send RST_STREAM
+                            // immediately. Contrast with NO_ERROR, which is
+                            // sent after a complete response (RFC 9113 §8.1)
+                            // and requires all queued data to be sent first.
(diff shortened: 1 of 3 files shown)


src/proto/streams/prioritize.rs
```diff
@@ -731,6 +731,27 @@ impl Prioritize {
...
                         Some(Frame::Data(mut frame)) => {
+                            // If a CANCEL reset is scheduled, the stream is
+                            // "no longer needed" (RFC 9113 §7) and we can
+                            // discard buffered DATA and send RST_STREAM
+                            // immediately. Contrast with NO_ERROR, which is
+                            // sent after a complete response (RFC 9113 §8.1)
+                            // and requires all queued data to be sent first.
+                            //
+                            // See https://github.com/hyperium/h2/issues/880.
+                            if stream.state.get_scheduled_reset() == Some(Reason::CANCEL) {
```
seanmonstar, line 742: I appreciate the comments just above pointing out that `NO_ERROR` likely should keep sending. But, I'm wondering, is `CANCEL` the only reason to drop all the data? Shouldn't _any other_ reset tear it all down?

Issue #885, opened 2026-03-31
GOAWAY kills in-progress response when client sends GOAWAY(0)


ofek-sha:
## Bug

When a server receives a client-sent `GOAWAY(NO_ERROR, last_stream_id=0)`, `Inner::recv_go_away` closes **all** streams with `id > last_stream_id` - including client-initiated streams that the server is still responding on.

This kills in-progress response bodies before any DATA frames are sent.

## Steps to reproduce

1. Client opens a connection and sends a request (stream 1)
2. Server sends response headers (no body yet)
3. Client sends `GOAWAY(NO_ERROR, last_stream_id=0)` (e.g. client that doesn't want to reuse the connection after receiving response headers)
4. Server-side h2 closes stream 1 and the connection, before the response body is sent

In my case, a [pingora](https://github.com/cloudflare/pingora)-based reverse proxy using h2 for upstream connections. The client is the AWS SDK's `NodeHttp2Handler` with [`disableConcurrentStreams: true`](https://github.com/aws/aws-sdk-js-v3/blob/15cfc71196c1aa0437c2646ce5042173ed281142/clients/client-bedrock-runtime/src/runtimeConfig.ts#L85), which calls `session.close()` after receiving response headers, sending `GOAWAY(NO_ERROR, last_stream_id=0)` and killing the in-progress streaming response body.

```
Connection{peer=Server}: h2::codec::framed_write: send frame=Headers { stream_id: StreamId(1), flags: (0x4: END_HEADERS) }
Connection{peer=Server}: h2::codec::framed_read: received frame=GoAway { error_code: NO_ERROR, last_stream_id: StreamId(0) }
Connection{peer=Server}: h2::codec::framed_write: send frame=GoAway { error_code: NO_ERROR, last_stream_id: StreamId(1) }
Connection{peer=Server}: h2::proto::connection: Connection::poll; connection error error=GoAway(b"", NO_ERROR, Library)
```

The server immediately sends its own GOAWAY and errors the connection instead of completing the response on stream 1.

## Expected behavior

Stream 1 should continue to completion. The client's `GOAWAY(last_stream_id=0)` refers to server-push streams (even IDs), not client-initiated streams. The server should finish sending the response body, then close the connection gracefully.

## Root cause

In [`src/proto/streams/streams.rs`, `Inner::recv_go_away`](https://github.com/hyperium/h2/blob/5634dddea8ff9ed4e8df327a64765738f3e997d8/src/proto/streams/streams.rs#L730):

[...]

PR #886, opened 2026-03-31
fix: filter stream initiator in recv_go_away (#885)

ofek-sha:
recv_go_away closes all streams with id > last_stream_id, regardless of which endpoint initiated them. When a client sends GOAWAY(0) to a server, client-initiated streams (e.g. stream 1) are incorrectly closed, killing in-progress response bodies before any DATA frames are sent.

Add peer.is_local_init(stream.id) check so only locally-initiated streams above last_stream_id are closed, per RFC 9113 Section 6.8.

Added test client_goaway_does_not_kill_remote_initiated_streams that verifies the server completes a streaming response body on stream 1 after receiving client GOAWAY(NO_ERROR, last_stream_id=0). The test fails without the fix and passes with it.

### DIFF
--- a/src/proto/streams/streams.rs
+++ b/src/proto/streams/streams.rs
@@ -727,8 +727,9 @@ impl Inner {
 
         let err = Error::remote_go_away(frame.debug_data().clone(), frame.reason());
 
+        let peer = counts.peer();
         self.store.for_each(|stream| {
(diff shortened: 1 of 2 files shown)


seanmonstar (approved):
Nice job investigating, thanks so much!

PR #888, opened 2026-04-07
Add `header_table_size` to server builder

ArniDagur:
The client builder already has it. See https://github.com/hyperium/h2/pull/638


seanmonstar (approved):
Thanks!

PR #889, opened 2026-04-07
check final response in poll_informational

zh-jq:
Make sure `poll_informational` returns `Poll::Ready(None)` after the final response received. So the caller can then use the `ResponseFuture::poll()` method to get the final response.

The lock in `OpaqueStreamRef` get released when `poll_informational` returned, so if it didn't returns `Poll::Ready(None)` on received final response header, the caller has to call `ResponseFuture::poll()` every time to check the final response, this may get some informational response get ignored.

### DIFF
--- a/src/proto/streams/recv.rs
+++ b/src/proto/streams/recv.rs
@@ -377,6 +377,13 @@ impl Recv {
         // If it's not, we put it back
         if let Some(event) = stream.pending_recv.pop_front(&mut self.buffer) {
             match event {
+                Event::Headers(Client(response)) => {
+                    // Final response
+                    stream
+                        .pending_recv
+                        .push_front(&mut self.buffer, Event::Headers(Client(response)));
+                    return Poll::Ready(None);
+                }
                 Event::InformationalHeaders(Client(response)) => {
                     // Found an informational response, return it
                     return Poll::Ready(Some(Ok(response)));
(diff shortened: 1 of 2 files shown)


seanmonstar:
Thanks for the PR!

I looked at the code already in the function, and noticed some things. Since it matches `other`, it will get pushed back into the queue. Are you saying that `stream.state.ensure_recv_open()` will be true, even though the informational events are complete?

zh-jq:
This is the code I used:
```rust
    fn poll_recv(&mut self, cx: &mut Context<'_>) -> Poll<Result<Response<()>, Error>> {
        if let Some(r) = ready!(self.rsp_fut.poll_informational(cx)) {
            return Poll::Ready(r);
        }

        let r = ready!(Pin::new(&mut self.rsp_fut).poll(cx))?;
        // ...
    }
```
I have to know there won't be any informational headers so I can `poll` for the final response. I can't call `poll` directly when `poll_informational` return `Poll::Pending` as `poll` will drop informational headers silently.

seanmonstar:
Hm, ok. I wonder, would it possible to add a test to https://github.com/hyperium/h2/blob/master/tests/h2-tests/tests/informational_responses.rs that fails without this change? 🤓

zh-jq:
> Hm, ok. I wonder, would it possible to add a test to https://github.com/hyperium/h2/blob/master/tests/h2-tests/tests/informational_responses.rs that fails without this change? 🤓

Added one here: https://github.com/hyperium/h2/pull/890

seanmonstar (approved):
Got it, thanks for the test!

PR #890, opened 2026-04-14
add failed testcase for poll_informational

zh-jq:


zh-jq:
This shouldn't be merged as the code has been in https://github.com/hyperium/h2/pull/889

PR #892, opened 2026-04-27
fix: WINDOW_UPDATE 0-increment RFC conformance

ArniDagur:
Previously, `h2` emitted a connection level `GOAWAY(PROTOCOL_ERROR)` when a peer sent a `WINDOW_UPDATE` frame with an increment of zero, regardless of which stream the bad frame was on. This killed every stream on the connection. Instead, we should only reset that particular stream.

This is described in https://www.rfc-editor.org/rfc/rfc9113.html#section-6.9-9

See:
* https://github.com/golang/net/blob/89624e152475881f05bb88aca261239d1b45a833/http2/frame.go#L1015-L1020

### DIFF
--- a/src/codec/framed_read.rs
+++ b/src/codec/framed_read.rs
@@ -219,11 +219,28 @@ fn decode_frame(
         Kind::WindowUpdate => {
             let res = frame::WindowUpdate::load(head, &bytes[frame::HEADER_LEN..]);
 
-            res.map_err(|e| {
-                proto_err!(conn: "failed to load WINDOW_UPDATE frame; err={:?}", e);
-                Error::library_go_away(Reason::PROTOCOL_ERROR)
-            })?
-            .into()
+            match res {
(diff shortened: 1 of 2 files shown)


seanmonstar:
Do you mind mentioning what you hope to unlock? From the looks of the code, it will already return an error, right? Just that it's a connection error instead of stream?

ArniDagur:
This came from an audit of `h2` for bugs, by comparing against Golang / nghttp2. This is not a big issue as clients shouldn't be sending these zero increment updates, but it's technically RFC non-conformant, and also kills all streams rather than just the specific one.

seanmonstar:
Thanks for working to make h2 better. I've given this some more thought, and I think it's best to keep it the way it is. It is spec-compliant to turn any stream error into a connection error instead, and anything that could be abused is better to kill the connection.

ArniDagur:
Sounds good

PR #893, opened 2026-04-28
fix: Avoid capacity leak or panic when a stream is cancelled after `reserve_capacity`

ArniDagur:
The following triggers a bug

```
let (resp, mut send) = client.send_request(req, false)?;
send.reserve_capacity(N);
send.send_reset(Reason::CANCEL);
```

In builds with debug assertions enabled, we panic:

```
thread 'reserve_capacity_then_cancel_does_not_leak' panicked at src/proto/streams/prioritize.rs:445:9:
state=Closed(Error(Reset(StreamId(1), CANCEL, User)))
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: h2::proto::streams::prioritize::Prioritize::try_assign_capacity
             at src/proto/streams/prioritize.rs:445:9
   3: h2::proto::streams::prioritize::Prioritize::poll_complete
             at src/proto/streams/prioritize.rs:535:22
   4: h2::proto::streams::send::Send::poll_complete
             at src/proto/streams/send.rs:350:14
   5: h2::proto::streams::streams::Inner::poll_complete
             at src/proto/streams/streams.rs:906:34
   6: h2::proto::streams::streams::Streams<B,P>::poll_complete
             at src/proto/streams/streams.rs:186:12
   7: h2::proto::connection::Connection<T,P,B>::poll
             at src/proto/connection.rs:287:55
   8: <h2::client::Connection<T,B> as core::future::future::Future>::poll
             at src/client.rs:1441:33
```

In production builds, it instead leaks capacity, since the assert is compiled out.

### DIFF
--- a/src/proto/streams/prioritize.rs
+++ b/src/proto/streams/prioritize.rs
@@ -442,14 +442,10 @@ impl Prioritize {
             return;
         }
 
-        // If the stream has requested capacity, then it must be in the
-        // streaming state (more data could be sent) or there is buffered data
-        // waiting to be sent.
-        debug_assert!(
-            stream.state.is_send_streaming() || stream.buffered_send_data > 0,
-            "state={:?}",
-            stream.state
-        );
+        // The stream may have been reset or closed since capacity was requested.
+        if !stream.state.is_send_streaming() && stream.buffered_send_data == 0 {
+            return;
+        }
 
         // The amount of currently available capacity on the connection
         let conn_available = self.flow.available().as_size();
--- a/tests/h2-tests/tests/flow_control.rs
+++ b/tests/h2-tests/tests/flow_control.rs
@@ -1587,6 +1587,80 @@ async fn reset_stream_waiting_for_capacity() {
     join(srv, client).await;
 }
 
+// Regression test for https://github.com/hyperium/h2/pull/893
+#[tokio::test]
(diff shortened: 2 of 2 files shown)


src/proto/streams/prioritize.rs
```diff
@@ -442,14 +442,10 @@ impl Prioritize {
...
-        // streaming state (more data could be sent) or there is buffered data
-        // waiting to be sent.
-        debug_assert!(
-            stream.state.is_send_streaming() || stream.buffered_send_data > 0,
-            "state={:?}",
-            stream.state
-        );
+        // The stream may have been reset or closed since capacity was requested.
+        if !stream.state.is_send_streaming() && stream.buffered_send_data == 0 {
+            return;
```
seanmonstar, line 447: What do you think, worth a trace, or is it not interesting enough?

ArniDagur: I don't have a strong opinion here, but I lean towards not adding a trace.

1. The existing guard for the same condition at another location [does not log](https://github.com/hyperium/h2/blob/dbc204e57e0f96ea25d023c82d8a16340675b271/src/proto/streams/prioritize.rs#L396-L398)
2. I'm not sure what the actionable item would be for someone that sees the log

seanmonstar (approved):

PR #896, opened 2026-04-30
fix: Discard buffered DATA when a scheduled reset is pending

ArniDagur:
A revival of https://github.com/hyperium/h2/pull/881

Implicit resets rely on `pop_frame` reaching the `None` arm to emit `RST_STREAM`. If `DATA` is queued but the stream has zero window capacity to send it, `pop_frame` hits this check
```rust
if sz > 0 && stream_capacity == 0 {
    tracing::trace!("stream capacity is 0");

    // The stream has no more capacity, this can
    // happen if the remote reduced the stream
    // window. In this case, we need to buffer the
    // frame and wait for a window update...
    stream.pending_send.push_front(buffer, frame.into());

    continue;
}
```
The frame is pushed back, but the stream is not re-enqueued until we receive a window update for it. If that update is never received, the stream reset will be delayed indefinitely (we're at the mercy of the remote). In debug builds this crashes with

```
panicked at src/proto/streams/counts.rs:282:13:
assertion failed: !self.has_streams()
```
when the connection is dropped.

```rust
fn maybe_cancel(stream: &mut store::Ptr, actions: &mut Actions, counts: &mut Counts) {
    if stream.is_canceled_interest() {
        // Server is allowed to early respond without fully consuming the client input stream
        // But per the RFC, must send a RST_STREAM(NO_ERROR) in such cases. https://www.rfc-editor.org/rfc/rfc7540#section-8.1
        // Some other http2 implementation may interpret other error code as fatal if not respected (i.e: nginx https://trac.nginx.org/nginx/ticket/2376)
        let reason = if counts.peer().is_server()
            && stream.state.is_send_closed()
            && stream.state.is_recv_streaming()
        {
            Reason::NO_ERROR
        } else {
            Reason::CANCEL
        };

        actions
            .send
            .schedule_implicit_reset(stream, reason, counts, &mut actions.task);
        actions.recv.enqueue_reset_expiration(stream, counts);
    }
}
```
Implicit resets can occur under three circumstances:
* `CANCEL`: In [`maybe_cancel`](https://github.com/hyperium/h2/blob/1e68f995edec9e5c462d4e9af906e2d5304412c2/src/proto/streams/streams.rs#L1579-L1598), when all user handles are dropped for a stream that is still open
* `NO_ERROR`: In [`maybe_cancel`](https://github.com/hyperium/h2/blob/1e68f995edec9e5c462d4e9af906e2d5304412c2/src/proto/streams/streams.rs#L1579-L1598), when all user handles are dropped for a server stream where the response is already queued
* `PROTOCOL_ERROR`: From the oversized headers path in [`streams.rs`](https://github.com/hyperium/h2/blob/1e68f995edec9e5c462d4e9af906e2d5304412c2/src/proto/streams/streams.rs#L509-L513), after a rejection response is sent

For context, `NO_ERROR` for a stream reset is only supposed to be sent after a complete response. Buffering data + response, and then dropping the stream handles sounds like a relatively normal thing to do, and I think we want to support it.

However, for the other cases we should just fast-track stream reset and delete the buffered data, as we don't care about the stream anymore. Sending it would be wasteful.

### DIFF
--- a/src/proto/streams/prioritize.rs
+++ b/src/proto/streams/prioritize.rs
@@ -731,6 +731,23 @@ impl Prioritize {
 
                     let frame = match stream.pending_send.pop_front(buffer) {
                         Some(Frame::Data(mut frame)) => {
+                            if let Some(reason) = stream.state.get_scheduled_reset() {
+                                // If a reset is scheduled due to cancellation or
+                                // an error, discard buffered DATA and let the `None`
+                                // arm emit the RST_STREAM on the next iteration.
+                                //
+                                // NO_ERROR is excluded. Per RFC 9113 §8.1, a NO_ERROR
+                                // stream reset may only be sent after a complete
+                                // response, which requires sending all queued DATA.
+                                if reason != Reason::NO_ERROR {
+                                    stream.pending_send.push_front(buffer, frame.into());
+                                    self.clear_queue(buffer, &mut stream);
+                                    self.reclaim_all_capacity(&mut stream, counts);
+                                    self.pending_send.push(&mut stream);
+                                    continue;
+                                }
+                            }
+
                             // Get the amount of capacity remaining for stream's
                             // window.
                             let stream_capacity = stream.send_flow.available();
--- a/tests/h2-tests/tests/flow_control.rs
+++ b/tests/h2-tests/tests/flow_control.rs
@@ -1588,6 +1588,95 @@ async fn reset_stream_waiting_for_capacity() {
 }
 
 #[tokio::test]
+async fn scheduled_reset_with_buffered_data_sends_rst() {
+    h2_support::trace_init!();
+    let (io, mut srv) = mock::new();
+
+    let srv = async move {
+        let settings = srv
+            .assert_client_handshake_with_settings(frames::settings().initial_window_size(0))
+            .await;
+        assert_default_settings!(settings);
+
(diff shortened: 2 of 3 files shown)


seanmonstar (approved):

shikhar:
Hi @seanmonstar, can we get a h2 release with this?

I am attempting to upgrade hyper from 1.9.0 to 1.10.1 and started to hit the assertion this commit fixes. This does not happen with h2 patched to main.

---
per fable

> So the hyper bump is the trigger, but the bug is in h2. Hyper 1.10 rewrote the request-body pipe (hyper #4061): instead of reserve_capacity(1), it now
  reserves the exact chunk size and buffers the chunk while waiting for stream capacity. When the request is then canceled while that DATA is buffered at
  zero stream capacity, h2's pop_frame keeps re-buffering the frame and never reaches the arm that emits RST_STREAM — the canceled stream stays "counted"
  forever, and the assert fires on connection drop.

> This is precisely h2 #896 "Discard buffered DATA when a scheduled reset is pending" (commit 733bba7, May 12) — whose commit message even quotes this exact
  assertion. It landed one week after h2 0.4.14 was cut (May 5), and there's no 0.4.15 on crates.io yet as of today.

seanmonstar:
Fair, I'll try to get a release out on Monday.

PR #897, opened 2026-05-01
fix: `set_reset` must notify send task (missed wakeup)

ArniDagur:
`set_reset` notified `recv_task` and `push_task`, but not `send_task`. This meant `poll_capacity` and `poll_reset` were never awoken for library resets. For example when a WINDOW_UPDATE overflows.


ArniDagur:
Fixed merge conflict

seanmonstar (approved):
❤️

PR #900, opened 2026-05-03
fix: Use correct error codes for frame errors

ArniDagur:
Previously all frame errors produced `GOAWAY(PROTOCOL_ERROR)`. RFC 9113 requires specific error codes:

| Error                        | Old                | New                    | Comment                                      |
|------------------------------|--------------------|------------------------|----------------------------------------------|
| `BadFrameSize`               | `PROTOCOL_ERROR`   | `FRAME_SIZE_ERROR`     |                                              |
| `InvalidPayloadLength`       | `PROTOCOL_ERROR`   | `FRAME_SIZE_ERROR`     |                                              |
| `InvalidPayloadAckSettings`  | `PROTOCOL_ERROR`   | `FRAME_SIZE_ERROR`     |                                              |
| `InvalidInitialWindowSize`   | `PROTOCOL_ERROR`   | `FLOW_CONTROL_ERROR`   | New variant, split from InvalidSettingValue |
| `Hpack(_)`                   | `PROTOCOL_ERROR`   | `COMPRESSION_ERROR`    |                                              |
| `TooMuchPadding`             | `PROTOCOL_ERROR`   | `PROTOCOL_ERROR`       |                                              |
| `InvalidSettingValue`        | `PROTOCOL_ERROR`   | `PROTOCOL_ERROR`       |                                              |
| `InvalidWindowUpdateValue`   | `PROTOCOL_ERROR`   | `PROTOCOL_ERROR`       |                                              |
| `InvalidStreamId`            | `PROTOCOL_ERROR`   | `PROTOCOL_ERROR`       |                                              |
| `MalformedMessage`           | `PROTOCOL_ERROR`   | `PROTOCOL_ERROR`       |                                              |
| `InvalidDependencyId`        | `PROTOCOL_ERROR`   | `PROTOCOL_ERROR`       |                                              |

One motivation to fix this beyond RFC compliance is that it's a source of "uninteresting" discrepancies when doing comparative fuzzing between different HTTP2 implementations.

### DIFF
--- a/src/codec/framed_read.rs
+++ b/src/codec/framed_read.rs
@@ -160,7 +160,7 @@ fn decode_frame(
                 },
                 Err(e) => {
                     proto_err!(conn: "failed to load frame; err={:?}", e);
-                    return Err(Error::library_go_away(Reason::PROTOCOL_ERROR));
+                    return Err(Error::library_go_away(e.reason()));
                 }
             };
 
@@ -177,7 +177,7 @@ fn decode_frame(
                 },
                 Err(e) => {
                     proto_err!(conn: "failed HPACK decoding; err={:?}", e);
-                    return Err(Error::library_go_away(Reason::PROTOCOL_ERROR));
+                    return Err(Error::library_go_away(e.reason()));
                 }
             }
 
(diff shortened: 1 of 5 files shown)


seanmonstar:
This is probably fine. I just wanted to note some things from [RFC 9113 Section 5.4](https://httpwg.org/specs/rfc9113.html#rfc.section.5.4) (and we use for security reasons in h2 sometimes):

> Additionally, an endpoint MAY use any applicable error code when it detects an error condition; a generic error code (such as PROTOCOL_ERROR or INTERNAL_ERROR) can always be used in place of more specific error codes.

> An endpoint can end a connection at any time. In particular, an endpoint MAY choose to treat a stream error as a connection error.

Issue #902, opened 2026-05-05
Sending and receiving data frames introduces significant overhead


NedAnd1:
## What is the issue?
I was looking at the perf of requests & responses sent via h2 on a TCP stream versus the perf of raw messages sent back and forth over a TCP stream,
and found throughput with h2 to be reduced by an average of around 50%.
The following table compares the perf that I observed on the test VM that I used with h2 requests & responses versus raw messages:
```
OS: linux (kernel version 6.6)
Arch: arm64
TCP_NODELAY: True
Data Payload Size | Message Kind | Avg Throughput | Avg Round-Trip Latency
   4kiB           | h2 over TCP  | 0.97 GB/s      |   8.43 µs
   4kiB           | raw over TCP | 1.84 GB/s      |   4.45 µs
  16kiB           | h2...        | 2.68 GB/s      |  12.23 µs
    ...           | raw...       | 5.31 GB/s      |   6.17 µs
  64kiB           | h2...        | 3.54 GB/s      |  37.05 µs
    ...           | raw...       | 8.43 GB/s      |  15.54 µs
 256kiB           | h2...        | 4.64 GB/s      | 113.00 µs
    ...           | raw...       | 9.43 GB/s      |  55.57 µs
1024kiB           | h2...        | 4.28 GB/s      | 490.00 µs
    ...           | raw...       | 9.22 GB/s      | 227.40 µs
```

## Primary root causes
Currently, the encoder can only hold a single pending DATA frame, and blocks the acceptance of any new frame until the
previous one is fully flushed to the socket and reclaimed.
This means every DATA frame triggers its own write() syscall and TCP segment — even when the TCP send buffer could carry far more.
With TCP_NODELAY enabled, there is no Nagle buffering to coalesce these small writes, so the per-frame overhead is fully realized.

On the receiver side, the recv buffer used by the frame decoder is small by default, and even if its buffer size was increased,
the effective capacity of that buffer would degrade over time and never recover, causing small `read()` calls.
This is because:
1. The buffer is repeatedly split to yield decoded frames.
2. If the recv buffer runs out of capacity, BytesMut::reserve is called, & if that buffer is still in use (likely),
    the current buffer is abandoned & a new buffer with a maximum capacity of 64kB would be created
    due to how BytesMut encodes its original capacity to a max of 64kB & caps new reservations
    based on that encoded original capacity.

Together, these things inflate the syscall-to-payload ratio on both ends.

NedAnd1:
[...]

PR #903, opened 2026-05-05
perf: allow multiple DATA frames per write

NedAnd1:
#### Summary
This PR addresses the sender-side perf issue described in #902
by allowing multiple DATA frames to be coalesced into a single write_vectored() syscall,
rather than the current behavior of flushing each DATA frame individually.

#### Problem
The existing encoder holds at most one pending DATA frame at a time and blocks accepting new frames until it is fully flushed. With TCP_NODELAY enabled, each frame triggers its own write() syscall and TCP segment — inflating the syscall-to-payload ratio and cutting throughput roughly in half compared to raw TCP.
  
#### Solution
   - Replace the single next `Option<Next<B>>` slot with a `VecDeque<BufElement<B>>` that can queue multiple DATA frames (up to 512 on vectored-IO transports).
   - Implement `Buf` directly on the Encoder, providing `chunks_vectored()` so that all queued frame headers + payloads are written in one `poll_write_vectored()` call.
   - Replace the per-connection `in_flight_data_frame` field with a per-stream `in_flight_partial_send` field of type `Option<ControlFlow<()>>`, allowing each stream to send up to one partial data frame at a time without blocking other streams.
   - Introduce `take_used_data_frames()` iterator to reclaim all fully-written frames in a single pass, replacing the old `take_last_data_frame()` which could only return one.
   - Add a custom `poll_write_buf` that returns `ControlFlow` to cleanly distinguish "nothing left to write" from "wrote some bytes, keep going".
   
### Validation
- h2 cargo tests
- external scenario tests
- external perf tests

Perf results for a branch including the commit in this PR:
```
   OS: linux (kernel 6.6), Arch: arm64, TCP_NODELAY: True

   Data Frame Size │ master (throughput) │ batch-data-frames (throughput) │ Improvement
      4 kiB        │ 0.97 GB/s           │ 1.14 GB/s                      │ 1.18x
     16 kiB        │ 2.68 GB/s           │ 3.47 GB/s                      │ 1.29x
     64 kiB        │ 3.54 GB/s           │ 6.19 GB/s                      │ 1.75x
    256 kiB        │ 4.64 GB/s           │ 7.50 GB/s                      │ 1.62x
   1024 kiB        │ 4.28 GB/s           │ 7.81 GB/s                      │ 1.82x
```
Receiver-side PR: #904

### DIFF
--- a/src/codec/framed_write.rs
+++ b/src/codec/framed_write.rs
@@ -4,12 +4,13 @@ use crate::frame::{self, Frame, FrameSize};
 use crate::hpack;
 
 use bytes::{Buf, BufMut, BytesMut};
+use std::collections::VecDeque;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
-use tokio_util::io::poll_write_buf;
 
-use std::io::{self, Cursor};
+use std::io::{self, Cursor, IoSlice};
+use std::ops::ControlFlow;
 
 // A macro to get around a method needing to borrow &mut self
 macro_rules! limited_write_buf {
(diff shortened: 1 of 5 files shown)


sprasanna-oai:
@seanmonstar any update on when this might be merged in?

seanmonstar:
I was distracted with a separate large refactor, but I'm less certain the big one will work. I'm turning attention to smaller performance wins. (Sorry about the merge conflicts.)

I think this could be a good improvement! Originally we were concerned about being able to quickly write control frames when needed, but I think this should be fine in practice. Especially if after a batch is started and a partial write happens, a follow-up only needs to finish the current partial DATA frame, and then could allow control frames through again.

PR #904, opened 2026-05-06
feat: add recv buffer size option with improved recv buffer capacity recovery

NedAnd1:
### Summary
This PR addresses the receiver-side perf issue described in #902
by allowing the frame decoder's read buffer size to be configurable
and by mitigating its capacity degradation with a simple a buffer manager.

### Problem
The recv buffer size of the frame decoder is small, and even if its buffer size was increased,
the effective capacity of that buffer would degrade over time and never recover, causing small `read()` calls.

### Solution
Adds a recv_buffer_size option to both `client::Builder` and `server::Builder`,
allowing users to control the initial capacity of the frame decoder's read buffer.

Introduces a `BufferManager` that proactively recovers buffer capacity for decoders.
When the underlying decoder returns Ok(None) (no complete frame available)
and the buffer's capacity has fallen below half the initial capacity,
the manager replaces the primary buffer with a secondary buffer
that has a higher likelihood of being able to reclaim its original buffer space.

This mitigates the BytesMut capacity degradation problem without depending on upstream changes.

### Validation
- h2 cargo tests
- external scenario tests
- external perf tests

Perf results for a branch including the commits in this PR:
```
   OS: linux (kernel 6.6), Arch: arm64, TCP_NODELAY: True

   Data Frame Size │ master (throughput) │ batch-data-frames (throughput) │ Improvement
      4 kiB        │ 0.97 GB/s           │ 1.14 GB/s                      │ 1.18x
     16 kiB        │ 2.68 GB/s           │ 3.47 GB/s                      │ 1.29x
     64 kiB        │ 3.54 GB/s           │ 6.19 GB/s                      │ 1.75x
    256 kiB        │ 4.64 GB/s           │ 7.50 GB/s                      │ 1.62x
   1024 kiB        │ 4.28 GB/s           │ 7.81 GB/s                      │ 1.82x
```

Sender-side PR: #903

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -324,6 +324,9 @@ pub struct Builder {
     /// Maximum amount of bytes to "buffer" for writing per stream.
     max_send_buffer_size: usize,
 
+    /// Maximum number of bytes to read at a time (for the entire connection).
+    recv_buffer_size: usize,
+
     /// Maximum number of locally reset streams to keep at a time.
     reset_stream_max: usize,
 
@@ -655,6 +658,7 @@ impl Builder {
     pub fn new() -> Builder {
         Builder {
             max_send_buffer_size: proto::DEFAULT_MAX_SEND_BUFFER_SIZE,
+            recv_buffer_size: proto::DEFAULT_RECV_BUFFER_SIZE,
             reset_stream_duration: Duration::from_secs(proto::DEFAULT_RESET_STREAM_SECS),
             reset_stream_max: proto::DEFAULT_RESET_STREAM_MAX,
             pending_accept_reset_stream_max: proto::DEFAULT_REMOTE_RESET_STREAM_MAX,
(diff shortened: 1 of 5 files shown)


seanmonstar:
Improving the recv buffer performance is a great idea! Though, I wonder why you swap between two? With hyper's http1 path, we keep track of a target minimum read amount (though doesn't have to be dynamic), and before each new message, a call to `reserve(min_read_size)` makes sure that the `BytesMut` has enough space. Is that not sufficient?

NedAnd1:
> Improving the recv buffer performance is a great idea! Though, I wonder why you swap between two? With hyper's http1 path, we keep track of a target minimum read amount (though doesn't have to be dynamic), and before each new message, a call to `reserve(min_read_size)` makes sure that the `BytesMut` has enough space. Is that not sufficient?

Thanks, that approach does lead to higher memory allocation churn.
Calling `reserve(...)` on the current buffer has a very high likelihood of needing a new allocation if it's below the threshold,
since the yielded data frames are still in use by higher-level components.
Calling `reserve(...)` on the secondary buffer is much more likely to reclaim that buffer's original space,
since it has a much smaller chance of still being referenced by higher-level components.

This observation was initially made by explicitly calling `BytesMut::try_reclaim(capacity - len + 1)`
at different locations with a single buffer, and it consistently failed,
especially in the location where the buffer is most likely to be empty.
Whereas `BytesMut::try_reclaim` on the secondary buffer was consistently succeeding
in the location where the primary buffer is most likely to be empty.

Any additional thoughts or questions?

seanmonstar:
I see what you're trying to fix, that makes a lot of sense! There's a possible concern with it, though. It would mean that probably in most cases, every connection would now have 2 read buffers allocated, likely doubling the amount of memory it was holding on to.

I've heard from a few others the desire to implement a buffer provider, so someone could decide how best to pool buffers between a bunch of connections. I'll link the discussion here: https://github.com/hyperium/hyper/discussions/4139

PR #911, opened 2026-06-08
fix: Remove unnecessary clones

FranciscoTGouveia:
If I am not overlooking anything, these clones can be safely removed since they aren't used anywhere else later.

CC @nunoplopes


seanmonstar:
Looks like one of the clones was needed after all. 🤓

FranciscoTGouveia:
Oh, this was an oversight on my end.
I have only run `cargo test --all-features`, and `pending_send_request_gets_reset_by_peer_properly` wasn't triggered.
The PR should be fixed now.

seanmonstar (approved):

PR #912, opened 2026-06-11
Return an error early if header list is overrun

IgorBelyi:
I've ran into a problem that excessive headers still consume CPU even after condition of exceeding `max_header_list_size` is detected. This PR is a proposed fix for that problem. It allows closure passed into `Decoder::decode()` to report that it won't be accepting any more headers thus allowing decoder to return earlier skipping unnecessary memory allocations. I did verify locally that CPU overrun is addressed with this change.

### DIFF
--- a/src/hpack/decoder.rs
+++ b/src/hpack/decoder.rs
@@ -179,7 +179,7 @@ impl Decoder {
         mut f: F,
     ) -> Result<(), DecoderError>
     where
-        F: FnMut(Header),
+        F: FnMut(Header) -> bool,
     {
         use self::Representation::*;
 
@@ -198,13 +198,13 @@ impl Decoder {
             // At this point we are always at the beginning of the next block
             // within the HPACK data. The type of the block can always be
             // determined from the first byte.
-            match Representation::load(ty)? {
+            if match Representation::load(ty)? {
                 Indexed => {
                     tracing::trace!(rem = src.remaining(), kind = %"Indexed");
                     can_resize = false;
                     let entry = self.decode_indexed(src)?;
                     consume(src);
-                    f(entry);
+                    f(entry)
                 }
                 LiteralWithIndexing => {
                     tracing::trace!(rem = src.remaining(), kind = %"LiteralWithIndexing");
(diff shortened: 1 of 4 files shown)


IgorBelyi:
Ok, looking into the test failures. Didn't realize there's more than just `cargo test`.

src/hpack/decoder.rs
```diff
@@ -179,7 +179,7 @@ impl Decoder {
         mut f: F,
     ) -> Result<(), DecoderError>
     where
-        F: FnMut(Header),
+        F: FnMut(Header) -> bool,
```
seanmonstar, line 182: I think returning a bool can be somewhat confusing, both for the calling (do I return true or false), and when it's used (what does that if mean again).

I'd suggest either using `ControlFlow`, or a `Result`. I think. What do you think of that?

src/hpack/decoder.rs
```diff
@@ -243,7 +243,10 @@ impl Decoder {
                     // Handle the dynamic table size update
                     self.process_size_update(src)?;
                     consume(src);
+                    false
                 }
+            } {
+                break;
```
seanmonstar, line 249: What's the effects of skipping the rest? Does it the table get out of sync?

Would it safer to have some heuristic that just says, "oversize by some we'll just ignore and reset you, but oversize by double (or whatever) and we'll GOAWAY your connection, you bad bad peer"?

src/hpack/decoder.rs
IgorBelyi: I was trying to stick with smallest change possible but `ControlFlow` indeed makes sense. Will switch.

src/hpack/decoder.rs
IgorBelyi: I'm not sure about that. Will need to take a closer look at how it fits into global flow. I was trusting tests to catch anything fishy and just wanted to improve performance to skip unnecessary code.

IgorBelyi:
Closing in favor of a better thought through https://github.com/hyperium/h2/pull/915

PR #915, opened 2026-06-12
fix: close connection when header size is way too large

seanmonstar:
Inspired by the problem highlighted in #912, this will now close a connection if decoded headers are "way" too large.

HPACK requires that we keep decoding no matter what. If a set of headers are larger than the configured limit, we stop saving them, mark them oversize, and then send a RST_STREAM about it. But we keep decoding, because the HPACK table is connection-level state.

The size of the table is bounded, so old entries keep getting evicted. That is fine. But it can mean a lot of CPU usage to ignore them.

This PR says `max_header_list_size * 4` is way too big. You go away if you send something way too big.

cc @IgorBelyi this is kind of what I had in mind.

### DIFF
--- a/src/codec/framed_read.rs
+++ b/src/codec/framed_read.rs
@@ -175,6 +175,13 @@ fn decode_frame(
                     proto_err!(stream: "malformed header block; stream={:?}", id);
                     return Err(Error::library_reset(id, Reason::PROTOCOL_ERROR));
                 },
+                Err(frame::Error::HeaderListWayTooLarge) => {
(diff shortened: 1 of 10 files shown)


IgorBelyi:
Nice! So, keep going after the limit until it becomes an annoyance and then close the connection. I'll close my PR in favor of this one. I've confirmed that it solves the problem in my case.

IgorBelyi:
@seanmonstar Thank you for merging! Is it possible to know the target date when that change be available in the published `h2` crate?

seanmonstar:
I was aiming for next Monday to do a release.

PR #917, opened 2026-06-16
Refactor big mutex to reduce contention

seanmonstar:
This refactor puts the big internal shared mutex on a diet. The goal is to reduce contention, providing performance improvements when `h2` is used on multi-threaded runtimes.

> [!Warning]
> This is a large internal refactor, and it hasn't _yet_ been sufficiently tested in a production deployment. While I tried to keep the behavior exactly the same, it's possible there's subtle changes or bugs.
>
> If you do test this, let me know. I hope to have this tested thoroughly before possibly landing.

### Architectural changes

There is now unique ownership of much of the connection state, owned solely by the "connection task".

Most conceptual things were split between an owned handle and shared values  such as `counts::Counts` and `counts::Shared`. The `Store` is now uniquely owned by the connection, and the `Stream`s stored in it contain an `Arc<stream::Shared>` for the parts that need to be shared with a "stream handle".

Stream handle operations now do very little. For receiving, it locks the shared connection `recv` buffer, pops a frame, and unlocks.

Likewise, when sending data, (or any other status-updating operation), it briefly locks a connection `pending_streams` queue, inserts itself, and unlocks. All actual changes are written on the _stream's_ `pending_ops` field. The connection task includes a loop to briefly lock the queue, pop a stream, unlock, and then process the streams pending ops.

So, while this does introduce _more_ locks, they are finer-grained, and should only be held very briefly. Longer "work" is no longer done while holding any lock.

This _should_ result in tasks needing to wait less time for another task that previously was holding the world-lock.

cc #531

Issue #531, opened 2021-04-21
Library does not scale with multiple cores


bIgBV:
As demonstrated by the benchmarks in [this reddit post](https://old.reddit.com/r/grpc/comments/muy8dj/grpc_bench_opensource_objective_grpc_benchmark/), you can see the `rust_tonic_mt` benchmark falling behind in performance as the number of threads are increased. 

The likely cause for this could be that a big portion of the shared state is behind [this `Mutex`](https://github.com/hyperium/h2/blob/2c53d600989209e263c8157df6e6e2280a9a3355/src/proto/streams/streams.rs#L25).

bIgBV:
A couple of things to try:

* Replace the `std::sync::mutex` with one from [parking_lot](https://crates.io/crates/parking_lot)
* Update hyper to keep all streams from a connection on a single thread.

seanmonstar:
A bigger effort would be to replace the single massive lock to per-stream locks. There might still need to be a large lock, but the goal would be to not need to keep it locked for long or for most operations.

It's not exactly the same here, but grpc-go did a similar change a couple years ago to reduce contention: https://github.com/grpc/grpc-go/pull/1962

bIgBV:
Some initial measurements from replacing the lock in [`streams.rs](https://github.com/hyperium/h2/blob/a6b414458fd7687f53df68861f4833cf142e5b76/src/proto/streams/streams.rs#L25) with one from parking lot:

https://gist.github.com/bIgBV/4d6d76773a948734ebef1367ef5221d5

w41ter:
@bIgBV It seems that the comparison results of parking_lot and the original implementation are similar?

notgull:
The libstd benchmark was recently replaced with a new implementation that is both much smaller and significantly faster. There is much less to lose now with per-stream locking.

jeffutter:
Resurrecting this old issue, but I think I'm hitting this bottleneck fairly acutely. I'm experimenting with using `tonic` to build something like a load-balancing proxy between grpc streams. I have X clients connecting over Y connections each with Z streams. I then load balance the requests (mostly 1-1 request-response type requests) across I connections each with J streams to K downstream servers.

I was seeing fairly disappointing performance. If I have the external clients hit the backends directly I'm requsets are taking ~200μs at a certain load level. With the proxy in play it's closer to 1ms. I started digging into this bottleneck and found this github issue.

[...]

PR #918, opened 2026-07-04
Avoid flushing h2 socket while holding stream locks

brentechols:
# Avoid flushing h2 socket while holding stream locks

Related: #531

There is also an active broader draft refactor in #917. This PR is intentionally smaller: it targets the specific path discussed in #531 where `Streams::poll_complete` can hold stream/send-buffer locks while the codec flush path reaches the socket.

## Summary

This change splits HTTP/2 stream draining into two phases so the connection task no longer holds the shared stream-state and send-buffer mutexes while flushing or polling writes on the underlying socket.

Previously, `Streams::poll_complete` held `Inner` and `SendBuffer` locks while `Prioritize::poll_complete` could call into `Codec::poll_ready` / `Codec::flush`, which can reach `poll_write_buf` on the socket. Under high write concurrency on one HTTP/2 connection, stream producer tasks contend with the connection task while the connection task is doing I/O progress.

The new flow is:

- ensure codec/socket write readiness before taking stream locks
- lock stream state and buffer pending frames only while the codec has local buffer capacity
- release the locks
- flush the codec/socket
- briefly re-lock to reclaim partially/fully written DATA frames and requeue as needed

The patch also adds a targeted write-contention benchmark that sends many response body chunks concurrently over a single HTTP/2 connection. It can be run with:

```sh
H2_BENCH=write-contention cargo bench --bench main
```

The benchmark is configurable:

```sh
H2_WRITE_CONTENTION_STREAMS=512
H2_WRITE_CONTENTION_CHUNKS_PER_STREAM=128
H2_WRITE_CONTENTION_CHUNK_SIZE=16384
H2_WRITE_CONTENTION_WORKER_THREADS=4
```

## Benchmark

Run environment:

- WSL Ubuntu 24.04
- Rust 1.96.1 stable
- workload: 512 streams x 128 chunks x 16 KiB = 1 GiB over one HTTP/2 connection

Median elapsed time over 3 alternating runs per worker-thread count:

| Worker threads | Baseline median | Patched median | Change |
| ---: | ---: | ---: | ---: |
| 1 | 336 ms | 337 ms | roughly flat |
| 2 | 358 ms | 238 ms | 33.5% faster |
| 4 | 382 ms | 337 ms | 11.8% faster |
| 8 | 407 ms | 340 ms | 16.5% faster |

Median throughput:

| Worker threads | Baseline | Patched |
| ---: | ---: | ---: |
| 1 | 3043 MiB/s | 3038 MiB/s |
| 2 | 2853 MiB/s | 4294 MiB/s |
| 4 | 2679 MiB/s | 3037 MiB/s |
| 8 | 2513 MiB/s | 3004 MiB/s |

The one-worker case is effectively flat, as expected: there is little cross-thread lock contention. The multi-worker runs show the intended effect.

## Testing

```sh
cargo fmt --all --check
cargo check
cargo test -p h2-tests reserved_capacity_assigned_in_multi_window_updates -- --exact --nocapture
cargo test -p h2-tests small_data_frames_reclaimed_before_buffering_next -- --exact
```

The new `small_data_frames_reclaimed_before_buffering_next` regression test fails on the pre-fix PR commit (`3a42547`) with the `in_flight_data_frame` assertion, and passes after reclaiming the DATA frame before buffering another frame.

The write-contention benchmark was run before/after this change in WSL Ubuntu 24.04. On this Windows machine, checking the benchmark target currently stops in the existing `tokio-rustls` dev-dependency chain because `aws-lc-sys` needs additional native build tools (`cmake`/NASM) before it reaches h2 code.

### DIFF
--- a/src/proto/streams/prioritize.rs
+++ b/src/proto/streams/prioritize.rs
@@ -543,6 +543,12 @@ impl Prioritize {
                         self.in_flight_data_frame = InFlightData::DataFrame(frame.payload().stream);
                     }
                     dst.buffer(frame).expect("invalid frame");
+
+                    // Small DATA frames can be fully encoded by `buffer`,
+                    // which records completion in a single codec slot. Reclaim
+                    // before accepting another frame so that slot is not
+                    // overwritten.
+                    self.reclaim_frame(buffer, store, dst);
                 }
                 None => {
                     return Ok(true);
--- a/src/proto/streams/prioritize.rs
+++ b/src/proto/streams/prioritize.rs
@@ -10,7 +10,7 @@ use bytes::buf::Take;
 use std::{
     cmp::{self, Ordering},
     fmt, io, mem,
-    task::{Context, Poll, Waker},
+    task::Waker,
 };
 
 /// # Warning
@@ -505,30 +505,30 @@ impl Prioritize {
         }
     }
 
-    pub fn poll_complete<T, B>(
+    pub fn buffer_pending<T, B>(
         &mut self,
-        cx: &mut Context,
         buffer: &mut Buffer<Frame<B>>,
         store: &mut Store,
         counts: &mut Counts,
         dst: &mut Codec<T, Prioritized<B>>,
-    ) -> Poll<io::Result<()>>
+    ) -> io::Result<bool>
     where
         T: AsyncWrite + Unpin,
         B: Buf,
     {
-        // Ensure codec is ready
-        ready!(dst.poll_ready(cx))?;
-
         // Reclaim any frame that has previously been written
         self.reclaim_frame(buffer, store, dst);
 
         // The max frame length
         let max_frame_len = dst.max_send_frame_size();
 
-        tracing::trace!("poll_complete");
+        tracing::trace!("buffer_pending");
 
         loop {
+            if !dst.has_send_capacity() {
+                return Ok(false);
+            }
+
             if let Some(mut stream) = self.pop_pending_open(store, counts) {
                 self.pending_send.push_front(&mut stream);
                 self.try_assign_capacity(&mut stream);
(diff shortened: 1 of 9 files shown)


seanmonstar:
Very cool! Also, a few weeks ago I put a significant rewrite in #917.

(Notably, I don't personally have great multi threaded testing environments.)

brentechols:
Ya I realized that after working on this :) maybe not needed then? I'll leave it up to you - I'm still reading through the PR to understand how everything is changing. Is this still required after the refactor?

seanmonstar:
Well, I can't guarantee what I did is good, though conceptually it moves more towards queues which tends to help. Some reports in that issue suggest good numbers, now that I've fixed up most bugs in it.

If you have an environment that can stress test any sorts of real world loads, more data is always welcome! (If it turns out what I did is bad, what you did here would be a great step!)

brentechols:
Let me see what I can do - should be able to take a glance at this at some point this coming week. Really excited to see how the perf looks :)

src/proto/streams/prioritize.rs
```diff
@@ -505,30 +505,30 @@ impl Prioritize {
...
-    pub fn poll_complete<T, B>(
+    pub fn buffer_pending<T, B>(
         &mut self,
-        cx: &mut Context,
         buffer: &mut Buffer<Frame<B>>,
         store: &mut Store,
         counts: &mut Counts,
         dst: &mut Codec<T, Prioritized<B>>,
-    ) -> Poll<io::Result<()>>
+    ) -> io::Result<bool>
```
seanmonstar: I find returned booleans in these cases to be hard to remember what they mean. Could we make a small two variant enum with clearer variant names?

seanmonstar (changes requested):
Very nice work! I'm leaning more towards landing smaller improvements than my gigantic one (that seems to keep having bugs in it). Just one inline point about booleans.

brentechols:
Hopefully this is closer to what you had in mind, lmk if you have any other feedback

seanmonstar:
Hm, looks like a few of the unit tests are hanging.

brentechols:
okay - I have all tests running happily, and also found another race condition that is now resolved

seanmonstar (approved):
Nice!

PR #921, opened 2026-07-16
fix: error instead of busy-looping when write returns `Ok(0)`

ArniDagur:
This fixes a bug we saw in production. There's a nasty interaction between `h2` and the `ktls` crate. Namely:

1. When the write-side of the connection is closed, the `ktls` crate returns `Ok(0)` instead of an error.
2. `h2` retries `Ok(0)` writes, instead of treating them as errors.

This led to busy looping, and 100% CPU utilization. Ideally, this should be fixed in both the `h2` and `ktls` crates, but I'm targeting `h2` first because of its central place in the ecosystem.

This fix mirrors existing logic in hyper HTTP/1: https://github.com/hyperium/hyper/blob/v1.10.1/src/proto/h1/io.rs#L285-L297

### DIFF
--- a/src/codec/framed_write.rs
+++ b/src/codec/framed_write.rs
@@ -135,7 +135,7 @@ where
 
         loop {
             while !self.encoder.is_empty() {
-                match self.encoder.next {
+                let n = match self.encoder.next {
                     Some(Next::Data(ref mut frame)) => {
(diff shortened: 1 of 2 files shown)


seanmonstar (approved):
Oof. Makes sense to me, thanks for the fix!

PR #922, opened 2026-07-17
fix: preserve received END_STREAM across HTTP/2 stream resets

### LINKED ISSUE #4130: HTTP/2 proxy loses END_STREAM on early header-only response
### Version

hyper 1.10.1  h2 0.4.15

### Platform

Linux zen 7.1.3-arch1-3.1 #1 SMP PREEMPT_DYNAMIC Tue, 14 Jul 2026 13:20:11 +0000 x86_64 GNU/Linux

### Summary

An HTTP/2 hyper proxy changes the framing of headers-only responses when the server follows it with `RST_STREAM(NO_ERROR)`.

The upstream server sends:

```text
HEADERS + END_STREAM
RST_STREAM(NO_ERROR)
```

The proxy sends downstream:

```text
HEADERS
empty DATA + END_STREAM
RST_STREAM(NO_ERROR)
```

This matters for any protocol where the final metadata is carried in the header-only response. I encountered it with a gRPC trailers-only response. gRPC clients assume that if the HEADERS frame does not contain END_STREAM, then there will be a trailers HEADER frame that does.

### Code Sample

```rust
# Run server with docker run -it --rm -p 9000:9000 -p 9001:9001 moul/grpcbin

use std::convert::Infallible;

use http::Request;
use hyper::body::Incoming;
use hyper::service::service_fn;
use hyper_util::rt::{TokioExecutor, TokioIo};
use tokio::net::{TcpListener, TcpStream};
use tokio::process::Command;

const BACKEND: &str = "127.0.0.1:9000";

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
    let listener = TcpListener::bind("127.0.0.1:0").await?;
    let proxy = listener.local_addr()?.to_string();
    tokio::spawn(run_proxy(listener));

    grpcurl("direct", BACKEND).await?;
    grpcurl("through proxy", &proxy).await?;
    Ok(())
}

async fn run_proxy(listener: TcpListener) {
    loop {
        let (socket, _) = listener.accept().await.unwrap();
        tokio::spawn(async move {
            let service = service_fn(|request: Request<Incoming>| async move {
                let upstream = TcpStream::connect(BACKEND).await.unwrap();
                let (mut sender, connection) =
                    hyper::client::conn::http2::Builder::new(TokioExecutor::new())
                        .handshake(TokioIo::new(upstream))
                        .await
                        .unwrap();
                tokio::spawn(async move {
                    let _ = connection.await;
                });

                Ok::<_, Infallible>(sender.send_request(request).await.unwrap())
            });

            let _ = hyper::server::conn::http2::Builder::new(TokioExecutor::new())
                .serve_connection(TokioIo::new(socket), service)
                .await;
        });
    }
}

async fn grpcurl(
    label: &str,
    address: &str,
) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
    println!("--- {label} ---");
    let status = Command::new("grpcurl")
        .arg("-plaintext")
        .arg("-d")
        .arg(r#"{"greeting":"x"}"#)
        .arg(address)
        .arg("hello.HelloService/SayHello")
        .status()
        .await?;
    println!("{status}\n");
    Ok(())
}
```

### Expected Behavior

```
--- direct ---
{
  "reply": "hello x"
}
exit status: 0

--- through proxy ---
Error invoking method "hello.HelloService/SayHello": rpc error: code = Internal desc = failed to query for service descriptor "hello.HelloService": server closed the stream without sending trailers
exit status: 1
```

Expected: gRPC traffic can proxy through a hyper reverse proxy succesfully.

### Actual Behavior

gRPC client breaks

### Additional Context

Below is my guesses on this, may be wrong.

After receiving the upstream response, Hyper exposes:

```text
response headers: present
body.is_end_stream(): false
first body frame: None
```

`Incoming::poll_frame` suppresses `RST_STREAM(NO_ERROR)` by returning `None`, which is appropriate for an early response. However, the body no longer reports that the response HEADERS already carried `END_STREAM`.

When that response is returned from a Hyper server, Hyper emits ordinary HEADERS followed by an empty DATA frame with `END_STREAM`.

A proxy cannot reconstruct the original framing without buffering or using protocol-specific response headers.

Maybe related
- #2872
- #3275
- #3681
- hyperium/h2#703

howardjohn:
Keep resets received after EOS distinct from ordinary stream errors. This allows the receive half to remain complete while the send half still observes the reset and its reason.

This fixes header-only responses followed by RST_STREAM(NO_ERROR), which could otherwise lose their END_STREAM state when passed through a proxy.

Fixes https://github.com/hyperium/hyper/issues/4130

### DIFF
--- a/tests/h2-tests/tests/stream_states.rs
+++ b/tests/h2-tests/tests/stream_states.rs
@@ -878,6 +878,15 @@ fn exceed_max_streams() {
 
 #[tokio::test]
 async fn recv_end_stream_survives_reset() {
+    assert_recv_end_stream_survives_reset(Reason::NO_ERROR).await;
+}
+
+#[tokio::test]
+async fn recv_end_stream_preserves_reset_reason() {
+    assert_recv_end_stream_survives_reset(Reason::INTERNAL_ERROR).await;
+}
+
+async fn assert_recv_end_stream_survives_reset(reset_reason: Reason) {
     h2_support::trace_init!();
     let (io, mut srv) = mock::new();
 
@@ -887,8 +896,7 @@ async fn recv_end_stream_survives_reset() {
         srv.recv_frame(frames::headers(1).request("POST", "https://example.com/"))
             .await;
         srv.send_frame(frames::headers(1).response(200).eos()).await;
-        srv.send_frame(frames::reset(1).reason(Reason::NO_ERROR))
-            .await;
+        srv.send_frame(frames::reset(1).reason(reset_reason)).await;
     };
 
     let client = async move {
@@ -905,7 +913,7 @@ async fn recv_end_stream_survives_reset() {
             .drive(poll_fn(move |cx| send_stream.poll_reset(cx)))
             .await
             .unwrap();
-        assert_eq!(reason, Reason::NO_ERROR);
+        assert_eq!(reason, reset_reason);
 
         let response = response.await.unwrap();
         assert!(response.body().is_end_stream());
--- a/tests/h2-tests/tests/stream_states.rs
+++ b/tests/h2-tests/tests/stream_states.rs
@@ -877,6 +877,48 @@ fn exceed_max_streams() {
 */
 
 #[tokio::test]
+async fn recv_end_stream_survives_reset() {
+    h2_support::trace_init!();
+    let (io, mut srv) = mock::new();
+
+    let srv = async move {
+        let settings = srv.assert_client_handshake().await;
+        assert_default_settings!(settings);
+        srv.recv_frame(frames::headers(1).request("POST", "https://example.com/"))
+            .await;
+        srv.send_frame(frames::headers(1).response(200).eos()).await;
+        srv.send_frame(frames::reset(1).reason(Reason::NO_ERROR))
+            .await;
+    };
+
+    let client = async move {
+        let (mut client, mut conn) = client::handshake(io).await.expect("handshake");
+        let request = Request::builder()
+            .method(Method::POST)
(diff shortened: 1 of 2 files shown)


seanmonstar:
Could we add an integration test, maybe stream_states or so?

tests/h2-tests/tests/stream_states.rs
```diff
@@ -876,6 +876,48 @@ fn exceed_max_streams() {
...
+    h2_support::trace_init!();
+    let (io, mut srv) = mock::new();
+
+    let srv = async move {
+        let settings = srv.assert_client_handshake().await;
+        assert_default_settings!(settings);
+        srv.recv_frame(frames::headers(1).request("POST", "https://example.com/"))
+            .await;
+        srv.send_frame(frames::headers(1).response(200).eos()).await;
+        srv.send_frame(frames::reset(1).reason(Reason::NO_ERROR))
```
seanmonstar: I was wondering, what about if after EOS a different error code is sent? So, perhaps a real world situation, the peer has sent all of its expected data, but is receiving what we're sending, and at one point later, decides _we_ did something bad, or they panicked and sent us an `INTERNAL_ERROR`, or something. Would we still propagate that onward?

howardjohn: Looks like yes - I just added a variant of the test sending a different error and we do propagate the error

seanmonstar (approved):
Nice!

Issue #883, opened 2026-03-30
load_hpack takes ~9% of total CPU at high request rates


aryehlev:
Running an HTTP/2 server at ~1M req/s on ARM64, load_hpack is ~9% of total CPU in our profiles. Feels way too high for header decoding. I looked through the source and I think there's a lot of low-hanging fruit. Happy to send patches.

<img width="383" height="324" alt="Image" src="https://github.com/user-attachments/assets/7ff6db0d-0e28-4644-8990-acd084c8c746" />

seanmonstar:
Sure, optimizations are always welcome!

seanmonstar:
FWIW, there's one optimization PR up now, #926.

aryehlev:
@seanmonstar  awesome! i didnt get to it, but once you merge i will test it

PR #924, opened 2026-07-22
Revert "perf(header): optimize `HeaderValue` creation via zero-copy sharing"

### LINKED ISSUE #923: Suspected memory usage regression since 0.4.14: zero-copy HeaderValue decoding (#884) pins connection read buffers
### Version

h2 0.4.14

### Platform

Linux 5.10.260-259.1053.amzn2.aarch64

### Summary

Disclaimer: I used AI to root cause this issue without full context on the details myself.

When attempting to deploy an upgrade from `h2 0.4.13 -> 0.4.15` and `hyper 1.9.0 -> 1.10.1`, we observed a significant step up in memory consumption in our process (from 13 GB to 21 GB). By analyzing memory profiles before/after the change and related commits, we found the regression was likely caused by #884. 

I admittedly don't have a full understanding of the implementation details, but the below summary from AI does plausibly explain the regression. It looks like the optimization to use `HeaderValue::from_maybe_shared` results in increased memory usage for the lifetime of a connection. Would love thoughts on this theory.

### Code Sample

```rust

```

### Expected Behavior

I expect memory usage to not increase significantly between versions.

### Actual Behavior

Memory usage increased significantly when updating to h2 0.4.14.

### Additional Context

As mentioned, I used AI when investigating this issue. Below is the AI summary of the cause:

### Summary

After upgrading `h2` 0.4.13 → 0.4.15, a production gRPC server (tonic 0.14.6 / hyper 1.10.1) saw steady-state heap usage increase by roughly 560 MiB (~16%). We identified a suspect change, #884 ("perf(header): optimize `HeaderValue` decoding via zero-copy sharing", commit `55a0d9d`, first released in 0.4.14).

The optimization makes decoded header *field values* `Bytes` slices that share the connection's read-buffer allocation instead of copying into a small dedicated allocation. Any header value that outlives the frame it arrived in now pins the **entire** backing read-buffer allocation — and two common retention paths make that the normal case rather than the exception:

1. Request/response `HeaderMap`s hold the values for the lifetime of each stream.
2. Headers encoded as *literal with incremental indexing* are cloned into the HPACK dynamic table, which lives for the **lifetime of the connection**.

### The change

`src/hpack/header.rs`, before and after #884:

```rust
// 0.4.13
let value = HeaderValue::from_bytes(&value)?;        // copies into a fresh allocation
// 0.4.14+
let value = HeaderValue::from_maybe_shared(value)?;  // retains the shared slice
```

The zero-copy slicing itself is not new — `StringMarker::consume` → `take()` → `split_to`/`freeze` already produced `Bytes` sharing the `FramedRead`/`LengthDelimitedCodec` read buffer in 0.4.13. What changed is that the shared slice used to be transient: `HeaderValue::from_bytes(&value)` copied the bytes out immediately and the slice was dropped before the decode call returned. After #884 the `HeaderValue` *is* the shared slice.

### Why this pins large allocations

- `FramedRead` uses `tokio_util`'s `LengthDelimitedCodec`, whose read `BytesMut` is a single large allocation that frames are carved out of. A retained header value of a few dozen bytes keeps a reference to that whole allocation.
- With outstanding `Bytes` references into the buffer, `BytesMut::reserve` can no longer reclaim/reuse the allocation and must allocate a fresh one — while old generations stay alive as long as any header value points into them. Our heap profiles show exactly this: allocations attributed to `LengthDelimitedCodec::decode_head → BytesMut::reserve → reserve_inner` grew from ~50 MiB to ~638 MiB.
- Huffman-decoded strings hit an analogous path through the decoder's shared scratch buffer (`huffman::decode` writes into `self.buffer`, which is split and frozen), so they pin that allocation's generations instead.

### Why long-lived connections make it worse (ratchet effect)

The HPACK dynamic table is capped at 4,096 logical bytes (≤ ~128 entries), but eviction only happens on *insertion*. gRPC clients typically have a stable header set: after warmup they mostly send `Indexed` representations and stop inserting. At that point the table freezes, and whatever read-buffer generations its entries pin stay pinned for the connection's entire remaining lifetime. Per-connection overhead ratchets up to (≤128 entries + in-flight streams) × (size of each backing allocation) and never comes back down. Total memory then scales with live connection count.

For a gRPC fleet with many long-lived connections this is a significant regression; the effective cost per connection can be several large buffer generations, dwarfing the memcpy the optimization saves.

### Observed impact

Comparing jemalloc heap profiles (sampled, `lg_prof_sample≈24`, so values are weighted estimates) on the same service/traffic before and after the upgrade:

| Allocation path | 0.4.13 | 0.4.15 |
| --- | ---: | ---: |
| `h2::server::Connection::poll_closed → LengthDelimitedCodec::decode_head → BytesMut::reserve_inner` | ~50 MiB | ~638 MiB |
| client-side response decode (same mechanism via `SendRequest`) | ~34 MiB | ~117 MiB |

The post-upgrade profile also newly shows retention under `h2::hpack::decoder::Table::insert` / `Headers::load_hpack`, consistent with the dynamic table holding shared slices. No other change in the h2/hyper/bytes/tonic diff range touches receive-buffer behavior (we diffed all four; e.g. `bytes::BytesMut::reserve_inner` is byte-identical between 1.11.1 and 1.12.0).

0x676e67:
close: https://github.com/hyperium/h2/issues/923

### DIFF
--- a/src/hpack/header.rs
+++ b/src/hpack/header.rs
@@ -96,7 +96,7 @@ impl Header {
         } else {
             // HTTP/2 requires lower case header names
             let name = HeaderName::from_lowercase(&name)?;
-            let value = HeaderValue::from_maybe_shared(value)?;
+            let value = HeaderValue::from_bytes(&value)?;
 
             Ok(Header::Field { name, value })
         }
@@ -231,7 +231,7 @@ impl<'a> Name<'a> {
         match self {
             Name::Field(name) => Ok(Header::Field {
                 name: name.clone(),
-                value: HeaderValue::from_maybe_shared(value)?,
+                value: HeaderValue::from_bytes(&value)?,
             }),
             Name::Authority => Ok(Header::Authority(BytesStr::try_from(value)?)),
             Name::Method => Ok(Header::Method(Method::from_bytes(&value)?)),


seanmonstar:
Thanks! Just for history sake, I've been trying to find why we made that change originally. But I cannot find an issue anywhere describing it. I vaguely remember someone showing a flamegraph pointing at that function? Or it's all blurring together...

0x676e67:
I think.

https://github.com/hyperium/h2/issues/883

seanmonstar:
Hm, I mean maybe. But it doesn't quite look like as closely related as I thought. My memory fails me. Oh well, yea let's revert for now. Thanks!

seanmonstar (approved):

PR #942, opened 2026-08-20
feat: allow configuring the DATA frame budget

muhamadazmy:
Follow up to #935

## Why
Our application protocol can (and will) send many small frames over the h2
connection. This was not an issue until the latest h2 update (PR #935) which
penalize small frames. 

## Fix
Until now the data-frame-budget was hard-coded to DEFAULT_DATA_FRAME_BUDGET, which
leaves no way to tune the budget value.

- [x] The description, docs, and comments (words meant for humans) are written by a human, not by an LLM. (https://seanmonstar.com/micro/20260729-i-want-your-own-words/)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -343,6 +343,11 @@ pub struct Builder {
     ///
     /// When this gets exceeded, we issue GOAWAYs.
     local_max_error_reset_streams: Option<usize>,
+
+    /// connection-level budget for DATA framing overhead.
+    ///
+    /// When this gets exhausted, we issue a GOAWAY with `ENHANCE_YOUR_CALM`.
+    data_frame_budget: usize,
 }
 
 #[derive(Debug)]
@@ -663,6 +668,7 @@ impl Builder {
             settings: Default::default(),
             stream_id: 1.into(),
             local_max_error_reset_streams: Some(proto::DEFAULT_LOCAL_RESET_COUNT_MAX),
+            data_frame_budget: proto::DEFAULT_DATA_FRAME_BUDGET,
         }
     }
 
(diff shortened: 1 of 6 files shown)


seanmonstar:
Thanks for the PR! Sorry for the disruption. I had hoped that by releasing back to the budget when the data frame was consumed, it would just work out.

We could add a config option, as you've done. But, I'd like to explore if we can make it just work without one first.

Is the problem that you send a bunch of small data frames, and don't read them that quickly? Or are they sent on a much larger number of active streams?

muhamadazmy:
Hey @seanmonstar, thank you so much for checking this PR so quickly. 

The problem is that we have small data frames on large number of active streams. Of course it varies depends on the workload shape. Some streams can generate many small steps each with tiny outputs, which triggers the issue if we have many multiple invocations running concurrently on the same connection.

Once I put the budget to higher value, the system had enough time to process (and replenish) the budget and I no longer hit the issue.

Of course there is no "good" value to the budget, and it's totally up to our workload shape. Hence I thought a configuration option could be the best approach.

cbpodd:
Hi @seanmonstar - I'd like to confirm that I've hit this too. We run a gRPC server where we send lots of small updates. We're quite latency sensitive, so batching isn't really an option for us. Would love to see this configuration option exposed in `h2` (and eventually `hyper` and `hyper-util`).

seanmonstar:
Looking at this some more, would an increase to about 1000 (currently 100) be enough in most cases?

Again, if we have to, we can add a config option, but I'd rather not have to propagate something through all the other dependencies. Especially if we eventually came up with a better budget/defense mechanism in the first place.

muhamadazmy:
@seanmonstar IMHO since h2 is a low level crate, it shouldn't have any assumptions about the default values. 

I am not sure that a budget of 250KiB would be enough for all our use cases. `hyper` and `hyper-util` can still choose a sane default and they don't have to expose all the low level options.

We personally use `h2` directly to have maximum control, and is working great for us.

muhamadazmy:
I mean 250k would definitely work with my local test env where I can reproduce this issue. But I am not sure about all other scenarios in production where we can have many h2 connections open at max number of stream capacity per connection (and heavily loaded server) specially that the number of frames and sizes of the data is largely unknown.

cbpodd:
@seanmonstar I agree with @muhamadazmy here. I don't think `h2` should take a policy on this. For us we'd likely set it at ~2M+. 

Are you sure that this hard cap is the best way to handle this? I wonder if it would be better to handle this via limiting actually-empty frames and/or backpressuring via the existing connection/stream windows. 

Maybe including the overhead of `DATA` frames into the existing connections/stream window calculations would prevent this vulnerability without penalizing legitimate applications sending a lot of small frames.

cbpodd:
Also we use this through Tonic - so having `hyper`/`hyper-utils` set s low default would not solve our problem.

seanmonstar:
OK, yea true, a config here doesn't _mean_ it has to bubble all the way up just yet. We can adjust defaults and heuristics at a higher layer.

> Are you sure that this hard cap is the best way to handle this? I wonder if it would be better to handle this via limiting actually-empty frames and/or backpressuring via the existing connection/stream windows.

Actually empty frames are one case, but it's also been noticed that, for example, 1-length frames are limited by flow control, but they still 'cost' more memory than that because of overhead. We can do things to reduce that overhead, but it still is a potential abuse vector.

So, I'll move ahead with providing the config. If a better way to distinguish and improve the heuristic it possible, we can discuss in a separate issue. Thanks all!

seanmonstar (approved):

muhamadazmy:
Thank you so much @seanmonstar for your time and great effort ❤️

PR #946, opened 2026-08-21
refactor: better auto default for data frame budget

### LINKED ISSUE #944: h2 fix for the recent security vulnerability breaks downstream consumers
### Version

v0.14.6+

### Platform

All

### Summary

Hyper fixed [this security vulnerability in v0.4.16](https://github.com/hyperium/hyper/security/advisories/GHSA-q83h-524g-xf6h) by having a fixed budget for undersized `DATA` frames. However, this penalizes real-time streaming applications which regularly send large numbers of undersized frames and buffer them in clients. This causes legitimate streams to disconnect using `v0.4.16`+ that did not in `v0.4.15`.

### Expected Behavior

Legitimate streams that worked in `v0.4.15` should work in `v0.4.16` without additional configuration (it's a patch version bump).

### Actual Behavior

Streams with high-volume undersized `DATA` frames were cut off with `GOAWAY ENHANCE_YOUR_CHILL` in `v0.4.16` that streamed fine in `v0.4.15`.

### Additional Context

[h2 PR #942](https://github.com/hyperium/h2/pull/942) partially fixes this problem by allowing users to set the budget for undersized data frames. However, our stack is `tonic` -> `hyper-utils` -> `hyper` -> `h2`. 

I'd be happy to submit PRs that plumb this code through the `hyperium` stack. However, I think there might be a better option here that allows h2 to infer these windows from `connection_window`. 

As I understand it, there are two issues at play here:
1. 0-byte `DATA` frames are never exposed to the application. These need to be rate limited and a connection may need to be killed if too many are sent. The hard-cap is fine here - I don't think there's a legitimate use-case for non-final 0-byte `DATA` frames.
2. Undersized `DATA` frames need to be controlled in some way. This is where the configurable buffer comes in. 
    - I think it might be better to derive that buffer from the `connection_window`.
    - This will allow streams with limited requirements to break quickly, while allowing those with more memory to accumulate, buffer, and drain more bytes before breaking.

I'm happy to go down either of these paths and I'm happy to submit PRs. I'd like to know where best to direct my efforts.
1. I can submit PRs to expose these configuration parameters in `hyper` and `hyper-utils`. 
2. I can derive the allowed buffer from the `connection_window`. 

Let me know what's best here!

@seanmonstar

seanmonstar:
Instead of choosing a specific small default constant, the small data frame overhead budget will now be calculated based on the configured connection window size. Setting the budget directly will skip the default auto value.

Closes #944


Sruhvx-jpg:
Small nit on `resolve`:
   we can avoid bouncing through `u64` with `(window as usize / 2).max(DEFAULT_DATA_FRAME_BUDGET)`.

   Also as a naive proposal, `window >> 1` could be used instead of `/ 2`.

PR #951, opened 2026-08-31
fix: reject GOAWAY frames with a non-zero stream identifier

jaideeppyne:
RFC 9113 section 6.8:

> The GOAWAY frame applies to the connection, not a specific stream. An endpoint MUST treat a GOAWAY frame with a stream identifier other than 0x00 as a connection error (Section 5.4.1) of type PROTOCOL_ERROR.

`GoAway::load` is the only frame loader that never got the frame head, so the stream id was never looked at. A GOAWAY sent on stream 1 was accepted as a normal connection shutdown.

I found this probing the connection-level frames against the spec. SETTINGS, PING and RST_STREAM all already reject a bad stream id; GOAWAY was the odd one out. Against `examples/server`, GOAWAY on stream 1 got `GOAWAY(NO_ERROR)` back, byte for byte the same as the legal stream-0 case. With this change it gets `GOAWAY(PROTOCOL_ERROR)`, and stream 0 still gets `NO_ERROR`.

The fix passes `head` into `load` and checks `is_zero()`, matching what `Ping::load` does. The existing generic error mapping in `framed_read` already turns that into a connection-level PROTOCOL_ERROR.

Oracle and counts:
- new codec test `read_goaway_stream_id_not_zero` fails before, passes after (I reverted only the two source files and kept the test to confirm: `expected error; actual=Some(Ok(GoAway { error_code: NO_ERROR, .. }))`)
- h2spec 2.6 `--strict` against `examples/server` is unchanged: 147 tests, 146 passed, 1 skipped, 0 failed. h2spec's 6.8 test passes either way because it accepts a closed connection as evidence of a connection error, and h2 was closing anyway by honouring the GOAWAY
- full `cargo test --all-features` is unchanged

Unrelated, but `proto::streams::recv::tests::clear_recv_buffer_caps_capacity_before_overflow` already fails on a pristine master for me (`assertion failed: self.slab.is_empty()` at store.rs:231). Not touched here, happy to file it separately.

AI disclosure: I used Claude Code to help with this change.


seanmonstar (approved):
Nice improvement, thanks!

Issue #631, opened 2022-08-01
Mem leak when releasing stream


LYF1999:
https://github.com/hyperium/h2/blob/756384f4cdd62bce3af7aa53a156ba2cc557b5ec/src/proto/streams/counts.rs#L164-L168

There may still leave some bytes releated to the stream in https://github.com/hyperium/h2/blob/756384f4cdd62bce3af7aa53a156ba2cc557b5ec/src/proto/streams/recv.rs#L52 when `Counts` remove a closed `Stream`. 
This may cause  mem leak, Should we remove bytes releated to the stream in `remove` function?

LYF1999:
I am gald to submit a PR to fix that, but we can not access `Recv` in `Counts`. if we do that, it will be a large change commit.

Millione:
@seanmonstar can we have a look?

seanmonstar:
Sure!

PR #957, opened 2026-09-10
fix(stream): reclaim unaccepted push buffers and capacity

0x676e67:
This might fix the issue: [https://github.com/hyperium/h2/issues/631](https://github.com/hyperium/h2/issues/631). However, it is quite old, and no one has reported a similar problem recently. I’m leaving it here in case anyone remembers that they need it.

- [x] The description, docs, and comments (words meant for humans) are written by a human, not by an LLM. (https://seanmonstar.com/micro/20260729-i-want-your-own-words/)


seanmonstar (approved):
Nice catch!

